PR #1, opened 2017-07-19
Instance interface with MemoryInstance reference

werkt:
Initial implementation of grpc services, instance routing, instance
interface definition, a reference in-memory instance implementation, as
well as a worker implementation. The worker implementation uses an
operation queue grpc API for requesting operations.


damienmg:
Added googlers that should be interested :)

PR #3, opened 2017-07-21
Use ByteString#writeTo for operation stream output

werkt:
Prevent the awkward toByteArray and use the supported stream output
mechanism for ByteString for operation stream writes.

### DIFF
--- a/src/main/java/build/buildfarm/server/ByteStreamService.java
+++ b/src/main/java/build/buildfarm/server/ByteStreamService.java
@@ -262,7 +262,7 @@ public class ByteStreamService extends ByteStreamGrpc.ByteStreamImplBase {
         String operationStream = parseOperationStream(resourceName);
 
         OutputStream outputStream = instance.getStreamOutput(operationStream);
-        outputStream.write(request.getData().toByteArray());
+        request.getData().writeTo(outputStream);
         if (request.getFinishWrite()) {
           outputStream.close();
         }


src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -262,9 +262,11 @@ private void writeOperationStream(
         String operationStream = parseOperationStream(resourceName);
 
         OutputStream outputStream = instance.getStreamOutput(operationStream);
-        outputStream.write(request.getData().toByteArray());
-        if (request.getFinishWrite()) {
-          outputStream.close();
+        if (outputStream != null) {
```
ulfjack: Document when this can be null?

src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -262,9 +262,11 @@ private void writeOperationStream(
...
+          request.getData().writeTo(outputStream);
+          if (request.getFinishWrite()) {
+            outputStream.close();
```
ulfjack: What's your strategy for making sure all output streams are closed? In the remote worker, we've discovered at least one file handle leak due to streams not being closed properly in all cases.

src/main/java/build/buildfarm/server/ByteStreamService.java
werkt: This got merged in as a part of continuing work where I didn't have an implementation of getStreamOutput (it returned null in that case).  I will drop the check, as it shouldn't happen with a properly implemented instance.  I've toyed with the idea of a stream enabled feature/config, if I put that in, I'll put a check at a level higher than this.

src/main/java/build/buildfarm/server/ByteStreamService.java
werkt: I'm open to suggestions - it should be the instance's responsibility to maintain that strategy, and making a forced close at complete operation post or requeue seems like a start.

src/main/java/build/buildfarm/server/ByteStreamService.java
ulfjack: Right now I still see the null check on GitHub, so I'll wait with merging this PR.

werkt: Got distracted yesterday.  Updated now.

PR #6, opened 2017-08-02
Concurrency validation fixes

werkt:
A number of changes for addressing some concurrency, validation, and general behavior issues encountered with further testing.


src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -324,6 +324,7 @@ public void execute(
           if (o.getDone()) {
             onOperation.accept(o);
           }
+          return true;
```
ulfjack: Should this change have gone in the previous commit?

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -153,15 +153,26 @@ protected void enqueueOperation(Operation operation) {
...
       if (operation.getDone()) {
         watchers.remove(operation.getName());
       }
+      long unfilteredWatcherCount = operationWatchers.size();
       super.updateOperationWatchers(operation);
-      for (Consumer<Operation> watcher : operationWatchers) {
-        watcher.accept(operation);
+      ImmutableList.Builder<Function<Operation, Boolean>> filteredWatchers = new ImmutableList.Builder<>();
+      long filteredWatcherCount = 0;
+      for( Function<Operation, Boolean> watcher : operationWatchers ) {
```
ulfjack: style: space before (, not inside ()

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -321,17 +332,18 @@ private boolean satisfiesRequirements(Platform platform, Operation operation) {
...
+      Function<Operation, Boolean> watcher) {
+    if (watchInitialState &&
+        !watcher.apply(getOperation(operationName))) {
+      return false;
+    }
+    List<Function<Operation, Boolean>> operationWatchers = watchers.get(operationName);
     if (operationWatchers == null) {
       if (watchInitialState) {
-        watcher.accept(null);
+        boolean ignored = watcher.apply(null);
```
ulfjack: the ignored variable is not used anywhere

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
ulfjack: Oh, I see. Ignore this.

src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -473,6 +473,7 @@ public static void main(String[] args) throws Exception {
     Path configPath = Paths.get(args[0]);
     InputStream configInputStream = Files.newInputStream(configPath);
```
ulfjack: Prefer try-with-resources here to guarantee closing even if there's an exception.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -436,6 +444,14 @@ public String listOperations(
   }
 
   @Override
+  public void deleteOperation(String name) {
+    if (completedOperations.remove(name) == null &&
+        outstandingOperations.containsKey(name)) {
+      throw new UnsupportedOperationException();
```
ulfjack: Throw IllegalArgumentException or IllegalStateException instead?

ulfjack:
Is there any way for me to merge some of the commits, but not all of them?

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: I'll go with IllegalStateException - I wanted to only allow 'unsupported' behavior to be passed through akin to the interface's only defined exceptional behavior, and then ignore the rest due to the Empty response.  This should probably be made into a concrete implementation requirement, rather than the abstract server.

ulfjack:
I found a page that describes how to do that on the command line. I'm not sure I want to do that, though...

werkt:
I can merge any of the commits into sets, that will just be a rebase on my part, and submit - let me know what sort of grouping you want.

ulfjack:
There are a bunch of simple refactorings that I'd be happy to merge immediately. The non-simple changes are what's holding up submitting the simple ones.

src/main/java/build/buildfarm/worker/Worker.java
werkt: The try-with-resources makes this (and most scenarios I want to use it in) awkward when I really wanted to just release the resource early, but now I've got a worker that needs to be declared outside the block, I've got to check for null, etc etc.  I'll put the block in, with the manual close, proceeding with the other operations.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: This was an attempt to suppress a return ignored error, which I've thought better of.  When we can't find the operation, the watch will simply fail with no callback invocation.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: Added ca1b5d839433cccc7d8b9a866eb5208fe572d4a1 to deal with this one, and fixed it for the subsequent additions.

werkt:
I'll rework into some independent streams

PR #11, opened 2017-08-03
Chunk ReadResponses from ByteStream service

werkt:
Requests for sufficiently large files will trigger frame errors in grpc
response delivery. Chunking the data in ByteStream responses allows for
requests up to the arbitrary request limit.

### DIFF
--- a/src/main/java/build/buildfarm/server/ByteStreamService.java
+++ b/src/main/java/build/buildfarm/server/ByteStreamService.java
@@ -106,9 +106,20 @@ public class ByteStreamService extends ByteStreamGrpc.ByteStreamImplBase {
       return;
     }
 
-    responseObserver.onNext(ReadResponse.newBuilder()
-        .setData(blob)
-        .build());
+    while (!blob.isEmpty()) {
+      ByteString chunk;
+      if (blob.size() < DEFAULT_CHUNK_SIZE) {
+        chunk = blob;
+        blob = ByteString.EMPTY;
+      } else {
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -106,9 +106,20 @@ private void readBlob(
...
+    while (!blob.isEmpty()) {
+      ByteString chunk;
+      if (blob.size() < DEFAULT_CHUNK_SIZE) {
+        chunk = blob;
+        blob = ByteString.EMPTY;
+      } else {
+        chunk = blob.substring(0, (int) DEFAULT_CHUNK_SIZE);
+        blob = blob.substring((int) DEFAULT_CHUNK_SIZE);
+      }
+      responseObserver.onNext(ReadResponse.newBuilder()
```
ulfjack: FYI: This code loads the entire file into memory and buffers it in non-heap memory in Netty, so the memory doesn't show up in most Java memory profilers. In Bazel, we were running into the problem of running out of non-heap memory when we were uploading many large files concurrently. This could happen here as well. Unfortunately, gRPC only supports application-level flow control for the lower-level API (ServerCall), which we may have to implement here if this turns out to crash the server.

werkt: Duly noted - the interface for the blob itself could be unnecessarily consumptive as well - since the content is guaranteed by the digest, and we can error the stream safely, I should only be requesting a chunk at a time from the instance.

PR #14, opened 2017-08-03
Directory cycle check fix, refactor validation

werkt:
Properly checking for directory cycles - previous implementation did not
account for the same directory content appearing in sibling or unrelated
directories under a root. Prevent directory revisits when cycles are not
possible, refactored the validation checks into separate methods.

### DIFF
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -17,15 +17,18 @@ package build.buildfarm.instance;
 import build.buildfarm.common.Digests;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.devtools.remoteexecution.v1test.Action;
 import com.google.devtools.remoteexecution.v1test.ActionResult;
 import com.google.devtools.remoteexecution.v1test.Command;
 import com.google.devtools.remoteexecution.v1test.Digest;
 import com.google.devtools.remoteexecution.v1test.Directory;
+import com.google.devtools.remoteexecution.v1test.DirectoryNode;
 import com.google.devtools.remoteexecution.v1test.ExecuteOperationMetadata;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -177,78 +180,91 @@ private void environmentVariablesUniqueAndSortedPrecondition(
...
-    Digest commandDigest = action.getCommandDigest();
-    ImmutableList.Builder<Digest> inputDigests = new ImmutableList.Builder<>();
-    inputDigests.add(commandDigest);
-    Iterator<Directory> directoriesIterator =
-      createTreeIterator(action.getInputRootDigest(), "");
+  private void validateActionInputDirectory(
+      Directory directory,
+      Stack<Digest> path,
+      Set<Digest> visited,
+      ImmutableSet.Builder<Digest> inputDigests) {
```
ulfjack: There may be multiple files with the same digest, but ImmutableSet.Builder doesn't allow that (it's not like a set). Use LinkedHashSet or something instead?

werkt: An empirical test of this indicates that ImmutableSet.Builder<> allows multiple adds to take place with the same value (a digest of the string "123") with no error and expected reduced behavior:

```
import build.buildfarm.common.Digests;
import com.google.common.collect.ImmutableSet;
import com.google.devtools.remoteexecution.v1test.Digest;
import com.google.protobuf.ByteString;

class test {
  public static void main(String[] args) {
    ImmutableSet.Builder<Digest> set = new ImmutableSet.Builder<>();

    set.add(Digests.computeDigest(ByteString.copyFromUtf8("123")));
    set.add(Digests.computeDigest(ByteString.copyFromUtf8("123")));

    System.out.println(set.build());
  }
}
```

```
$ test
[hash: "40bd001563085fc35165329ea1ff5c5ecbdbbeef"
size_bytes: 3
]
```

werkt: And this inputDigests set becomes only what we verify against findMissing with, meaning that multiple files may refer to the same content as you've described.

PR #16, opened 2017-08-03
Protect against duplicate directories from getTree

werkt:
If the list created as a result of (successive) getTree invocations
contains duplicates, it must be filtered for duplicates to populate the
directoriesIndex.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Worker.java
+++ b/src/main/java/build/buildfarm/worker/Worker.java
@@ -20,6 +20,7 @@ import build.buildfarm.instance.stub.StubInstance;
 import build.buildfarm.v1test.WorkerConfig;
 import build.buildfarm.v1test.CASInsertionControl;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
 import com.google.devtools.remoteexecution.v1test.Action;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -222,17 +225,21 @@ private boolean fetchInputs(Path execDir, Digest inputRoot) throws IOException {
...
-
     do {
       pageToken = instance.getTree(inputRoot, config.getTreePageSize(), pageToken, directories);
     } while (!pageToken.isEmpty());
 
-    directoriesIndex = Maps.uniqueIndex(
-        directories.build(),
-        directory -> Digests.computeDigest(directory));
+    Set<Digest> directoryDigests = new HashSet<>();
+    ImmutableMap.Builder<Digest, Directory> directoriesIndex = new ImmutableMap.Builder<>();
```
ulfjack: What happens if you get multiple directories with the same name but different digests? Is that something we need to protect against?

werkt: As in two entries in the DirectoryNode list (or FileNode list for that matter) with the same name? AbstractServerInstance#validateActions from #1 (and #validateActionInputs refactored in #14) has a precondition check for duplicate file/directory nodes (considers them in one Set). I will be refactoring those validations out into a common Directories (a la Digests) toolset for validating them both on the instance server and the worker in upcoming changes.

PR #19, opened 2017-08-03
Correctly interpret executable FileNodes

werkt:
The worker must set or strip the the executable bit on files it creates
with FileNodes, and should perform upkeep on previously created files
without the bit, which could occur with OOB filesystem manipulation.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Worker.java
+++ b/src/main/java/build/buildfarm/worker/Worker.java
@@ -21,6 +21,7 @@ import build.buildfarm.v1test.WorkerConfig;
 import build.buildfarm.v1test.CASInsertionControl;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
 import com.google.common.io.ByteStreams;
 import com.google.devtools.remoteexecution.v1test.Action;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -205,9 +207,20 @@ private boolean linkInputs(Path execDir, Digest inputRoot, Map<Digest, Directory
...
+      boolean createFile = !Files.exists(fileCachePath);
+      if (createFile) {
+        OutputStream outputStream = Files.newOutputStream(fileCachePath);
+        instance.getBlob(digest).writeTo(outputStream);
+        outputStream.close();
+      }
+      if (createFile || Files.isExecutable(fileCachePath) != fileNode.getIsExecutable()) {
+        ImmutableSet.Builder<PosixFilePermission> perms = new ImmutableSet.Builder<PosixFilePermission>()
+          .add(PosixFilePermission.OWNER_READ);
+        if (fileNode.getIsExecutable()) {
```
ulfjack: If you're sharing the file cache between multiple executions, then you could have a problem with a concurrent execution changing the permissions if there are actions that have the same file but with executable on/off. Is that a concern? Otherwise, I'd like to see it documented somewhere that the file cache is exclusive.

werkt: I'll say that there are other areas of the worker's operation that will have to be adjusted if the cache is shared amongst many content manipulators - the individual files in the cache will need to be guarded for example.  I would be happy to document that for this worker implementation, the cache needs to be exclusive (for now), where do you see that going?  README.md in the root?

ulfjack: Noone reads README.md files. How about documenting it here? Just a brief sentence is enough.

werkt: I decided to just modify the filename to correspond to the executable bit.  We'll double copy if the content matches, but I don't want to reimplement a filesystem to decouple file metadata from content.

PR #20, opened 2017-08-04
Use Concurrent data structures in memory instance

werkt:
Protect against basic data structure race conditions and collisions by
using the builtin Concurrent data structures and synchronizing
modifications to watchers in logical transactions.

### DIFF
--- a/src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
+++ b/src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
@@ -43,6 +43,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -77,8 +78,8 @@ public class MemoryInstance extends AbstractServerInstance {
     this(
         name,
         config,
-        /*contentAddressableStorage=*/ new HashMap<Digest, ByteString>(),
-        /*actionCache=*/ new HashMap<Digest, ActionResult>(),
+        /*contentAddressableStorage=*/ new ConcurrentHashMap<Digest, ByteString>(),
+        /*actionCache=*/ new ConcurrentHashMap<Digest, ActionResult>(),
         /*outstandingOperations=*/ new TreeMap<String, Operation>());
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -151,29 +152,31 @@ protected void enqueueOperation(Operation operation) {
...
-      }
-      long unfilteredWatcherCount = operationWatchers.size();
-      super.updateOperationWatchers(operation);
-      ImmutableList.Builder<Function<Operation, Boolean>> filteredWatchers = new ImmutableList.Builder<>();
-      long filteredWatcherCount = 0;
-      for (Function<Operation, Boolean> watcher : operationWatchers) {
-        if (watcher.apply(operation)) {
-          filteredWatchers.add(watcher);
-          filteredWatcherCount++;
+    synchronized(watchers) {
```
ulfjack: You're mixing synchronized with ConcurrentHashMap for watchers, which can confuse future maintainers of this code. I'd recommend to pick one or the other but not both.

ulfjack: Also, consider documenting on the watchers field that accesses need to be synchronized.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -77,8 +78,8 @@ public MemoryInstance(String name, MemoryInstanceConfig config) {
     this(
         name,
         config,
-        /*contentAddressableStorage=*/ new HashMap<Digest, ByteString>(),
-        /*actionCache=*/ new HashMap<Digest, ActionResult>(),
+        /*contentAddressableStorage=*/ new ConcurrentHashMap<Digest, ByteString>(),
```
ulfjack: Consider declaring the field and param as ConcurrentMap to prevent other callers from passing in a plain map.

werkt: tl;dr: actionCache/contentAddressableStorage population needs to be lifted out of constructor

I believe this betrays a larger interface problem here with usage vs. implementation.  The watchdog behavior (from operation-poll) represents concurrent manipulations of the structures under putOperation, but CAS and ActionCache are not represented in those manipulations.  Since the MemoryInstance does not perform other concurrent actions implicitly, the structures backing those two members needing to be concurrent is a function of a usage pattern (i.e. the multithreaded grpc handlers), and I'd rather not necessarily accommodate all usage patterns the implementation.

PR #21, opened 2017-08-09
Retain Done operations for getOperation/watchers

werkt:
Provide initial watcher data and a getOperation response for operations
which have completed. Expiration of this data will be left up to the
concrete instance implementation, with the deleteOperation request now
supported for completed operations.

### DIFF
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -46,16 +46,19 @@ public abstract class AbstractServerInstance implements Instance {
   protected final Map<Digest, ByteString> contentAddressableStorage;
   protected final Map<Digest, ActionResult> actionCache;
   protected final Map<String, Operation> outstandingOperations;
+  protected final Map<String, Operation> completedOperations;
 
   public AbstractServerInstance(
       String name,
       Map<Digest, ByteString> contentAddressableStorage,
       Map<Digest, ActionResult> actionCache,
-      Map<String, Operation> outstandingOperations) {
+      Map<String, Operation> outstandingOperations,
+      Map<String, Operation> completedOperations) {
     this.name = name;
     this.contentAddressableStorage = contentAddressableStorage;
     this.actionCache = actionCache;
     this.outstandingOperations = outstandingOperations;
+    this.completedOperations = completedOperations;
   }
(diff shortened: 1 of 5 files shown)


src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -434,7 +438,11 @@ protected void updateOperationWatchers(Operation operation) {
 
   @Override
   public Operation getOperation(String name) {
-    return outstandingOperations.get(name);
+    Operation operation = completedOperations.get(name);
+    if (operation == null) {
```
ulfjack: There's a possible race here. If there is a concurrent call to updateOperationWatchers, which removes an operation from outstandingOperations and adds it to completedOperations, then a call to this method may be unable to find the operation. I understand that's why you're doing a put before the remove above, but I'm not sure that the Jvm guarantees that the writes complete in that order. In any case, this subtlety should be documented in the code.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -466,6 +474,14 @@ public String listOperations(
   }
 
   @Override
+  public void deleteOperation(String name) {
```
ulfjack: This method doesn't actually do anything yet. What's the plan here?

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: You have it exactly right, both in intent and the likelihood of a race - this is a textbook transaction and needs to be treated as such - I'll update further in a bit.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: If the operation is in completedOperations, it is removed.  If the operation is outstanding, an exception is thrown (changed this to IllegalStateException as you recommended).  If the operation is neither completed nor outstanding, the request does nothing.

werkt: Sorry, not sure if my update was clear, it does do something - the remove is within the if block (probably shouldn't have side effects there with the test).  I'll update

PR #24, opened 2017-08-30
MemoryInstance Watchdog stop guarantee

werkt:
Synchronize a watchdog stop with its execution


src/main/java/build/buildfarm/instance/memory/Watchdog.java
```diff
@@ -62,6 +66,12 @@ private synchronized void reset(Duration timeout) {
 
   public synchronized void stop() {
     stopped = true;
-    this.notify();
+    try {
+      while (!done) {
+        this.notify();
+        this.wait();
+      }
+    } catch (InterruptedException ex) {
```
ulfjack: Swallowing interrupts is a problematic pattern that may come back to bite you.

PR #25, opened 2017-08-30
Implement newStreamInput for stub, refactor

werkt:
Provide a streaming download mechanism for blobs to minimize memory
pressure when fetching inputs. Refactored the stub getBlob method
to use this, factored out digest -> resourceName into an instance
interface mechanism (tightly couples server instance implementation to
worker implementation).

### DIFF
--- a/src/main/java/build/buildfarm/instance/stub/StubInstance.java
+++ b/src/main/java/build/buildfarm/instance/stub/StubInstance.java
@@ -31,6 +31,7 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
 import com.google.devtools.remoteexecution.v1test.Action;
 import com.google.devtools.remoteexecution.v1test.ActionCacheGrpc;
 import com.google.devtools.remoteexecution.v1test.ActionCacheGrpc.ActionCacheBlockingStub;
@@ -240,23 +241,27 @@ public class StubInstance implements Instance {
 
   @Override
   public InputStream newStreamInput(String name) {
-    throw new UnsupportedOperationException();
+    Iterator<ReadResponse> replies = bsBlockingStub
+        .get()
+        .read(ReadRequest.newBuilder().setResourceName(name).build());
+    return new ByteStringIteratorInputStream(Iterators.transform(replies, (reply) -> reply.getData()));
+  }
+
+  @Override
+  public String getBlobName(Digest blobDigest) {
+    return String.format(
+        "%s/blobs/%s",
+        getName(),
+        Digests.toString(blobDigest));
   }
 
   @Override
   public ByteString getBlob(Digest blobDigest) {
-    String resourceName = String.format(
-        "%s/blobs/%s",
-        getName(),
-        Digests.toString(blobDigest));
-    Iterator<ReadResponse> replies = bsBlockingStub
-      .get()
-      .read(ReadRequest.newBuilder().setResourceName(resourceName).build());
-    ByteString blob = ByteString.EMPTY;
-    while (replies.hasNext()) {
-      blob = blob.concat(replies.next().getData());
+    try (InputStream in = newStreamInput(getBlobName(blobDigest))) {
+      return ByteString.readFrom(in);
+    } catch (IOException ex) {
+      return null;
     }
-    return blob;
   }
 
   @Override
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -74,6 +74,14 @@ public abstract class AbstractServerInstance implements Instance {
   }
 
   @Override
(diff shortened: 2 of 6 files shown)


src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -73,10 +73,21 @@
   private final Channel channel;
   private final ByteStreamUploader uploader;
 
+  private static final InputStream nullInputStream = new InputStream() {
+    @Override
+    public int read() {
```
ulfjack: I'm not sure if you're worried about efficiency here, but you might want to override some of the other InputStream methods as well, e.g., the other read() method.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -240,23 +251,90 @@ public void write(byte[] b, int off, int len) throws IOException {
 
   @Override
   public InputStream newStreamInput(String name) {
-    throw new UnsupportedOperationException();
+    Iterator<ReadResponse> replies = bsBlockingStub
+      .get()
```
ulfjack: use four spaces indentation here?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -240,23 +251,90 @@ public void write(byte[] b, int off, int len) throws IOException {
...
+        }
+        return chunkIn.available();
+      }
+
+      @Override
+      public int read() throws IOException {
+        if (closed) {
+          throw new IOException("stream is closed");
+        }
+        byte[] b = new byte[1];
```
ulfjack: return super.read()?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -240,23 +251,90 @@ public void write(byte[] b, int off, int len) throws IOException {
...
+        return b[0];
+      }
+
+      @Override
+      public int read(byte[] b, int off, int len) throws IOException {
+        if (closed) {
+          throw new IOException("stream is closed");
+        }
+        int totalLen = 0;
+        while (len != 0) {
```
ulfjack: I've always found that it's surprisingly difficult to write a correct read method. Maybe move this to a named class, and write some tests for it?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -240,23 +251,90 @@ public void write(byte[] b, int off, int len) throws IOException {
...
+      @Override
+      public void close() {
+        closed = true;
+      }
+
+      private void advance() {
+        if (replies.hasNext()) {
+          chunkIn = replies.next().getData().newInput();
+        } else {
+          chunkIn = nullInputStream;
```
ulfjack: This seems awkward. What if you have a chunk that happens to be zero-lengths?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -240,23 +251,90 @@ public void write(byte[] b, int off, int len) throws IOException {
...
+          throw new IOException("stream is closed");
+        }
+        int totalLen = 0;
+        while (len != 0) {
+          int readLen = chunkIn.read(b, off, len);
+          if (readLen == -1) {
+            advance();
+            readLen = chunkIn.read(b, off, len);
+            if (readLen == -1) {
+              return totalLen == 0 ? -1 : totalLen;
```
ulfjack: I think it would be better to have an end-of-stream boolean indicator and use that to detect end of stream. Then you can just continue after the call to advance().

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: Agreed, I'm chucking the nullInputStream for ByteString.EMPTY.newInput() and iterating in advance() while hasNext and the data retrieved is empty.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: I need to put a linter in place for this.  Any recommendations?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: Agreed, if I were leaving this in.  Throwing it out per the awkward advance() fix below.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt:
```
src/main/java/build/buildfarm/instance/stub/StubInstance.java:263: error: abstract method read() in InputStream cannot be accessed directly
        return super.read();
```

Gotta have my simple implementation, it seems

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: Moved and tested

werkt:
Welcome back!  I updated the change wrt comments

PR #30, opened 2017-09-08
ContentAddressableStorage interface/reference

werkt:
Define a ContentAddressableStorage interface that simplifies Map-style
operations to capitalize on addressable Set-like behavior. The reference
implementation maintains a maximum size of content through an
access-based expiration, with all member operations (contains, get, put)
representing an entry access.

### DIFF
--- a/src/main/java/build/buildfarm/common/ContentAddressableStorage.java
+++ b/src/main/java/build/buildfarm/common/ContentAddressableStorage.java
@@ -1,0 +1,75 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.common;
+
+import build.buildfarm.common.ThreadSafety.ThreadSafe;
+import com.google.devtools.remoteexecution.v1test.Digest;
+import com.google.protobuf.ByteString;
+
+public interface ContentAddressableStorage {
+  /**
+   * Blob storage for the CAS. This class should be used at all times when interacting with
+   * complete blobs in order to cut down on independent digest computation.
+   */
+  public static final class Blob {
+    private final Digest digest;
+    private final ByteString data;
+
+    public Blob(ByteString data) {
+      digest = Digests.computeDigest(data);
+      this.data = data;
+    }
+
+    public Digest getDigest() {
+      return digest;
+    }
+
+    public ByteString getData() {
+      return data;
+    }
+
+    public long size() {
+      return digest.getSizeBytes();
+    }
+
+    public boolean isEmpty() {
+      return size() == 0;
+    }
+  }
+
+  /** Indicates presence in the CAS for a digest. */
+  @ThreadSafe
+  boolean contains(Digest digest);
+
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/instance/memory/DelegateCASMap.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import com.google.protobuf.Parser;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+class DelegateCASMap<K,V extends Message> implements Map<K,V>
+{
```
ulfjack: style: put the opening { on the previous line

src/main/java/build/buildfarm/instance/memory/DelegateCASMap.java
```diff
@@ -0,0 +1,121 @@
...
+import com.google.devtools.remoteexecution.v1test.Digest;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import com.google.protobuf.Parser;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+class DelegateCASMap<K,V extends Message> implements Map<K,V>
```
ulfjack: Should this be a ConcurrentMap?

src/main/java/build/buildfarm/common/ContentAddressableStorage.java
```diff
@@ -0,0 +1,37 @@
+package build.buildfarm.common;
+
+import com.google.devtools.remoteexecution.v1test.Digest;
+import com.google.protobuf.ByteString;
+
+public interface ContentAddressableStorage {
```
ulfjack: What about thread-safety?

src/main/java/build/buildfarm/instance/memory/MemoryLRUContentAddressableStorage.java
```diff
@@ -0,0 +1,156 @@
...
+import com.google.devtools.remoteexecution.v1test.Digest;
+import com.google.protobuf.ByteString;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class MemoryLRUContentAddressableStorage implements ContentAddressableStorage {
+  private final long maxSizeInBytes;
+  private final Map<Digest, Entry> storage;
```
ulfjack: You're basically reimplementing LinkedHashMap. Did you know that LinkedHashMap supports being used as an LRU-style cache?

werkt: It does, but it does not facilitate making the expiration decision based on an aggregate value (size in this case, I first tried to make it work as a subclass). There are also a couple of other optimizations that might not be obvious regarding the self-contained nature of the Blob and its keys.  This implementation was influenced heavily by the nature of LinkedHashMap.

src/main/java/build/buildfarm/instance/memory/DelegateCASMap.java
werkt: IMHO making this a ConcurrentMap is required if either of the following cases applies:
1. Only extending from ConcurrentMap connotes to users that this store is thread-safe
2. Specific features from the ConcurrentMap interface are in use (by users) and likely do not represent trivial method implementations

Regarding 1, Since it is a delegator, I'm inclined to say that it is only as thread safe as the ContentAddressableStorage, which as I'll comment below, is fully expected to be thread-safe, but I could see motivations in either direction.  Adding additional fluff methods just to satisfy that interface is not something that motivates me strongly in the direction of 2

src/main/java/build/buildfarm/common/ContentAddressableStorage.java
werkt: I'll add the thread safety annotations to the relevant members

PR #31, opened 2017-09-08
Refactor worker cache, impose size-expiration

werkt:
Use a CAS-fetching file cache supplier of Paths in the worker, which
manages expiration based on a configurable maximum number of bytes. The
file cache refreshes its access time based on each action execution, and
as such is capable of reporting actual input edge traversals, as well as
retaining those file legitimately used most recently - benefit may be
negligible in this case, since inputs must be fetched even if unused,
but may aggregate over distinct actions to be effective. The file cache
will allow temporary oversubscription in this case, when all of the
inputs combined for a single action resolve to be larger than the max,
but it will prevent a single input that is larger than the max from
being inserted.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -1,0 +1,330 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.worker;
+
+import static java.nio.file.LinkOption.NOFOLLOW_LINKS;
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+
+import build.buildfarm.common.Digests;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.devtools.remoteexecution.v1test.Digest;
+import io.grpc.StatusRuntimeException;
+import java.io.IOException;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.io.InputStream;
+import java.util.concurrent.ConcurrentHashMap;
+
+class CASFileCache {
+  private final InputStreamFactory inputStreamFactory;
+  private final Path root;
+  private final long maxSizeInBytes;
+  private final Map<String, Entry> storage;
+
+  private transient long sizeInBytes;
+  private transient Entry header;
+
+  CASFileCache(InputStreamFactory inputStreamFactory, Path root, long maxSizeInBytes) {
+    this.inputStreamFactory = inputStreamFactory;
+    this.root = root;
+    this.maxSizeInBytes = maxSizeInBytes;
+
+    sizeInBytes = 0;
+    header = new SentinelEntry();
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -0,0 +1,238 @@
...
+  public CASFileCache(Instance instance, Path root, long maxSizeInBytes) {
+    this.instance = instance;
+    this.root = root;
+    this.maxSizeInBytes = maxSizeInBytes;
+    paths = new HashMap<>();
+    accessScanPaths = new ArrayList<>();
+    sizeInBytes = 0;
+  }
+
+  public void start() throws IOException {
```
ulfjack: I'd appreciate some documentation here on what this method is intended to do. Maybe also comment on the cost of running this.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -0,0 +1,238 @@
...
+              attrs.lastAccessTime().to(TimeUnit.NANOSECONDS)));
+          sizeInBytes += size;
+        }
+        return FileVisitResult.CONTINUE;
+      }
+    });
+    update();
+  }
+
+  public void update() {
```
ulfjack: Here also.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    @Override
+    public int compareTo(FileEntry entry) {
+      // the most recently accessed file is greater
+      if (accessTime != entry.accessTime) {
+        return entry.accessTime > accessTime ? 1 : -1;
+      }
+      // the smaller file is greater
+      if (sizeInBytes != entry.sizeInBytes) {
```
ulfjack: Is this and the stuff below necessary? If the access time is sufficiently fine-grained, then this is never executed, i.e., dead code.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -0,0 +1,238 @@
...
+    if (sizeInBytes + size > maxSizeInBytes) {
+      System.out.println(String.format(
+          "Expiration did not yield sufficient space: %d + %d > %d",
+          sizeInBytes,
+          size,
+          maxSizeInBytes));
+      return null;
+    }
+
+    try (InputStream in = instance.newStreamInput(instance.getBlobName(digest))) {
```
ulfjack: This seems like dependency inversion. I'd expect a class called CASFileCache to not have a reference back to the Instance, but to work more as a library class.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -0,0 +1,238 @@
...
+      if (copySize != size) {
+        Files.delete(tmpPath);
+        System.out.println(String.format(
+            "Invalid file size copied into cache: was %d, expected %d",
+            copySize,
+            size));
+        return null;
+      }
+      setPermissions(tmpPath, isExecutable);
+      Files.move(tmpPath, path, REPLACE_EXISTING);
```
ulfjack: Do we need to fsync the file before moving?

src/main/java/build/buildfarm/worker/CASFileCache.java
ulfjack: Consider passing in an InputStream or some such. Also, if we're eventually switching to store files on disk in the instance, we might want to not have an intermediate copy here.

werkt: I'll adjust to use an inputStreamFactory instead of the instance to clean up the dependencies.
While I know the origin of `eventually switching to store files on disk in the instance`, I want to head there in reasonable steps.  The only intermediate copy here may eventually be replaced by a Path response from the instance, but let's cross that bridge when we set fire to this one.

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: Not for any posix filesystem I've ever worked with for close->rename sequencing? The sibling construction is purposeful and the resulting syscall generation seems to indicate java's compliance.

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: Function comment added

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: Dropped this in favor of reference counters, hope that name is documentation enough.

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: Dropped the access time stuff entirely

PR #34, opened 2017-09-28
rename all BUILD files' name to BUILD.bazel

ddwolf:
Rename BUILD to BUILD.bazel and fix some errors followed.

[this](https://github.com/bazelbuild/bazel-buildfarm/compare/master...ddwolf:rename_BUILD_files?expand=1#diff-20ca8c7d9af5efe6a1cae9e9883f9151L9) is not referenced anywhere, so I removed it.

And this pr is related with this issue: #27


ulfjack:
Why?

ddwolf:
@ulfjack there are directories named "build" in lowercase and BUILD file in uppercase, they are treated as same objects in windows. And there is also this issue #27 .

ulfjack:
Where are lowercase "build" directories?

ddwolf:
@ulfjack in this directory: [src/main/protobuf/](https://github.com/bazelbuild/bazel-buildfarm/tree/master/src/main/protobuf)

ulfjack:
I don't think #27 is worth fixing, but #28 may require having a BUILD.bazel file?

ddwolf:
well this could be closed for #28 has been fixed. And not all BUILD files have to be renamed, by ulfjack's suggest.

PR #39, opened 2017-10-11
Break circular dependency on BuildFarmService.

abergmeier-dsfishlabs:
Due to SOC, if Services are only interested in Instances, there is no need to make BuildFarmService accessible.
Rather seperate out BuildFarmInstances. This also enables a simpler API later on.


abergmeier-dsfishlabs:
Rebased so it applies cleanly.

abergmeier-dsfishlabs:
Rebased second time.

PR #42, opened 2017-10-12
Extend support for option parsing

abergmeier-dsfishlabs:
Uses pcj/google_options for option parsing (similar to Bazel).
Seems to have not direct positional support - so using residue for that.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Worker.java
+++ b/src/main/java/build/buildfarm/worker/Worker.java
@@ -508,7 +508,7 @@ class Worker {
     List<String> residue = parser.getResidue();
     if (residue.isEmpty()) {
       printUsage(parser);
-      return;
+      throw new IllegalArgumentException("Missing CONFIG_PATH");
     }
     Path configPath = Paths.get(residue.get(0));
     try (InputStream configInputStream = Files.newInputStream(configPath)) {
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -486,17 +489,30 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+  }
+
   public static void main(String[] args) throws Exception {
-    Path configPath = Paths.get(args[0]);
+    OptionsParser parser = OptionsParser.newOptionsParser(WorkerOptions.class);
+    parser.parseAndExitUponError(args);
+    List<String> residue = parser.getResidue();
+    if (residue.isEmpty()) {
+      printUsage(parser);
+      return;
```
werkt: This will exit the server with a good status, likely not a desirable result if a config isn't specified.

abergmeier-dsfishlabs: Well that is IMO a matter of philosophy. Generally I do not care enough either way. So what behavior do you want exactly?

werkt: I want the worker to exit with a bad status when the config argument is omitted.

abergmeier-dsfishlabs: So throw an exception here!? Which?

werkt: Common practice appears to be an `IllegalArgumentException`

abergmeier-dsfishlabs:
All done.

abergmeier-dsfishlabs:
Should be good to merge.

PR #43, opened 2017-10-12
Minor cleanup for Server Options.

abergmeier-dsfishlabs:


abergmeier:
> Can you just merge this in with #40?

Topic wise it does not really fit with #40. I'll find other changes to incorporate these, though.

PR #45, opened 2017-10-12
Handle missing values for root and casCacheDirectory fields.

abergmeier-dsfishlabs:
For both throw ConfigurationException. Previously there was no error but still did not work.
Also added belonging Tests for Worker.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Worker.java
+++ b/src/main/java/build/buildfarm/worker/Worker.java
@@ -19,6 +19,7 @@ import build.buildfarm.instance.Instance;
 import build.buildfarm.instance.stub.StubInstance;
 import build.buildfarm.v1test.WorkerConfig;
 import build.buildfarm.v1test.CASInsertionControl;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -61,8 +62,9 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 import java.util.logging.Logger;
+import javax.naming.ConfigurationException;
 
-class Worker {
+public class Worker {
   public static final Logger logger = Logger.getLogger(Worker.class.getName());
   private final Instance instance;
   private final WorkerConfig config;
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -124,10 +126,30 @@ private static ManagedChannel createChannel(String target) {
...
+    String casCacheValue = config.getCasCacheDirectory();
+    // Have to handle empty string here because all later APIs happily ignore
+    // empty strings (WTF)
+    if (Strings.isNullOrEmpty(casCacheValue)) {
+        throw new ConfigurationException("root value in config missing");
+    }
+    return root.resolve(casCacheValue);
+  }
+
+  public Worker(WorkerConfig config) throws ConfigurationException{
```
werkt: Whitespace before {

src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -124,10 +126,30 @@ private static ManagedChannel createChannel(String target) {
...
+    if (Strings.isNullOrEmpty(rootValue)) {
+        throw new ConfigurationException("root value in config missing");
+    }
+    return Paths.get(rootValue);
+  }
+
+  private static Path getValidCasCacheDirectory(WorkerConfig config, Path root) throws ConfigurationException {
+    String casCacheValue = config.getCasCacheDirectory();
+    // Have to handle empty string here because all later APIs happily ignore
+    // empty strings (WTF)
```
werkt: Can we clean up this comment and the one above? Also, APIs for what? Path#resolve?

src/test/java/build/buildfarm/BuildFarmWorkerTest.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Test(expected = ConfigurationException.class)
+  public void missingCasCacheDirectory() throws ConfigurationException {
+    WorkerConfig.Builder builder = WorkerConfig.newBuilder();
+    builder.setRoot("/");
+    new Worker(builder.build());
+  }
+}
+
```
werkt: Trailing empty line

src/main/java/build/buildfarm/worker/Worker.java
abergmeier-dsfishlabs: Not just `Path` resolve. Literally everything that this data gets used on totally ignores empty strings. But you are right - since we have the test, we may also remove the comment, since the Test should catch regressions.

src/test/java/build/buildfarm/BuildFarmWorkerTest.java
abergmeier-dsfishlabs: Would like to keep that to be in git's good graces.

werkt: Trailing newline is not the same thing as trailing empty line, but w/e

PR #47, opened 2017-10-12
Handle unknown instance in OperationQueueService.take.

abergmeier-dsfishlabs:
Error message should now point to the actual problem.


abergmeier-dsfishlabs:
Right - just would like to keep the actual error message.

PR #48, opened 2017-10-12
Update protobuf to 3.4.1

vladmos:
Starting from Bazel 0.8 the += operator applied to lists will mutate them instead of copying. This is an incompatible behavior and repostories should be fixed in advance.

To reproduce the future behavior one can use the following flag:

    bazel build --incompatible_list_plus_equals_inplace //...

The currently used version of protobuf (3.2.0) is not compatible with this behavior and therefore should be updated.


vladmos:
The diff is quite large, but I just updated protobuf by following the instructions at `third_party/protobuf/README.md` (and updated the instructions to match [Bazel instructions](https://github.com/bazelbuild/bazel/tree/master/third_party/protobuf), just removed the 5th step as it seems to be not necessary here).

PR #49, opened 2017-10-12
Throw exception on missing instance

werkt:
This will provide more comprehensive coverage than #47 in guaranteeing no unavailable instance names are used.


src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -96,7 +96,14 @@ private void readBlob(
       StreamObserver<ReadResponse> responseObserver) {
     String resourceName = request.getResourceName();
 
-    Instance instance = server.getInstanceFromBlob(resourceName);
+    Instance instance;
+    try {
+      instance = server.getInstanceFromBlob(resourceName);
+    } catch (InstanceNotFoundException ex) {
+      responseObserver.onError(new StatusException(Status.NOT_FOUND));
```
abergmeier-dsfishlabs: Can you please add a message what was not found? This seems a tad bit generic.

PR #51, opened 2017-10-13
Prevent race condition with done jobs on watch

werkt:


src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -345,13 +345,25 @@ public boolean watchOperation(
         return true;
       }
     }
+    Operation completedOperation = null;
     synchronized(watchers) {
       List<Function<Operation, Boolean>> operationWatchers = watchers.get(operationName);
       if (operationWatchers == null) {
-        return false;
+        /* we can race on the synchronization and miss a done, where the
```
abergmeier-dsfishlabs: Why is there another access to `watchers`, without a synchronized block to begin with!?

werkt: This is the only access to `watchers` in this function, not sure what 'another access' refers to.  If you mean 'why not guard the larger section including the above checks', I didn't want to introduce nested locking without establishing a rigorous ordering (`getOperation` locks `operationLock(operationName)`).

This guard actually *always* has to take place, since we may be being called with `watchInitialState = false`, allowing us to get here without a valid outstanding or completed operation.

abergmeier: I rather meant why there is no lock in line 189.

werkt: Gotcha.  Locking the watchers is a bit confusing:

`watchers` is by definition (if not by interface, there was some discussion about that) a ConcurrentHashMap in the MemoryInstance
Being a ConcurrentHashMap means that insertion, is completely thread isolated, as in line 189.
The name of an operation, used as the key to watchers, is compelled through `UUID.randomUUID()` generation to be unique and novel at line 189, preventing possible key collisions.
The `synchronized(watchers)` blocks that are scattered throughout are meant to guard the following rule: `watchers` values (lists of callbacks) must be protected against concurrent accesses

abergmeier-dsfishlabs: Ok, thanks for the explanation (perhaps move some of that into comments in code!?).

PR #52, opened 2017-10-13
Reduce area under exception guard for putAllBlobs

werkt:


src/main/java/build/buildfarm/server/ContentAddressableStorageService.java
```diff
@@ -98,15 +98,15 @@ public void batchUpdateBlobs(
...
     try {
       instance.putAllBlobs(validBlobsBuilder.build());
-      responseObserver.onNext(BatchUpdateBlobsResponse.newBuilder()
-          .addAllResponses(responses.build())
-          .build());
-      responseObserver.onCompleted();
-    } catch (InterruptedException ex) {
-      responseObserver.onError(new StatusException(Status.fromThrowable(ex)));
-    } catch (IOException ex) {
+    } catch (InterruptedException|IOException ex) {
```
ulfjack: Catching InterruptedException is frowned upon. This can cause the server to not exit if interrupted with ctrl-c.

werkt: This was added because the ByteStreamUploader#uploadBlobs method explicitly caught and declared the thread interruption, due to the InterruptedException declared thrown from ListenableFuture<>#get.  The ByteStreamUploader module was lifted directly from bazel, and calls Thread.interrupted() on its catch, with the (clear?) intent of resetting the interrupted state.

I cannot override batchUpdateBlobs from the CAS service interface with a declared throw of InterruptedException. This forces me into catching InterruptedException here, and I figured why not indicate an error to the client rather than a (maybe) transport shutdown.  Would you prefer a Thread.interrupt() to the return?

ulfjack: What do you want to happen if someone sends SIGINT to the server?

Presumably, you want the server to shut down. However, that means that any code that calls a method that throws InterruptedException has to make sure that the interrupted status doesn't get lost. Right?

abergmeier-dsfishlabs: > and calls Thread.interrupted() on its catch, with the (clear?) intent of resetting the interrupted state.

@werkt I think the previous handling of `InterruptedException` was better. It started to gracefully give an error back from the interrupt. Personally I would prefer an error message, that clearly indicates that the server was interrupted. And after that, it probably should rethrow the exception @ulfjack?

werkt: InterruptedException now only interrupts (and would return if it did not break execution)

ola-rozenfeld: Agreed with Ulf, an interrupt() call should be here; however, I see no harm in having both responseObserver.onError() and interrupt(). You could notify the user of the problem before exiting gracefully. This will simplify debugging for the user, when it happens. That's what I did in the RemoteWorker in bazel, too.

werkt: Updated with onError from the throwable InterruptedException

PR #53, opened 2017-10-17
Extend generic NOT_FOUND error with actual error message.

abergmeier-dsfishlabs:
- Expose InstanceNotFoundException.instanceName publicly

### DIFF
--- a/src/main/java/build/buildfarm/server/ActionCacheService.java
+++ b/src/main/java/build/buildfarm/server/ActionCacheService.java
@@ -38,7 +38,7 @@ public class ActionCacheService extends ActionCacheGrpc.ActionCacheImplBase {
     try {
       instance = instances.getInstance(request.getInstanceName());
     } catch (InstanceNotFoundException ex) {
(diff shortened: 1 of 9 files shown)


abergmeier-dsfishlabs:
Rebased

src/main/java/build/buildfarm/server/ActionCacheService.java
```diff
@@ -38,7 +38,7 @@ public void getActionResult(
     try {
       instance = instances.getInstance(request.getInstanceName());
     } catch (InstanceNotFoundException ex) {
-      responseObserver.onError(new StatusException(Status.NOT_FOUND));
```
ulfjack: My personal preference is to do something like this:
responseObserver.onError(BuildFarmInstances.exceptionToError(ex))

Passing in the full observer means that the method could do all sorts of weird shenanigans, and is also more difficult to test.

abergmeier-dsfishlabs: Done.

PR #55, opened 2017-10-17
Explicitly assert for UUID uniqueness in watchers.

abergmeier-dsfishlabs:
Previously this was implicitly assumed.

### DIFF
--- a/src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
+++ b/src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
@@ -186,7 +186,8 @@ public class MemoryInstance extends AbstractServerInstance {
   protected Operation createOperation(Action action) {
     String name = createOperationName(UUID.randomUUID().toString());
 
-    watchers.put(name, new ArrayList<Function<Operation, Boolean>>());
+    List<Function<Operation, Boolean>> previous = watchers.put(name, new ArrayList<Function<Operation, Boolean>>());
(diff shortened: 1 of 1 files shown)


werkt:
Requesting a review from @ulfjack: the style and consistency of the assert, plus the fact that we're now asserting over UUID's uniqueness, which would seem more Jepsen's territory, makes me want his recommendation.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -186,7 +186,8 @@ protected void updateOperationWatchers(Operation operation) {
   protected Operation createOperation(Action action) {
     String name = createOperationName(UUID.randomUUID().toString());
 
-    watchers.put(name, new ArrayList<Function<Operation, Boolean>>());
+    List<Function<Operation, Boolean>> previous = watchers.put(name, new ArrayList<Function<Operation, Boolean>>());
+    assert previous != null : "Operation name " + previous + " expected to be unique - but seems like there were some collisions";
```
ulfjack: Assert is frowned upon by Google style, since it's usually not enabled - prefer Preconditions.checkState or Asserts.checkState (I think it's called Asserts - the class in guava).

This assert actually looks incorrect - don't you want to assert previous == null?

werkt:
This requires some changes to go in, Andi. Looking to merge this or it will be closed no sooner than Jan 25. If we close, I'll try to get around to putting it up myself with the modifications.

PR #56, opened 2017-10-18
Use typeCase reference.

abergmeier-dsfishlabs:


abergmeier-dsfishlabs:
@hhclam Do you want to review?

hhclam:
Sorry! Didn't check emails for a couple day. Looks like this is merge already.

PR #58, opened 2017-10-18
Enable passing root and cas_cache_directory via commandline.

abergmeier-dsfishlabs:


src/main/java/build/buildfarm/worker/WorkerOptions.java
```diff
@@ -29,4 +29,19 @@
...
+  @Option(
+      name = "root",
+      help = "Root base directory for all work being performed.",
+      defaultValue = ""
+    )
+  public String root;
+
+  @Option(
+      name = "cas_cache_directory",
+      help = "(Absolute or relative) path to cached files from CAS.",
```
werkt: `(Absolute or relative to root)`

abergmeier-dsfishlabs: Done.

PR #61, opened 2017-10-18
Add description for every INVALID_ARGUMENT in ByteStreamService.

abergmeier-dsfishlabs:
Without that debugging errors is a PITA.

### DIFF
--- a/src/main/java/build/buildfarm/server/ByteStreamService.java
+++ b/src/main/java/build/buildfarm/server/ByteStreamService.java
@@ -234,7 +234,7 @@ public class ByteStreamService extends ByteStreamGrpc.ByteStreamImplBase {
         if (data == null) {
           digest = parseUploadBlobDigest(writeResourceName);
           if (digest == null) {
-            String description = "Could not extract digest off: " + writeResourceName;
+            String description = "Could not parse digest of: " + writeResourceName;
             responseObserver.onError(new StatusException(Status.INVALID_ARGUMENT.withDescription(description)));
             failed = true;
             return;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -232,7 +234,8 @@ private void writeBlob(
         if (data == null) {
           digest = parseUploadBlobDigest(writeResourceName);
           if (digest == null) {
-            responseObserver.onError(new StatusException(Status.INVALID_ARGUMENT));
+            String description = "Could not extract digest off: " + writeResourceName;
```
werkt: `Could not parse digest of:`

src/main/java/build/buildfarm/server/ByteStreamService.java
```diff
@@ -259,7 +263,8 @@ private void writeBlob(
           active_write_requests.remove(writeResourceName);
           Digest blobDigest = Digests.computeDigest(data);
           if (!blobDigest.equals(digest)) {
-            responseObserver.onError(new StatusException(Status.INVALID_ARGUMENT));
+            String description = String.format("Digest mismatch %s <-> %s", blobDigest.getHash(), digest.getHash());
```
werkt: Prefer `Digests.toString(blobDigest)` etc here - the equals comparison occurs between both hash and size, and we have the toString mechanism representing all other presentations of digests

abergmeier-dsfishlabs:
Done.

werkt:
The web editor, which is capable of cleaning up this merge, creates commits that have Author lines that the googlebot cla resolver cannot identify as me (with a signed CLA). I cannot commit directly to @abergmeier-dsfishlabs' branch via the commandline, so I need to close this to get the merge in place. That PR will also require a review.

werkt:
Recheck cla

PR #62, opened 2017-10-19
Introduced initial ByteStreamServiceTest

abergmeier-dsfishlabs:
This might need some more work. Need someone with Java experience to look at this.

### DIFF
--- a/src/test/java/build/buildfarm/ByteStreamServiceTest.java
+++ b/src/test/java/build/buildfarm/ByteStreamServiceTest.java
@@ -34,8 +34,8 @@ import static org.mockito.Mockito.verify;
 @RunWith(MockitoJUnitRunner.class)
 public class ByteStreamServiceTest {
 
-  private static byte[] generateFileData(Random random) {
-    byte[] data = new byte[1024 * 1024];
+  private static byte[] generateFileData(Random random, int length) {
+    byte[] data = new byte[length];
     random.nextBytes(data);
     return data;
   }
@@ -55,7 +55,7 @@ public class ByteStreamServiceTest {
     String uuid = "1d4e65cd-7c0d-44ca-b49b-b912988c8967"; // dummy value
 
     Random random = new Random(49);
-    byte[] data = generateFileData(random);
+    byte[] data = generateFileData(random, 1024 * 1024);
     ByteString bytes = ByteString.copyFrom(data);
     Digest digest = Digests.computeDigest(bytes);
     String hash = digest.toString();
--- a/src/test/java/build/buildfarm/ByteStreamServiceTest.java
+++ b/src/test/java/build/buildfarm/ByteStreamServiceTest.java
@@ -1,0 +1,76 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/test/java/build/buildfarm/ByteStreamServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ByteStreamServiceTest {
+
+  private static byte[] generateFileData(Random random) {
```
ulfjack: I'd pass in the length of the data to randomly generate.

src/test/java/build/buildfarm/ByteStreamServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    Random random = new Random(49);
+    byte[] data = generateFileData(random);
+    ByteString bytes = ByteString.copyFrom(data);
+    Digest digest = Digests.computeDigest(bytes);
+    String hash = digest.toString();
+    int size = data.length;
+
+    WriteRequest request = WriteRequest.newBuilder()
+      .setResourceName(String.format("%s/uploads/%s/blobs/%s/%d", instance, uuid, hash, size))
```
ulfjack: Google style is four spaces indentation on continuation lines

abergmeier-dsfishlabs: Ok, but then we should really add a pre-commit hook to the repo. I guess you have one floating around at Google?

buchgr:
@abergmeier-dsfishlabs 

gRPC provides some guidance and code examples on how to do unit testing. https://github.com/grpc/grpc-java/tree/master/examples#unit-test-examples

As of recently, they also provide a junit rule!

src/test/java/build/buildfarm/ByteStreamServiceTest.java
werkt: Agreed, and this also warrants a squash with the other commit for the addition of this test class.

abergmeier-dsfishlabs: > I'd pass in the length of the data to randomly generate.

Long done.

Want to get the format hook in before continuing. Depends on upstream google-java-format so might take a while.

werkt:
Similar to #55, looking to merge or close this out by Jan 25.

src/test/java/build/buildfarm/ByteStreamServiceTest.java
ttsugriy: @abergmeier-dsfishlabs, I highly recommend using https://github.com/google/google-java-format which automatically formats the code using Google style. Ideally this should indeed be configured as part of presubmit hook or some other automated way.

werkt: @ttsugriy I believe we can count Andreas out of maintaining this.  If you'd like to pick up the mantle and carry it forward, I will review in turn. Otherwise I will close this when I get finished with larger scale code coverage-based test additions.

werkt:
Closing this out in favor of following the grpc unit testing guidance.

PR #63, opened 2017-10-23
Refactor worker into staged pipeline

werkt:
Move the major operations - Input Fetch, Execute, Writeback, of the
worker into a pipeline with concurrent stage execution. The number of
available concurrent slots for the execution stage is configurable, and
each stage will stall on a transition to the next if it is (fully)
occupied.

### DIFF
--- a/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
+++ b/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
@@ -1,0 +1,78 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.worker;
+
+import com.google.common.collect.Iterables;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+
+class ExecuteActionStage extends PipelineStage {
+  private final Set<Thread> executors;
+  private BlockingQueue<OperationContext> queue;
+
+  ExecuteActionStage(Worker worker, PipelineStage output, PipelineStage error) {
+    super(worker, output, error);
+    queue = new ArrayBlockingQueue<>(1);
+    executors = new HashSet<>();
+  }
+
+  @Override
+  public void offer(OperationContext operationContext) {
+    queue.offer(operationContext);
+  }
+
+  @Override
+  public OperationContext take() throws InterruptedException {
+    return queue.take();
+  }
+
+  @Override
+  public synchronized boolean claim() throws InterruptedException {
+    if (isClosed()) {
+      return false;
+    }
+
+    while (executors.size() >= worker.config.getExecuteStageWidth()) {
+      wait();
+    }
+    return true;
+  }
+
+  @Override
+  public synchronized void release() {
+    if (!executors.remove(Thread.currentThread())) {
+      throw new IllegalStateException();
+    }
+    this.notify();
+  }
+
+  @Override
+  protected boolean isClaimed() {
+    return Iterables.any(executors, (executor) -> executor.isAlive());
+  }
+
+  @Override
+  protected void iterate() throws InterruptedException {
+    Thread executor = new Thread(new Executor(worker, take(), this));
+
+    synchronized(this) {
+      executors.add(executor);
+    }
+
+    executor.start();
+  }
+}
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -1,0 +1,223 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.worker;
+
+import build.buildfarm.v1test.CASInsertionControl;
+import com.google.common.io.ByteStreams;
+import com.google.devtools.remoteexecution.v1test.ActionResult;
+import com.google.devtools.remoteexecution.v1test.Command;
+import com.google.devtools.remoteexecution.v1test.ExecuteOperationMetadata;
+import com.google.devtools.remoteexecution.v1test.ExecuteResponse;
+import com.google.longrunning.Operation;
+import com.google.protobuf.Any;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Duration;
+import com.google.protobuf.InvalidProtocolBufferException;
+import java.nio.file.Path;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+class Executor implements Runnable {
+  private final Worker worker;
+  private final OperationContext operationContext;
+  private final ExecuteActionStage owner;
+
+  private static final OutputStream nullOutputStream = ByteStreams.nullOutputStream();
+
+  Executor(Worker worker, OperationContext operationContext, ExecuteActionStage owner) {
+    this.worker = worker;
+    this.operationContext = operationContext;
+    this.owner = owner;
+  }
+
+  @Override
+  public void run() {
+    Command command;
+    try {
+      command = Command.parseFrom(worker.instance.getBlob(operationContext.action.getCommandDigest()));
+    } catch (InvalidProtocolBufferException ex) {
+      owner.error().offer(operationContext);
+      owner.release();
+      return;
+    }
+
+    ExecuteOperationMetadata executingMetadata = operationContext.metadata.toBuilder()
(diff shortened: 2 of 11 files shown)


ola-rozenfeld:
I just returned from my vacation, will do my best to review this today.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public PipelineStage output() {
+    return this.output;
+  }
+
+  @Override
+  public PipelineStage error() {
+    return this.output;
```
ola-rozenfeld: this.error, presumably?

src/main/java/build/buildfarm/worker/AbstractPipelineStage.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ArrayBlockingQueue;
+
+abstract class AbstractPipelineStage implements PipelineStage {
+  private final PipelineStage output;
+  private final PipelineStage error;
+  private final BlockingQueue<OperationContext> queue;
+
+  private PipelineStage input;
```
ola-rozenfeld: It looks like you need all these variables in subclasses, so maybe add protected accessors?

ola-rozenfeld:
Err, I did not realize this sends you comments right away -- sorry, my
first time using this review tool. My intent was to write a lot of drafts
and send them all at once.

werkt:
No problem, ulf had the same reaction - not sure if there's any other way with github review.  I can deal with it like this, or if you figure out how to have a bulk review come in, let me know how!

src/main/java/build/buildfarm/worker/AbstractPipelineStage.java
werkt: There's no usage of these, but that highlights that these are completely unnecessary except for Executor->ExecuteActionStage communication.  I've updated the diff to clean this up.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: removed code, good catch though!

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -0,0 +1,182 @@
...
+    try {
+      resultBuilder = operationContext
+          .operation.getResponse().unpack(ExecuteResponse.class).getResult().toBuilder();
+    } catch (InvalidProtocolBufferException ex) {
+      poller.stop();
+      return null;
+    }
+
+    ImmutableList.Builder<ByteString> contents = new ImmutableList.Builder<>();
+    if (resultBuilder.getExitCode() == 0) {
```
ola-rozenfeld: Wait, are you only uploading artifacts to CAS on success? Then, how will the user see logs of failed tests, for example?

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -0,0 +1,182 @@
...
+        worker.instance.putAllBlobs(contents.build());
+      } catch (IOException ex) {
+      } catch (InterruptedException ex) {
+        poller.stop();
+        return null;
+      }
+    }
+
+    ActionResult result = resultBuilder.build();
+    if (!operationContext.action.getDoNotCache()) {
```
ola-rozenfeld: and if exit code is 0, right? You don't want to put failed actions into the cache.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -0,0 +1,244 @@
...
+    ByteString stdoutRaw = stdoutReader.getData();
+    if (stdoutRaw.size() > 0) {
+      CASInsertionControl control = worker.config.getStdoutCasControl();
+      boolean withinLimit = stdoutRaw.size() <= control.getLimit();
+      if (withinLimit) {
+        resultBuilder.setStdoutRaw(stdoutRaw);
+      }
+      if (control.getPolicy() == CASInsertionControl.Policy.ALWAYS_INSERT ||
+          (!withinLimit && control.getPolicy() == CASInsertionControl.Policy.INSERT_ABOVE_LIMIT)) {
+        resultBuilder.setStdoutDigest(worker.instance.putBlob(stdoutRaw));
```
ola-rozenfeld: So I guess this is an exception to the stage separation concept -- you upload stuff here from the "wrong" stage? I guess you could separate the stages completely, but then you'd need to save the stdout/err as "special" files, which might not be worth it, or pass them around in some other way.

src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -161,323 +101,27 @@ public void start() {
...
+    executeActionStage.setInput(inputFetchStage);
+    PipelineStage matchStage = new MatchStage(this, inputFetchStage, errorStage);
+    inputFetchStage.setInput(matchStage);
+
+    Pipeline pipeline = new Pipeline();
+    // pipeline.add(errorStage, 0);
+    pipeline.add(matchStage, 1);
+    pipeline.add(inputFetchStage, 2);
+    pipeline.add(executeActionStage, 3);
+    pipeline.add(reportResultStage, 4);
```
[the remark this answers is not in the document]
werkt: Yes, exactly. From the pipeline standpoint, it doesn't make any sense for a stage to be running if there is nowhere for its output to go.

src/main/java/build/buildfarm/worker/Pipeline.java
```diff
@@ -0,0 +1,96 @@
...
+          System.out.println("Closing stage at priority " + maxPriority);
+          stageToClose.close();
+        }
+      }
+      for (Map.Entry<PipelineStage, Thread> stageThread : stageThreads.entrySet()) {
+        PipelineStage stage = stageThread.getKey();
+        Thread thread = stageThread.getValue();
+        try {
+          thread.join(closeStage ? 1 : 1000);
+        } catch (InterruptedException ex) {
```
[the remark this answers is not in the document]
werkt: You have prompted me to investigate fully the ramifications of this.  According to the docs, and the observed behavior of Java, `InterruptedException` is thrown when the *current* thread is interrupted, which is the one calling .join (and not the object it is calling it on). It sounds an awful lot in your comment here that you're under the impression that `InterruptedException` is thrown on `Thread#join(...)` when the thread being joined with has been interrupted... (Can supply example with this exhaustedly proven false)

I'm on board with throwing an InterruptedException if we get same during this process, however I definitely don't want to bail out early, leaving main/the system exit to sit with unstopped attached threads (and hang forever).

src/main/java/build/buildfarm/worker/AbstractPipelineStage.java
```diff
@@ -0,0 +1,108 @@
...
+      return false;
+    }
+    claimed = true;
+    return true;
+  }
+
+  @Override
+  public synchronized void release() {
+    claimed = false;
+    notify();
```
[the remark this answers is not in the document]
werkt: This will not notify all threads, it will only notify one (the notifyAll method satisfies the all case). The notify is unspecified, which makes it a this-call, on the stage object, which makes that object itself a 'per-stage lock object' (that also happens to be the 'per-stage object')

src/main/java/build/buildfarm/worker/AbstractPipelineStage.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public void close() {
```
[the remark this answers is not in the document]
werkt: Correct.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -0,0 +1,115 @@
...
+
+    while (executors.size() >= worker.config.getExecuteStageWidth()) {
+      wait();
+    }
+    return true;
+  }
+
+  @Override
+  public synchronized void release() {
+    if (!executors.remove(Thread.currentThread())) {
```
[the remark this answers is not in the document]
werkt: Yes, that is exactly the pattern.

src/main/java/build/buildfarm/worker/Executor.java
werkt: This is definitely an issue - I will have to transport these outputs through the context to the report result stage, good catch.

src/main/java/build/buildfarm/worker/Pipeline.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if (!thread.isAlive()) {
+          System.out.println("Stage has exited at priority " + stageClosePriorities.get(stage));
+          if (!stage.isClosed()) {
```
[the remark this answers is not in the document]
werkt: Yes, you're definitely right - this was either vestigial or paranoid. I would say that interrupted is not the only way for a stage thread to die, but in the normal sequence of operations, this should be sufficient.

src/main/java/build/buildfarm/worker/ReportResultStage.java
werkt: I was. This appears to be incorrect as we've discussed offline: regardless of exit status, all output files (that exist) should be read from and uploaded into the CAS at the completion of an action execution. Will adjust

src/main/java/build/buildfarm/worker/ReportResultStage.java
werkt: Likewise added this test. Mentioned that we may want to make this configurable, in cases where we want to trust the failure.  The bazel-supported configuration will only have this happen on success, correct.

src/main/java/build/buildfarm/worker/Pipeline.java
```diff
@@ -0,0 +1,96 @@
...
+            System.out.println("Closing stage after exit at priority " + stageClosePriorities.get(stage));
+            stage.close();
+          }
+          inactiveStages.add(stage);
+        } else if (stage.isClosed()) {
+          System.out.println("Interrupting unterminated closed thread at priority " + stageClosePriorities.get(stage));
```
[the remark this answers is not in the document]
werkt: Yes, if they haven't reacted to being closed yet (i.e. are still in the middle of matching/fetching/executing/reporting)

werkt:
I think I've addressed all points except for the refactor of AbstractPipelineStage, just because I'm trying to normalize the behaviors of the Abstract, Match, and Execute stages (specifically their #iterate facsimiles) for a common interface. I won't leave it hanging, but need another day or so.

ola-rozenfeld:
Is this ready for another pass?

werkt:
> Is this ready for another pass?

Yes, please take a look.  AbstractPipelineStage is now PipelineStage.

PR #65, opened 2017-10-24
Extract all url path handling code to new UrlPath class

abergmeier-dsfishlabs:
Remove is* methods since implementation is inefficient and repetitive.
Instead introduce a UrlPath.detectServiceType, which is only a bit inefficient.


abergmeier-dsfishlabs:
@hhclam You want to review?

hhclam:
LGTM.

abergmeier:
Personally I used a `enum` because this should be fairly easy for the VM to inline. With an interface I am somewhat sceptical on that. If we do not care about perf I'll make the change.

abergmeier-dsfishlabs:
@werkt I think `UrlPath.ResourceOperation` should be fine now. Looking at its usage, I do not see the benefit of introducing more object orientation.
Regardless I think you can introduce a package level `ResourceOperation` interface just fine and have all Url handling be in `UrlPath`. I think that would make sense anyway. But that is then probably best introduced by you on top of this PR.

PR #67, opened 2017-12-09
Bring bazel-buildfarm repo up to date with API changes and third_party management.

philwo:
@buchgr, @werkt, could you please have a look at my changes and see whether they make sense?

I think using bazel-deps instead of checking in stuff is preferred these days due to automatic version and dependency management and makes the repo leaner, as we don't carry all these third-party files.

I could also get rid of some unused deps and files, e.g. the entire `tools` folder seemed unused.

While I was at it, I updated the code to the newest API proto changes.

Builds for me:

```bash
philwo@philwo-imac ~/src/bazel-buildfarm (master) bazel build //src/...
DEBUG: /Users/philwo/src/bazel-buildfarm/third_party/grpc_java/java_grpc_library.bzl:101:5: Multiple values in 'deps' is deprecated in google_devtools_remoteexecution_v1test_remote_execution_java_grpc
INFO: Analysed 16 targets (1 packages loaded).
INFO: Found 16 targets...
INFO: Elapsed time: 0.288s, Critical Path: 0.00s
INFO: Build completed successfully, 1 total action
```

Haven't tested anything else, but apart from the (necessary) semantic changes due to little changes in the API protos and bumping third party dependency versions quite a bit, this should be fine :)


werkt:
Why are there separate `3rdparty` and `third_party` paths as suggested by this change?

third_party/grpc_java/java_grpc_library.bzl
```diff
@@ -0,0 +1,132 @@
...
+    flavor: (str) "normal" (default) for normal proto runtime. "lite"
+        for the lite runtime.
+    visibility: (list) the visibility list
+    **kwargs: Passed through to generated targets
+  """
+  if flavor == None:
+    flavor = "normal"
+
+  if len(deps) > 1:
+    print("Multiple values in 'deps' is deprecated in " + name)
```
werkt: The upgrade to this implementation of java_grpc_library generates the following DEBUG message indicating deprecation from `BUILD.googleapis`:

DEBUG: $repo_root/bazel-buildfarm/third_party/grpc_java/java_grpc_library.bzl:101:5: Multiple values in 'deps' is deprecated in google_devtools_remoteexecution_v1test_remote_execution_java_grpc

Would prefer to not regress like this in this explicit upgrade.

werkt: I see that we cannot simply remove the longrunning operations dependency due to references within the remote execution protobuf definition, but I'm confused as to why the grpc authors believe this to be a valid mechanism to deprecate.

werkt:
There's a comment on pgr0ss/bazel-deps which states that the project has been deprecated. Among the many questions this raises about the decision to use bazel-deps to manage our dependencies for me, is the bazelbuild org going to take on hosting/maintaining a forked release of this tool?

PR #70, opened 2017-12-14
Fixing my faulty merge.

werkt:
Needed to move the description declaration into the default block


werkt:
They aren't, we need to put some CI in place for this repo/branch

PR #81, opened 2018-01-18
Configurize Instance Digest Function

werkt:
Utilize modified variant of bazel's DigestUtil for providing a hash
function in instance configuration. Since this relates to the routing/
common features across instances, it will be exposed to the multiplexing
server layer.


src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -46,57 +112,51 @@ public InputStream openStream() throws IOException {
...
-    private ActionKey(Digest digest) {
-      this.digest = digest;
-    }
+  public Digest compute(Message message) {
+    return compute(message.toByteString());
   }
 
-  public static ActionKey computeActionKey(Action action) {
-    return new ActionKey(computeDigest(action));
+  public DigestUtil.ActionKey computeActionKey(Action action) {
```
ola-rozenfeld: Why not just ActionKey?

src/main/java/build/buildfarm/worker/Worker.java
```diff
@@ -70,17 +72,30 @@ private static Path getValidCasCacheDirectory(WorkerConfig config, Path root) th
     return root.resolve(casCacheValue);
   }
 
+  private static HashFunction getValidHashFunction(WorkerConfig config) throws ConfigurationException {
+    switch (config.getHashFunction()) {
```
ola-rozenfeld: One way to not duplicate this would be to add a cast constructor to the DigestUtil.HashFunction enum that takes this proto enum.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -82,11 +82,18 @@ message MemoryInstanceConfig {
   int64 cas_max_size_bytes = 7;
 }
 
+enum InstanceHashFunction {
```
ola-rozenfeld: So the reason this exists for the workers and the instances is that the workers have to know this and we don't yet have the capabilities API, right? (We were thinking for a while to add it, in particular the "what's my hash function" call).

src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -11,26 +11,92 @@
...
+    public com.google.common.hash.HashFunction getHash() {
+      return hash;
+    }
+
+    public boolean isValidDigest(byte[] digest) {
+      return digest != null && digest.length * 8 == hash.bits();
+    }
+
+    public boolean isValidHexDigest(String hexDigest) {
+      return hexDigest != null && hexDigest.length() * 8 / 2 == hash.bits();
```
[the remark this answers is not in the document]
werkt: This was actually lifted from bazel itself:

https://github.com/bazelbuild/bazel/blob/93b329417bba4ac4860628ebefc9fc64faec8cdf/src/main/java/com/google/devtools/build/lib/vfs/FileSystem.java#L68

I don't mind the extra validity check, I was a little surprised that vfs.FileSystem.HashFunction does.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
werkt: Sure! Also I had no idea that a capabilities API was in the works.

src/main/java/build/buildfarm/common/DigestUtil.java
werkt: why not indeed... broadening this to actually make the Instance defined in terms of it.

src/main/java/build/buildfarm/worker/Worker.java
werkt: Done, though with a static getter, since we already constructed them...

PR #82, opened 2018-01-18
Configurable symlink of input-only directories

werkt:
Provide the option of symlinking directories which are present in the
input tree and have no intersection with output file paths. This results
in substantial throughput improvement for the IF->EX transition,
allowing extremely large input trees to be coherently factored out of
the input fetch stage. Invalidation of expired content of a directory
occurs recursively in the CASFileCache.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -44,7 +44,7 @@ class CASFileCache {
   private final Path root;
   private final long maxSizeInBytes;
   private final DigestUtil digestUtil;
-  private final Map<String, Entry> storage;
+  private final Map<Path, Entry> storage;
   private final Map<Digest, DirectoryEntry> directoryStorage;
 
   private transient long sizeInBytes;
@@ -74,7 +74,7 @@ class CASFileCache {
     Files.createDirectories(root);
 
     Files.walkFileTree(root, new SimpleFileVisitor<Path>() {
-      private String parseFileEntryKey(String fileName, long size) {
+      private Path parseFileEntryKey(String fileName, long size) {
         String[] components = fileName.toString().split("_");
         if (components.length < 2 || components.length > 3) {
           return null;
@@ -112,7 +112,7 @@ class CASFileCache {
         if (sizeInBytes + size > maxSizeInBytes) {
           Files.delete(file);
         } else {
-          String key = null;
+          Path key = null;
           if (file.getParent().equals(root)) {
             key = parseFileEntryKey(file.getFileName().toString(), size);
           }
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -156,32 +178,114 @@ public synchronized void update(Iterable<String> inputs) {
...
+    fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
+
+    e = new DirectoryEntry(directoriesIndex.get(digest), inputsBuilder.build());
+
+    directoryStorage.put(digest, e);
+
+    return path;
+  }
+
   public String put(Digest digest, boolean isExecutable) throws InterruptedException {
```
ola-rozenfeld: I'm curious why you sometimes use String and other times Path as a result of put. IIUC, this is always the local "true" path of the file or directory, right?

src/main/java/build/buildfarm/worker/InputFetchStage.java
```diff
@@ -143,12 +200,27 @@ private void linkInputs(
...
+        Files.createDirectories(dirPath);
+        linkInputs(dirPath, digest, directoriesIndex, childOutputDirectory, inputFiles, inputDirectories);
+      } else {
+        inputDirectories.add(digest);
+        linkDirectory(dirPath, digest, directoriesIndex);
+      }
     }
   }
 
+  private void linkDirectory(
```
ola-rozenfeld: Do I understand correctly that you are symlinking only the direct children of the root directory, in full, never recursively? I mean, if input of action A is:
a/b/c/d/foo.txt
and input of action B is:
a/x/b/c/d/foo.txt
then there will be no reuse of the b/c/d/foo.txt subtree? If so, is that because you think it would not be as frequently useful?

werkt: You are correct, and it is not that the net result of recursive subtrees would be unuseful, but that none of my (freely available open-source versions of) tools were able to function with multiple levels of symlink subdirectories (too much realpath() usage). Even if the base resolution of libraries worked for the loader, runtimes would typically fail, and the top-level directory based on how bazel functions and the nature of the highest level yielded strong performance improvements (since only an intersection with outputs will prevent symlinking).

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: I think that was just a result of trying to avoid creating additional objects, but it is likely moot or better (certainly simplifies some things) when compared to storing Paths. I just tried it out here, seems legit - posting change

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -156,32 +178,114 @@ public synchronized void update(Iterable<String> inputs) {
...
+  private Path getDirectoryPath(Digest digest) {
+    return root.resolve(digestFilename(digest));
   }
 
   /** must be called in synchronized context */
   private String expireEntry() throws IOException, InterruptedException {
     while (header.before == header.after) {
       wait();
     }
+    // should assert that e.referenceCount == 0
```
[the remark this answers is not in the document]
werkt: Too right.  Removing

PR #85, opened 2018-01-20
Switch file/stdout/stderr limits to total result

werkt:
The individual limits per-file could not prevent a large number of small
files from exceeding the reasonable limit of a grpc-handled inbound
message. Prevent a cumulative limit from being exceeded for variable-
sized inline output content. Updated example configs for new limit
field. Improves behavior for #83

### DIFF
--- a/examples/worker.config.example
+++ b/examples/worker.config.example
@@ -22,7 +22,6 @@ stream_stdout: true
 # whether to insert stdout into the CAS, can be:
 #   ALWAYS_INSERT: stdout is always inserted into the CAS
 #   INSERT_ABOVE_LIMIT: stdout is inserted into the CAS when it exceeds the inline limit above
-#   NEVER_INSERT: stdout is never inserted into the CAS
 stdout_cas_policy: ALWAYS_INSERT
 
 # whether the stderr of running processes should be streamed
@@ -31,13 +30,11 @@ stream_stderr: true
 # whether to insert stderr into the CAS, can be:
 #   ALWAYS_INSERT: stderr is always inserted into the CAS
 #   INSERT_ABOVE_LIMIT: stderr is inserted into the CAS when it exceeds the inline limit above
-#   NEVER_INSERT: stderr is never inserted into the CAS
 stderr_cas_policy: ALWAYS_INSERT
 
 # whether to insert output files into the CAS, can be:
 #   ALWAYS_INSERT: output files are always inserted into the CAS
 #   INSERT_ABOVE_LIMIT: output files are inserted into the CAS when it exceeds the inline limit above
-#   NEVER_INSERT: output files are never inserted into the CAS
 file_cas_control: ALWAYS_INSERT
 
 # the worker will take it upon itself to requeue (exceptionally)
--- a/examples/worker.config.example
+++ b/examples/worker.config.example
@@ -11,47 +11,34 @@ root: "/tmp/worker"
 # the local cache location relative to the 'root', or absolute
 cas_cache_directory: "cache"
 
+# total size in bytes of inline content for action results
+# output files, stdout, and stderr content, in that order
+# will be inlined if their cumulative size does not exceed this limit.
+inline_content_limit: 1048567 # 1024 * 1024
+
 # whether the stdout of running processes should be streamed
 stream_stdout: true
 
-# policy control for stdout in results injection and CAS
-stdout_cas_control: {
-  # maximum content size above which inlined output will not be present
-  limit: 16384 # 1024 * 16
-
-  # whether to insert stdout into the CAS, can be:
-  #   ALWAYS_INSERT: stdout is always inserted into the CAS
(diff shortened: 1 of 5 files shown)


src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -76,35 +76,61 @@ public void offer(OperationContext operationContext) {
...
+      CASInsertionPolicy policy = worker.config.getStdoutCasPolicy();
+      boolean withinLimit = inlineContentBytes + stdoutRaw.size() <= worker.config.getInlineContentLimit();
+      if (withinLimit) {
+        inlineContentBytes += stdoutRaw.size();
+      } else {
         resultBuilder.setStdoutRaw(ByteString.EMPTY);
       }
-      if (control.getPolicy() == CASInsertionControl.Policy.ALWAYS_INSERT ||
-          (!withinLimit && control.getPolicy() == CASInsertionControl.Policy.INSERT_ABOVE_LIMIT)) {
+      if (policy.equals(CASInsertionPolicy.ALWAYS_INSERT) ||
```
ola-rozenfeld: Wait, so if we have NEVER_INSERT, and it's above limit, we just pretend there was no stdout at all? Ditto with the output files? That seems wrong. I think this should be an error instead.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -90,21 +90,14 @@ message InstanceConfig {
...
-    INSERT_ABOVE_LIMIT = 2;
-
-    NEVER_INSERT = 3;
-  }
-
-  Policy policy = 2;
+  NEVER_INSERT = 3;
 };
 
 message WorkerConfig {
```
ola-rozenfeld: You don't persist these protos anywhere, right?

werkt: Not in binary form - these are all intended to be loaded like CROSSTOOL

src/main/java/build/buildfarm/worker/ReportResultStage.java
werkt: I'm fine with this being an error, but since NEVER_INSERT has this problem everywhere - it makes unusable configurations out of everything over the inline limit, which can be as low as 0 bytes - I would also be on board with removing NEVER_INSERT entirely as a policy.

Also, glad you pointed this out, because I missed a `contents.add` for `stdoutRaw`.

ola-rozenfeld: Right, removing it entirely sounds good, too!
I missed that too -- I think it would be much less easy to miss if the code snippet was organized a bit differently. It took me a while to process the "contents are already in std*raw field to begin with, so here he just needs to CLEAR that field when the condition does NOT hold, instead of setting it when the condition does hold". It's the reverse of the canonical case, and is a bit hard to read.

How about restructuring it something like:

private int32 inlineOrDigest(
        ByteString content,
        CASInsertionPolicy policy,
        ImmutableList.Builder<ByteString> contents,
        int inlineContentBytes, int inlineContentLimit,
        Function<ByteString,Void> setRaw,
        Function<Digest,Void> setDigest) {
    boolean withinLimit = inlineContentBytes + content.size() <= inlineContentLimit;
    if (withinLimit) {
      setRaw(content);
      inlineContentBytes += content.size();
    }
    if (policy.equals(CASInsertionPolicy.ALWAYS_INSERT) ||
        (!withinLimit && policy.equals(CASInsertionPolicy.INSERT_ABOVE_LIMIT))) {
      contents.add(content);
      setDigest(Digests.computeDigest(content));
    }
    return inlineContentBytes;
}

and call this function for all three cases -- stdout, stderr, and output files?

werkt: I was thinking something similar originally, making this static and adding a test or two for the logic preservation, but didn't want to completely jump that functional hurdle. I'll still need a special case to clear stdout and stderr, but yup, will apply.

werkt:
@ola-rozenfeld friendly ping

examples/worker.config.example
```diff
@@ -11,47 +11,34 @@ root: "/tmp/worker"
...
-  # whether to insert stdout into the CAS, can be:
-  #   ALWAYS_INSERT: stdout is always inserted into the CAS
-  #   INSERT_ABOVE_LIMIT: stdout is inserted into the CAS when it exceeds the inline limit above
-  #   NEVER_INSERT: stdout is never inserted into the CAS
-  policy: ALWAYS_INSERT
-}
+# whether to insert stdout into the CAS, can be:
+#   ALWAYS_INSERT: stdout is always inserted into the CAS
+#   INSERT_ABOVE_LIMIT: stdout is inserted into the CAS when it exceeds the inline limit above
+#   NEVER_INSERT: stdout is never inserted into the CAS
```
ola-rozenfeld: NEVER_INSERT is no more. Ditto below.

PR #87, opened 2018-02-16
cancelOperation instance behavior refinement

werkt:
Cancellation updates should perform the outstanding/requeue/timeout
for all 'done'. Polls for cancelled operations should yield unavailable
responses. Added tests for cancellation behaviors.


src/test/java/build/buildfarm/BuildFarmServerTest.java
```diff
@@ -148,4 +170,130 @@ public void listOperations() {
...
+    operationsStub.cancelOperation(cancelRequest);
+
+    // should now be gone
+    listResponse = operationsStub.listOperations(listRequest);
+
+    assertThat(listResponse.getOperationsList()).isEmpty();
+  }
+
+  @Test
+  public void cancellingExecutingOperationFailsPoll()
```
ola-rozenfeld: nit: it seems you are testing three things for a canceled operation:
- poll should fail
- list should not contain it
- get should contain it with error = CANCELLED
but you do it in two unit tests. Wouldn't just the last one do?

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -225,7 +225,19 @@ public boolean putOperation(Operation operation) {
...
+        requeuer.stop();
+      }
+      // destroy action timed out failure
+      Watchdog operationTimeoutDelay =
+          operationTimeoutDelays.remove(operationName);
+      if (operationTimeoutDelay != null) {
+        operationTimeoutDelay.stop();
+      }
+    } else if (isExecuting(operation)) {
       requeuers.get(operationName).pet();
```
ola-rozenfeld: Lol, this is super cute! I love how you renamed the Watchdog refresh() to pet(), makes so much thematic sense!

src/main/java/build/buildfarm/server/OperationsService.java
```diff
@@ -120,6 +120,7 @@ public void cancelOperation(
     }
 
     instance.cancelOperation(request.getName());
+    responseObserver.onNext(null);
```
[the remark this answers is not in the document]
werkt: I think I was looking for the right behavior here, and didn't know where to look for an example, thanks for the recommendation

src/test/java/build/buildfarm/BuildFarmServerTest.java
werkt: duly nitted - moved the operation state check of a getOperation response into its own test with a precondition validation.

At some point I'm going to want to have a full FSM verification for the operation states for each instance type

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: heh, that's how watchdogs were explained to me ages ago in terms of function, you have to pet them to keep them from biting. I felt it made sense both as a dog person and a user of the java functional with a different method for each 'call' (accept, offer, apply...)

PR #89, opened 2018-02-16
Intercept StatusRuntimeExceptions

werkt:
The TransmitStatusRuntimeExceptionInterceptor can be used to translate
our StatusRuntimeExceptions into onError responses.


werkt:
I'm going to hold this back for the moment - you're right, and I think I might have been getting ahead of myself here - I `throw new StatusRuntimeException` in the shard impl. Since the goal is really to improve the client error context, I'll circle back to this with some guards against the remoteexecution API requests and tests to validate client error information improvements

PR #91, opened 2018-02-17
Remove Encoding with custom latin1 conversion

werkt:
Prefer an InputStreamReader without an intermediate string to hold
configuration data for the server/worker.


werkt:
The presumption here was that the platform native charset Charset.defaultCharset() could be used to interpret the config file that someone writes, and that we didn't need to be in the business of enforcing one charset or another for input. (you work in Big-5, have your config be in Big-5, for whatever reason).

ola-rozenfeld:
Make sense, thanks!

PR #92, opened 2018-02-17
Simplify ByteStringStreamSource, cleanup

werkt:
ByteStringStreamSource is not used in any capacity which requires the
ByteSource interface/impl. Clean up some syntax and indicate thread
interruption on InterruptedException catch during stream operations.

### DIFF
--- a/src/main/java/build/buildfarm/instance/memory/ByteStringStreamSource.java
+++ b/src/main/java/build/buildfarm/instance/memory/ByteStringStreamSource.java
@@ -14,12 +14,11 @@
 
 package build.buildfarm.instance.memory;
 
-import com.google.common.io.ByteSource;
 import com.google.protobuf.ByteString;
 import java.io.InputStream;
 import java.io.OutputStream;
 
-class ByteStringStreamSource extends ByteSource {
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/memory/ByteStringStreamSource.java
```diff
@@ -45,15 +44,15 @@ public void write(byte[] b) {
 
       @Override
```
ola-rozenfeld: now that it no longer extends anything, you need to remove the @Override from everywhere

werkt: and I am unimpressed that that did not warn me! thanks!

werkt: Ah, now I recall, note that this is in the anonymous override of InputStream and OutputStream (@Override does error in the line 75 case below

ola-rozenfeld: Ah, right, sorry!

PR #97, opened 2018-02-17
Switch PipelineStage 'offer' to 'put' for blocking

werkt:
Reflect the patterned blocking queue's 'put' method in PipelineStage's
transfer, indicating and properly behaving with a blocking, rather than
offer with an unchecked response.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -50,7 +50,11 @@ class Executor implements Runnable {
     try {
       command = Command.parseFrom(worker.instance.getBlob(operationContext.action.getCommandDigest()));
     } catch (InvalidProtocolBufferException ex) {
-      owner.error().offer(operationContext);
+      try {
+        owner.error().put(operationContext);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
       owner.release();
       return;
     }
@@ -64,7 +68,11 @@ class Executor implements Runnable {
         .build();
 
     if (!worker.instance.putOperation(operation)) {
-      owner.error().offer(operationContext);
+      try {
+        owner.error().put(operationContext);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      }
       owner.release();
       return;
     }
@@ -98,7 +106,11 @@ class Executor implements Runnable {
           operationContext.metadata.getStderrStreamName());
     } catch (IOException|InterruptedException ex) {
       poller.stop();
-      owner.error().offer(operationContext);
+      try {
+        owner.error().put(operationContext);
+      } catch (InterruptedException intEx) {
+        Thread.currentThread().interrupt();
+      }
       owner.release();
       return;
     }
@@ -112,7 +124,7 @@ class Executor implements Runnable {
                 .setResult(resultBuilder.build())
                 .build()))
             .build();
-        owner.output().offer(new OperationContext(
+        owner.output().put(new OperationContext(
             operation,
             operationContext.execDir,
             operationContext.metadata,
@@ -120,10 +132,18 @@ class Executor implements Runnable {
             operationContext.inputFiles,
             operationContext.inputDirectories));
       } else {
-        owner.error().offer(operationContext);
+        try {
+          owner.error().put(operationContext);
+        } catch (InterruptedException intEx) {
+          Thread.currentThread().interrupt();
+        }
       }
     } catch (InterruptedException ex) {
-      owner.error().offer(operationContext);
+      try {
+        owner.error().put(operationContext);
+      } catch (InterruptedException intEx) {
+        Thread.currentThread().interrupt();
+      }
     }
 
     owner.release();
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/worker/MatchStage.java
```diff
@@ -48,13 +48,18 @@ private boolean fetch(Operation operation) {
...
+      output.put(new OperationContext(
+          operation,
+          execDir,
+          metadata,
+          action,
+          new ArrayList<>(),
+          new ArrayList<>()));
+    } catch (InterruptedException intEx) {
+      Thread.currentThread().interrupt();
+      return false;
```
ola-rozenfeld: But if an interrupt happens here, wouldn't the operation be potentially requeued? https://github.com/bazelbuild/bazel-buildfarm/blob/ca740c29147bfe74cc695730602e2f6607df8fa8/src/main/java/build/buildfarm/instance/stub/StubInstance.java#L339

In general, I'm not sure this interrupt handling preserves the current behavior. Currently, Pipeline.join will catch the exception from any thread, and handle it correctly. But in this change, even though you're calling interrupt(), you're not throwing an exception, and interrupt() on its own will not throw it.
Now, if I understand correctly, the way the code currently works is every single put is follows a successful claim() (except error.put), so it's not supposed to block on put ever anyway. But, just in case the code later changes so that it does -- can you please verify that this will terminate the worker, as it's supposed to?

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -112,18 +124,26 @@ public void run() {
...
             operationContext.inputDirectories));
       } else {
-        owner.error().offer(operationContext);
+        try {
+          owner.error().put(operationContext);
+        } catch (InterruptedException intEx) {
+          Thread.currentThread().interrupt();
+        }
       }
     } catch (InterruptedException ex) {
```
ola-rozenfeld: I think another Thread.currentThread().interrupt(); is missing here. In fact, I think it should replace the owner.error().put(operationContext); If you're getting an interrupt, you don't want to start another potentially long operation as a response to the interrupt, you just want to exit. Ditto with line 107 -- you might want to split them.

src/main/java/build/buildfarm/worker/MatchStage.java
ola-rozenfeld: That sounds good, actually. I also don't have better suggestions.

src/main/java/build/buildfarm/worker/Executor.java
werkt: Agreed. I'm splitting this into a single runInterruptible() that can throw InterruptedException and bail

src/main/java/build/buildfarm/worker/MatchStage.java
```diff
@@ -33,9 +33,11 @@ protected void iterate() throws InterruptedException {
     if (!output.claim()) {
       return;
     }
-    worker.instance.match(worker.config.getPlatform(), worker.config.getRequeueOnFailure(), (operation) -> {
-      return fetch(operation);
-    });
+    worker.instance.match(worker.config.getPlatform(), worker.config.getRequeueOnFailure(), this::fetch);
```
ola-rozenfeld: Oh, cool, I didn't know about this new :: syntax!

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -201,10 +201,8 @@ protected OperationContext tick(OperationContext operationContext) {
 
     try {
       worker.instance.putAllBlobs(contents.build());
-    } catch (IOException ex) {
-    } catch (InterruptedException ex) {
-      poller.stop();
-      return null;
+    } catch (IOException e) {
+      e.printStackTrace();
```
ola-rozenfeld: Doesn't IOException here (in theory) mean that you failed to report results, and the action should be an INTERNAL_ERROR?

src/main/java/build/buildfarm/worker/MatchStage.java
werkt: It was new to me too (I'm really not a Java guy) I had some help from a ex-googler colleague who hopefully will be able to be an objective reviewer here eventually

src/main/java/build/buildfarm/worker/ReportResultStage.java
werkt: Figured you'd catch that, but I wonder how much sense it really makes to report that as an error on the operation (I'm on board with an early exit, though). The IOException is actually a channel for RetryException, which means that we exhausted all possible attempts to communicate with the CAS, which was most likely transiently unavailable, so we should now put the operation into an error state (by again communicating to the same service that we just failed to) rather than letting some other worker that didn't have a beaver gnawing through half of its uplinks try it?

ola-rozenfeld: Right, sorry, I misread the code. Agreed that it makes most sense to let some other worker try this, but this is better achieved by poller.stop; return null; here rather than continuing. Because if these were transient beavers, then putOperation might succeed and the client will get a successful response, causing their build to fail (because Bazel will not retry if it fails to download action results).

werkt: These are low voltage lines, the beavers are much more likely to be nomadic than transient.

(switched to `poller.stop(); return null;`)

PR #99, opened 2018-02-19
Improve error message for invalid hashes.

ttsugriy:
In case client and server are not using the same
digest hash function, server would fail with a message
like:
```
java.lang.NumberFormatException: Invalid hash size
```

PR changes the error message to make it clear which algorithm
was expected on the server side and which hash was checked:
```
java.lang.NumberFormatException: [8029246214105a97ea1eb37977679dab8a238f80d2d4ca4980f142f669b4bbce] is not a valid MD5 hash.
```

This is very useful for people trying to setup build farm
infra for the first time. This is especially useful, since
Bazel has changed the default digest hash from MD5 to SHA256,
but looks like BuildServer is expecting MD5.

### DIFF
--- a/src/test/java/build/buildfarm/common/BUILD
+++ b/src/test/java/build/buildfarm/common/BUILD
@@ -1,0 +1,10 @@
+java_test(
+    name = "DigestUtilTest",
+    srcs = ["DigestUtilTest.java"],
+    deps = [
+        "//src/main/java/build/buildfarm:common",
+        "//src/main/protobuf:build_buildfarm_v1test_buildfarm_java_proto",
+        "//3rdparty/jvm/com/google/truth",
+    ],
+    size = "small",
+)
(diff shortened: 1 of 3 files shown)


src/test/java/build/buildfarm/common/DigestUtilTest.java
```diff
@@ -0,0 +1,49 @@
+// Copyright 2017 The Bazel Authors. All rights reserved.
```
werkt: copyright date

ttsugriy: I must not copy/paste
I must not copy/paste
I must not copy/paste
I must not copy/paste

src/test/java/build/buildfarm/common/BUILD
```diff
@@ -0,0 +1,10 @@
+java_test(
+    name = "DigestUtilTest",
+    srcs = ["DigestUtilTest.java"],
+    deps = [
+        "//src/main/java/build/buildfarm:common",
+        "//src/main/protobuf:build_buildfarm_v1test_buildfarm_java_proto",
+        "//3rdparty/jvm/com/google/truth",
```
werkt: `deps` sorting is alpha (I think) standard from `buildifier`, with 3rdparty preceding src

PR #102, opened 2018-02-19
[Examples] Explicitly configure digest function.

### LINKED ISSUE #101: Update example config to use SHA256 instead of MD5 digest function.
Bazel [has switched to SHA256](https://github.com/bazelbuild/bazel/issues/3193) as a default digest function, which makes build farm incompatible with latest versions of Bazel.

It's possible that some older bazel clients will still use an MD5, which will not be compatible with SHA256, but an explicit configuration option in example config would at least make it very easy to spot and fix the server config.

ttsugriy:
Bazel has switched to SHA256 as a default digest function,
so example configs should be updated accordingly.

Fixes #101

### DIFF
--- a/examples/server.config.example
+++ b/examples/server.config.example
@@ -1,6 +1,8 @@
 # an instance specification
 instances {
   name: "default_memory_instance"
+  # Latest versions of Bazel are using SHA256 as a default digest function
+  hash_function: SHA256
 
   # the implicit type specifier for this instance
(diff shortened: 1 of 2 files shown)


werkt:
This is a duplicate of #90, it will address #101

ttsugriy:
@werkt, you should probably also update worker config in #90

werkt:
Done, thanks for the advice!

PR #109, opened 2018-02-20
Cleaning up unused code in worker package

werkt:


werkt:
Clearly because we're too cool for linting! Turning lints on locally now, will probably end up with a few more commits from that. (wow, protobuf and guava don't like to guard against -Xlint:all)

werkt:
Quick followup on this. The default javac linter would not catch this. If you have any recommendations on how to incorporate that level of warning (with promotion to error with -Werror) through bazel builds with CI, as it appears to only be available with IDE linters (like eclipse).

ola-rozenfeld:
Yep, I checked now, turns out this is part of our internal Java analysis
suite, which is not open sourced. Sorry!

PR #115, opened 2018-02-23
M Default max action timeout

werkt:
Add instance and worker specifications for action default and maximum
timeouts. These timeouts will not affect the action definition - results
will be stored with the requested action message digest key.
Supplement example configs with action timeouts

### DIFF
--- a/examples/server.config.example
+++ b/examples/server.config.example
@@ -39,6 +39,19 @@ instances {
 
     # limit for CAS total content size in bytes
     cas_max_size_bytes: 1073741824 # 1024 * 1024 * 1024
+
+    # an imposed action-key-invariant timeout used in the unspecified timeout case
+    default_action_timeout: {
+      seconds: 600
+      nanos: 0
+    }
+
+    # a limit on the action timeout specified in the action, above which
+    # the operation will report a failed result immediately
+    maximum_action_timeout: {
+      seconds: 3600
+      nanos: 0
+    }
   }
 }
 
@@ -67,6 +80,16 @@ instances {
     }
 
     cas_max_size_bytes: 655360 # 640k
+
+    default_action_timeout: {
+      seconds: 600
+      nanos: 0
+    }
+
+    maximum_action_timeout: {
+      seconds: 3600
+      nanos: 0
+    }
   }
 }
 
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -572,6 +572,9 @@ public abstract class AbstractServerInstance implements Instance {
             .build()))
         .setResponse(Any.pack(ExecuteResponse.newBuilder()
             .setResult(actionResult)
+            .setStatus(com.google.rpc.Status.newBuilder()
+                .setCode(com.google.rpc.Code.DEADLINE_EXCEEDED.getNumber())
+                .build())
             .setCachedResult(cachedResult)
             .build()))
         .build());
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -35,6 +35,7 @@ import com.google.longrunning.Operation;
 import com.google.protobuf.Any;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.rpc.Code;
 import io.grpc.Status;
 import java.util.HashSet;
 import java.util.Map;
@@ -372,6 +373,9 @@ public abstract class AbstractServerInstance implements Instance {
           .setDone(true)
           .setResponse(Any.pack(ExecuteResponse.newBuilder()
               .setResult(actionResult)
+              .setStatus(com.google.rpc.Status.newBuilder()
+                  .setCode(Code.OK.getNumber())
+                  .build())
               .setCachedResult(actionResult != null)
               .build()));
     } else {
@@ -545,7 +549,7 @@ public abstract class AbstractServerInstance implements Instance {
     putOperation(operation.toBuilder()
         .setDone(true)
         .setError(com.google.rpc.Status.newBuilder()
-            .setCode(com.google.rpc.Code.CANCELLED.getNumber())
+            .setCode(Code.CANCELLED.getNumber())
             .build())
         .build());
   }
@@ -558,9 +562,8 @@ public abstract class AbstractServerInstance implements Instance {
         .build();
     ExecuteResponse executeResponse = ExecuteResponse.newBuilder()
         .setResult(actionResult)
-        .setCachedResult(false)
         .setStatus(com.google.rpc.Status.newBuilder()
-            .setCode(com.google.rpc.Code.DEADLINE_EXCEEDED.getNumber())
+            .setCode(Code.DEADLINE_EXCEEDED.getNumber())
             .build())
         .build();
     ExecuteOperationMetadata metadata = expectExecuteOperationMetadata(operation);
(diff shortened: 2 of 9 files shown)


examples/server.config.example
```diff
@@ -39,6 +39,19 @@ instances {
...
+    # an imposed action-key-invariant timeout used in the unspecified timeout case
+    default_action_timeout: {
+      seconds: 600
+      nanos: 0
+    }
+
+    # a limit on the action timeout specified in the action, above which
+    # the operation will report a failed result immediately
+    maximum_action_timeout: {
+      seconds: 3600
```
ola-rozenfeld: Interesting! We had the opposite problem -- for a while, our system had a minimal supported timeout. Also, note 1h is quite restrictive, integration tests often run longer.

Also, not for this patch, but it seems that our systems are interpreting the timeout field differently. We have not meant for the timeout to  be an absolute expiry deadline -- rather, the clock should only start ticking when actual execution begins: https://github.com/googleapis/googleapis/blob/master/google/devtools/remoteexecution/v1test/remote_execution.proto#L343 It should not count scheduling time, bot creation/setup time, nor any other time required by the service.
If we want to add a limit on this "exec overhead time" (we have considered it), we will put it in a different field (probably in ExecuteRequest), and return RESOURSE_EXHAUSTED if it runs over, so that Bazel can retry.

examples/server.config.example
```diff
@@ -39,6 +39,19 @@ instances {
...
     cas_max_size_bytes: 1073741824 # 1024 * 1024 * 1024
+
+    # an imposed action-key-invariant timeout used in the unspecified timeout case
+    default_action_timeout: {
+      seconds: 600
+      nanos: 0
+    }
+
+    # a limit on the action timeout specified in the action, above which
+    # the operation will report a failed result immediately
```
ola-rozenfeld: nit: will report a timeout.

Timeout is a special kind of failure which is propagated differently throughout the API.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -207,6 +208,19 @@ protected void onQueue(Operation operation, Action action) {
     putBlob(action.toByteString());
   }
 
+  @Override
+  protected void validateAction(Action action) {
+    if (action.hasTimeout() && config.hasMaximumActionTimeout()) {
+      Duration timeout = action.getTimeout();
+      Duration maximum = config.getMaximumActionTimeout();
+      Preconditions.checkState(
```
ola-rozenfeld: Maybe not in this patch, but these precondition violations are of the kind that you want to propagate back to the user as a FAILED_PRECONDITION status, with proper details, so that the user sees the error in Bazel and knows what do they need to fix. One way to do it would be to throw a proper StatusRuntimeException from here, and let your new interceptor handle it. I do it almost right in https://github.com/bazelbuild/bazel/blob/master/src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/StatusUtils.java (some of my status codes are wrong, for example it should be FAILED_PRECONDITION instead of NOT_FOUND for missing inputs/command https://github.com/googleapis/googleapis/blob/master/google/devtools/remoteexecution/v1test/remote_execution.proto#L87), but at least it allows to distinguish between user-actionable errors and INTERNAL_ERROR, retriable vs non-retriable, etc.

werkt: Agreed, I have another diff in the pipeline that will catch IllegalStateException thrown by Preconditions in validateAction (which will declare the throw), and translate in either Instance::execute() or ExecutionService::execute() into StatusRuntimeException (depending on whether I care about introducing grpc deps at the AbstractInstanceServer level). Might just end up translating to differentiate IllegalState in validation from IllegalState during subsequent internal execute operations

examples/server.config.example
werkt: Whelp, that's why it's a config, and just an example - figured it was as good a marker as any, and that if you ran into it, you would be able to figure out what that limit was easily.

I don't think we are interpreting it differently - the fuse is lit when the worker transitions the operation to the Executing state, and when the timeout + config.operation_completed_delay has elapsed, the operationTimeoutDelay watchdog fires its runnable, the operation is forcibly completed with a -1 exit status and a stderr message indicating the failure due to timeout.
Since my pipeline does not change the status to EXECUTING until we have begun the Command, we should have the same measurements, except perhaps for the time spent in ReportResultsStage.

If you do set up a timeout like that, let me know.  Right now there is no limit on a properly functioning worker's time to setup for execution, as long as they are polling regularly enough. The only reaction to failure to poll is requeuing, not operation failure.

examples/server.config.example
werkt: I'm actually set up right now in `MemoryInstance#validateAction` to fail the execute with a Precondition violation if the requested timeout is above the `maximum_action_timeout`, so the execute will fail with FAILED_PRECONDITION (with updates to #73)

I suppose I would want to see something indicating that this should not be the reaction to such an improperly defined action - it does fly somewhat in the face of the specification comments indicating that 'if you fix the precondition state, the action can be retried', since all other instances of that are data consistency issues (missing blobs, directories, etc)

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: Updated #46 with the change to respond with FAILED_PRECONDITION on Precondition violations.

examples/server.config.example
ola-rozenfeld: Oh, sorry, you're right, I didn't get when the fuse was lit. There's one thing that bugs me about expireOperation function -- I don't see you doing executeResponse.setStatus(Status.newBuilder().setCode(DEADLINE_EXCEEDED)). That's the way Bazel will know about the timeout, and display the test status as TIMEOUT rather than FAILED. You could also use the operation.error field, but that is deprecated, so you shouldn't (because it's a oneof, so if you use error, you cannot return the result).

werkt: Gotcha. I threw that in there, but I assume there are numerous other places where I'm missing this style of response - I had filed #73 to deal with this independently

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -551,32 +551,26 @@ public void cancelOperation(String name) {
...
-        putActionResult(actionKey, actionResult);
-      }
+    ActionResult actionResult = ActionResult.newBuilder()
+        .setExitCode(-1)
+        .setStderrRaw(ByteString.copyFromUtf8(
+            "[BUILDFARM]: Action timed out with no response from worker"))
+        .build();
+    ExecuteResponse executeResponse = ExecuteResponse.newBuilder()
+        .setResult(actionResult)
+        .setCachedResult(false)
```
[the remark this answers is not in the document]
ola-rozenfeld: Also, please do the same thing when the worker sees the timeout, not just the watchdog.

werkt:
Dealt with it here, feel free to review last change there

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -190,6 +200,7 @@ public void run() {
     stderrReaderThread.start();
 
     boolean doneWaiting = false;
+    Code statusCode = Code.UNKNOWN;
     if (timeout == null) {
       exitValue = process.waitFor();
```
ola-rozenfeld: in this case it should be Code.OK, no?

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -199,11 +210,13 @@ public void run() {
         if (remainingNanoTime > 0) {
           if (process.waitFor(remainingNanoTime, TimeUnit.NANOSECONDS)) {
```
ola-rozenfeld: I don't really understand this line, btw -- if process.waitFor returns false, doesn't that mean the time has elapsed fully and we should be DEADLINE_EXCEEDED already? And then why do we need a loop?

src/main/java/build/buildfarm/worker/Executor.java
werkt: yup, fixing

src/main/java/build/buildfarm/worker/Executor.java
werkt: My C is showing (expecting EINTR here as a possible response to `wait(2)`). You're right, documentation indicates that if we don't get the waitFor true return, we will get false or InterruptedException will be thrown.

PR #116, opened 2018-02-23
S Cleanup of digestUtil, remove unused methods

werkt:
Several methods are not in use from digestUtil, and the use of
ByteSource is unnecessary and can be substituted for a simpler hashFn
interaction.

### DIFF
--- a/src/main/java/build/buildfarm/common/DigestUtil.java
+++ b/src/main/java/build/buildfarm/common/DigestUtil.java
@@ -15,9 +15,10 @@ package build.buildfarm.common;
 
 import build.buildfarm.v1test.InstanceConfig;
 import com.google.common.base.Preconditions;
+import com.google.common.hash.Funnels;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -96,34 +97,13 @@ public DigestUtil(HashFunction hashFn) {
   }
 
   public Digest compute(ByteString blob) {
+    Hasher hasher = hashFn.getHash().newHasher();
```
ola-rozenfeld: Nice, I need to do the same in Bazel code!

werkt: I really wanted to get to being able to treat the sink as a legit sink without the possibility of IOException due to the InputStream round trip, but my eyes glassed over as I pulled apart hasher code. lmk if you come up with something better for bazel!

PR #118, opened 2018-02-28
L Adopt testing suite from bazel

werkt:
In the interest of modularizing tests, I've adapted bazel's test suite
mechanism to our testing, and moved our previously available tests into
this model.


src/test/java/build/buildfarm/instance/stub/ByteStringIteratorInputStreamTest.java
```diff
@@ -116,4 +118,51 @@ public void readWithOffsetAndLengthSpanningChunks() throws IOException {
...
+  @Rule
+  public final ExpectedException exception = ExpectedException.none();
+
+  @Test
+  public void closedStreamAvailableThrowsIOException() throws IOException {
+    InputStream in = new ByteStringIteratorInputStream(
+        ImmutableList.<ByteString>of(
+            ByteString.copyFromUtf8("Hello, World")).iterator());
+    in.close();
+    exception.expect(IOException.class);
```
ola-rozenfeld: nit: For your use-cases here, the @Test(expected = IOException.class) idiom would be simpler, I think.
Also, you can use fail("excepted exception") instead of assert true = false

src/test/java/build/buildfarm/server/BuildFarmServerTest.java
```diff
@@ -332,4 +333,25 @@ private Operation executeAction(Action action) {
...
+        .setInstanceName("memory")
+        .setActionDigest(digestUtil.empty())
+        .build();
+
+    ActionCacheGrpc.ActionCacheBlockingStub actionCacheStub =
+        ActionCacheGrpc.newBlockingStub(inProcessChannel);
+
+    StatusRuntimeException notFound = null;
+    try {
+      actionCacheStub.getActionResult(request);
```
ola-rozenfeld: nit: I think this is a bit more readable:
try {
  something bad
  fail("expected exception")
} catch (MyException e) {
  assertThat(e).isWhatIThought()
}

PR #120, opened 2018-02-28
M delinting code under src

werkt:
static method calling on instances removed
Removed explicit close for try-with-resources on autoclose
  Triggered cleanup for CASFileCache exception handling
Removed Throwables.propagate usage for deprecated (ignored for md5 and sha1)
Specified Builder generic modifiers for rawtypes

### DIFF
--- a/src/main/java/build/buildfarm/server/InstanceNotFoundException.java
+++ b/src/main/java/build/buildfarm/server/InstanceNotFoundException.java
@@ -1,6 +1,8 @@
 package build.buildfarm.server;
 
 class InstanceNotFoundException extends Exception {
+  private static final long serialVersionUID = 1;
+
   public final String instanceName;
 
   InstanceNotFoundException( String instanceName ) {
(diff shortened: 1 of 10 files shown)


buchgr:
can you please push again - CI should then be triggered again.

src/main/java/build/buildfarm/server/InstanceNotFoundException.java
```diff
@@ -1,6 +1,8 @@
 package build.buildfarm.server;
 
 class InstanceNotFoundException extends Exception {
+  private static final long serialVersionUID = 1;
```
ola-rozenfeld: What are these fields for?

werkt:
```
src/main/java/build/buildfarm/instance/stub/RetryException.java:23: warning: [serial] serializable class RetryException has no definition of serialVersionUID
public final class RetryException extends IOException {
             ^
```

PR #121, opened 2018-03-05
L Namespace worker impl under 'operationqueue'

werkt:
Move the operationqueue worker implementation to its own namespace under
the worker package, and abstract away interactions with the
instance/queue to go through a WorkerContext. This change defines a
BlobPathFactory interface which can generate Paths for files or
directories, with the CASFileCache implementing this interface.


src/main/java/build/buildfarm/worker/InputFetchStage.java
```diff
@@ -57,14 +57,11 @@ public void put(OperationContext operationContext) throws InterruptedException {
...
-              operationName,
-              ExecuteOperationMetadata.Stage.QUEUED);
-          return success;
-        });
-    new Thread(poller).start();
+    Poller poller = workerContext.createPoller(
+        "InputFetchStage",
+        operationContext.operation.getName(),
+        ExecuteOperationMetadata.Stage.QUEUED,
+        () -> {});
```
ola-rozenfeld: nit: () -> {} can be the default argument

src/main/java/build/buildfarm/worker/InputFetchStage.java
```diff
@@ -75,7 +72,7 @@ protected OperationContext tick(OperationContext operationContext) throws Interr
     boolean success = true;
     try {
       if (Files.exists(execDir)) {
-        Worker.removeDirectory(execDir);
+        CASFileCache.removeDirectory(execDir);
```
ola-rozenfeld: I think I understand why you put the method in CASFileCache, but it is quite confusing to see CASFileCache behind an interface in most places, but then referenced directly here. Can you maybe expose it in WorkerContext instead?

src/main/java/build/buildfarm/worker/InputFetchStage.java
ola-rozenfeld: I meant an overload that only accepts three parameters.  https://stackoverflow.com/questions/997482/does-java-support-default-parameter-values

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -214,7 +234,7 @@ private void expireDirectory(Digest digest) throws IOException {
         fileEntry.containingDirectories.remove(digest);
       }
     }
-    Worker.removeDirectory(path);
+    CASFileCache.removeDirectory(path);
```
ola-rozenfeld: this is the same file, just do removeDirectory(path).

werkt: I could have sworn I saw a linting warning here, but maybe that's only on non-self-class invocations.

PR #122, opened 2018-03-05
Adding digestUtil tests for coverage

werkt:


src/test/java/build/buildfarm/common/DigestUtilTest.java
```diff
@@ -16,58 +16,89 @@
...
   }
 
   @Test
-  public void computesSha1Hash() {
-    ByteString content = ByteString.copyFromUtf8("bazel");
-    DigestUtil util = new DigestUtil(DigestUtil.HashFunction.SHA1);
-    Digest digest = util.compute(content);
-    assertThat(digest.getHash()).isEqualTo("287d5d65c10a8609e9c504c81f650b0e1669a824");
+  public void unrecognizedHashFunctionThrows() {
+    exception.expect(IllegalArgumentException.class);
```
ola-rozenfeld: move to @test to save a line?

src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -91,6 +91,13 @@ private ActionKey(Digest digest) {
   private final HashFunction hashFn;
   private final Digest empty;
 
+  public static DigestUtil forHash(String hashName) {
```
ola-rozenfeld: Just curious, where do you anticipate needing this function?

werkt: `bf-cat` argument interpretation

werkt: Also the other `bf-*` utils that communicate with instances.

PR #123, opened 2018-03-05
M Adding memory-instance-tests

werkt:


src/test/java/build/buildfarm/instance/memory/ByteStringStreamSourceTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Override
+    public void run() {
+      runCount++;
+    }
+  };
+
+  @Test
+  public void closeCallsOnClose() throws IOException {
+    CountingRunnable counter = new CountingRunnable();
+    ByteStringStreamSource source = new ByteStringStreamSource(counter);
```
ola-rozenfeld: Note you could also do it with mocks -- verifying how many times was something called is a very typical mock use-case. But it's simple enough either way, so up to you.

src/test/java/build/buildfarm/instance/memory/ByteStringStreamSourceTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Test
+  public void skipBlocksForInput() throws IOException, InterruptedException {
+    ByteStringStreamSource source = new ByteStringStreamSource(() -> {});
+    InputStream inputStream = source.openStream();
+    Thread thread = new Thread(() -> {
+      try {
+        assertThat(inputStream.skip(1)).isEqualTo(1);
+      } catch (IOException e) {
+        assertThat(false).isTrue();
```
ola-rozenfeld: Why do you need this catch at all? If IOException is thrown here, wouldn't you get a meaningful test failure? I guess if it doesn't compile otherwise, putting a fail("Unexpected IOException: " + e) will be better.

src/test/java/build/buildfarm/instance/memory/ByteStringStreamSourceTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Test
+  public void skipNegativePreventsSkip() throws IOException {
+    ByteStringStreamSource source = new ByteStringStreamSource(() -> {});
+    InputStream inputStream = source.openStream();
+    assertThat(inputStream.skip(-1)).isEqualTo(0);
+  }
+
+  @Test
+  public void skipBlocksForInput() throws IOException, InterruptedException {
```
ola-rozenfeld: Maybe add another test that read behaves the same way?

src/test/java/build/buildfarm/instance/memory/MemoryLRUContentAddressableStorageTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Test
+  public void expireShouldCallOnExpiration() {
+    ContentAddressableStorage storage = new MemoryLRUContentAddressableStorage(10);
+
+    DigestUtil sha1DigestUtil = new DigestUtil(DigestUtil.HashFunction.SHA256);
+    CountingRunnable onExpiration = new CountingRunnable();
+    storage.put(new Blob(ByteString.copyFromUtf8("stdout"), sha1DigestUtil), onExpiration);
+    assertThat(onExpiration.runCount).isEqualTo(0);
+    storage.put(new Blob(ByteString.copyFromUtf8("stderr"), sha1DigestUtil));
```
ola-rozenfeld: Maybe add another small test that the total size works as expected, expiring entries exactly after the limit is reached? I understand there's quite a few tests that could be added here, but that one should be copy/paste of this test case, more or less.

src/test/java/build/buildfarm/instance/memory/ByteStringStreamSourceTest.java
werkt: Added!

src/test/java/build/buildfarm/instance/memory/ByteStringStreamSourceTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+@RunWith(JUnit4.class)
+public class ByteStringStreamSourceTest {
+  @Test
+  public void closeCallsOnClose() throws IOException {
+    Runnable mockOnClose = mock(Runnable.class);
+    ByteStringStreamSource source = new ByteStringStreamSource(mockOnClose);
+    verify(mockOnClose, never()).run();
+    source.getOutputStream().close();
+    verify(mockOnClose).run();
```
ola-rozenfeld: nit: verify(mockOnClose, times(1)).run();
And ditto in all other cases. I think the default verifies at least once, not exactly once.

PR #128, opened 2018-03-16
L OutputDirectories implementation

werkt:


ola-rozenfeld:
Will definitely review before the end of the week, sorry!

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -186,19 +185,112 @@ protected OperationContext tick(OperationContext operationContext) throws Interr
...
-          policy,
+          workerContext.getFileCasPolicy(),
           contents,
           inlineContentBytes,
           workerContext.getInlineContentLimit(),
           () -> outputFileBuilder.setContent(content),
           (fileContent) -> outputFileBuilder.setDigest(getDigestUtil().compute(fileContent)));
     }
 
+    for (String outputDir : outputDirs) {
```
ola-rozenfeld: I would highly recommend to use an implementation similar to https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/remote/AbstractRemoteActionCache.java#L289
It is much easier to understand (to me, at least), and also more efficient in not computing the digests to upload twice (I guess you can also fix the inefficiency here by maintaining a similar map of digest to path or byte[]).

This will also allow you to optimize later by not loading all the contents to upload into memory at once (I suspect that you will have to do this soon).

Perhaps more importantly, the dirents of each directory need to be sorted per https://github.com/googleapis/googleapis/blob/master/google/devtools/remoteexecution/v1test/remote_execution.proto#L457. IIUC, walkFileTree gives no order guarantees.

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -186,19 +185,112 @@ protected OperationContext tick(OperationContext operationContext) throws Interr
...
+          FileNode.Builder fileNodeBuilder = currentDirectory.addFilesBuilder()
+              .setName(file.getFileName().toString())
+              .setDigest(getDigestUtil().compute(content))
+              .setIsExecutable(Files.isExecutable(file));
+          return FileVisitResult.CONTINUE;
+        }
+
+        @Override
+        public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
+          path.push(currentDirectory);
```
ola-rozenfeld: Are you sure you need to push current directory when it is equal to the root? It seems to break the invariant that currentDirectory's parent is equal to stack.top

src/test/java/build/buildfarm/worker/OutputDirectoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+    assertThat(outputDirectory).doesNotContainKey("foo");
+    assertThat(outputDirectory).containsKey("bar");
+    assertThat(outputDirectory.get("bar")).isEmpty();
+  }
+
+  @Test
+  public void outputDirectoryParsesOutputDirs() {
+    OutputDirectory outputDirectory = OutputDirectory.parse(
+        ImmutableList.<String>of(),
+        ImmutableList.<String>of("bar/baz", "foo"));
```
ola-rozenfeld: As noted, corner cases include directories with + in the name (or anything that sorts below /), like [a,a+b,a/b,aa/], etc.

src/main/java/build/buildfarm/worker/OutputDirectory.java
```diff
@@ -0,0 +1,82 @@
...
+    return parseDirectories(Iterables.mergeSorted(
+        ImmutableList.<Iterable<String>>of(
+            Iterables.transform(
+                outputFiles,
+                (file) -> file.contains("/") ? file.substring(0, file.lastIndexOf('/') + 1) : ""),
+            outputDirs),
+        (a, b) -> a.compareTo(b)));
+  }
+
+  private static OutputDirectory parseDirectories(Iterable<String> outputDirs) {
```
ola-rozenfeld: Do I understand correctly that the only usage of this class is so you could create all the directories involved (stamp)?
If that is the case, why not use mkdirs every time you need to create an output, like https://github.com/bazelbuild/bazel/blob/master/src/tools/remote/src/main/java/com/google/devtools/build/remote/worker/ExecutionServer.java#L201 ?

werkt: It is actually used as an efficient predicate for the symlinked directory optimization - a `linkInputs` subdirectory cannot be symlinked if it leads to an output path. Output files must have real and operation independent directories all the way up to the `execRoot`.

ola-rozenfeld: Ah, thank you. In that case, a nit: I think composition is better than inheritance. You want expose the users of this class to a very narrow subset of the HashMap interface (no modification, iteration not necessary either, just is-in check), and in fact it's not even a subset, the fact that you use a hashmap and not something else is an implementation detail. For me, a reason to inherit from X is only if I need to use instances of MyClass outside of MyClass in contexts where an X is expected.

src/main/java/build/buildfarm/worker/OutputDirectory.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+import java.util.Stack;
+
+// would be great to be able to inherit from ImmutableMap...
+public class OutputDirectory extends HashMap<String, OutputDirectory> {
+  private OutputDirectory() {
+  }
+
+  public static OutputDirectory parse(Iterable<String> outputFiles, Iterable<String> outputDirs) {
+    return parseDirectories(Iterables.mergeSorted(
```
ola-rozenfeld: Wait, now you no longer need to sort them here, right?
In fact, I think if you add a:
Iterables.transform(outputDirs, (d) -> d.charAt(d.length() - 1) == '/') ? d : d + "/")
here and leave the mergeSorted, then you don't have to do anything extra in the next function, and the code will be more obviously correct (everything has a / or is an empty string, no assumptions on input).

src/main/java/build/buildfarm/worker/ReportResultStage.java
werkt: Having looked at the bazel implementation, I would say it falls short of my ideal. It should, for each path to upload: start a digesting chunker (that computes a digest in a single pass of the file) and on completion decrements a latch count, which releases the action result `build()` and post. I will try to cruft that up.

src/main/java/build/buildfarm/worker/OutputDirectory.java
werkt: Completely agree with that, I'll cut it down.

src/main/java/build/buildfarm/worker/OutputDirectory.java
werkt: I will still need to sort the transformed Iterable - `mergeSorted` is predicated on non-descending order for the set of Iterables

werkt: Yeah, the `mergeSorted` is no longer applicable with your correct observation that the initial sorting of output files/dirs is not sufficient for sorting 'parent directory comes first'. This needs to become a `concat`, and everything has a / is my invariant (no empty strings)

werkt:
I've update this, but it now has a dependency on a separate change I made for Chunker during my endpoint factorization. I have not implemented the policy issues we've discussed (output files in output dirs, nested output dirs, root level output dir) yet.

ola-rozenfeld:
Also, please check out Jakob's change
https://github.com/bazelbuild/bazel/commit/ff008f445905bf6f4601a368782b620f7899d322
You might want to include some of these in your implementation.

src/main/java/build/buildfarm/worker/OutputDirectory.java
```diff
@@ -0,0 +1,122 @@
...
+
+    Builder currentBuilder = builder;
+    String prefix = "/";
+    for (String outputDir : sortedOutputDirs) {
+      while (!outputDir.startsWith(prefix)) {
+        currentBuilder = stack.pop();
+        int upPathSeparatorIndex = prefix.lastIndexOf('/', prefix.length() - 2);
+        prefix = prefix.substring(0, upPathSeparatorIndex + 1);
+      }
+      String prefixedFile = outputDir.substring(prefix.length(), outputDir.length() - 1);
```
pcj: Code fails at line 87 here with a string out of bounds exception when the outputDir is identical to the prefix.  I worked around it with a `if (outputDir.length() == prefix.length()) { continue; }`.  WIth that additional change this PR appears to work for me.

pcj:
Friendly ping here.  I'd like to use some of the newer commits on master but buildfarm is unusable for me without this PR, so unless there is an architectural issue here, I'd vote to see it in.  Until then I'm maintaining a separate branch off this one with a few cherry-picks.

ola-rozenfeld:
Will review today, sorry!

src/main/java/build/buildfarm/worker/UploadManifest.java
```diff
@@ -0,0 +1,338 @@
...
+   * directory, including the descendants, can be reconstructed and 2) uploading all the
+   * non-directory descendant files.
+   *
+   * <p>Attempting to a upload symlink results in a {@link
+   * IllegalStateException}, since cachable actions shouldn't emit symlinks.
+   */
+  public void addFiles(Iterable<Path> files, CASInsertionPolicy policy)
+      throws IllegalStateException, IOException, InterruptedException {
+    for (Path file : files) {
+      // TODO(ulfjack): Maybe pass in a SpawnResult here, add a list of output files to that, and
```
ola-rozenfeld: Remove comment

src/main/java/build/buildfarm/worker/UploadManifest.java
```diff
@@ -0,0 +1,338 @@
...
+      // rely on the local spawn runner to stat the files, instead of statting here.
+      FileStatus stat = statIfFound(file, /* followSymlinks= */ false);
```
ola-rozenfeld: Instead you could split the addFile and addDirectory methods which would allow you to make the expected output type check that we discussed.

src/main/java/build/buildfarm/worker/UploadManifest.java
```diff
@@ -0,0 +1,338 @@
...
+      // If this codepath is ever hit, then this method should be rewritten to properly distinguish
+      // between not-found exceptions and others.
+      throw new IllegalStateException(e);
+    }
+  }
+
+  private static Dirent.Type direntTypeFromStat(FileStatus stat) {
+    if (stat == null) {
+      return Dirent.Type.UNKNOWN;
+    } else if (stat.isSpecialFile()) {
```
ola-rozenfeld: nit: remove all the else's:

     if (<case1>) {
       return <v1>
     }
     if (<case2>) {
       return <v2>
     }

etc

src/test/java/build/buildfarm/worker/OutputDirectoryTest.java
ola-rozenfeld: Your current code should work well with these, but would still be good to have one test with these values, to prevent future regressions.

src/main/java/build/buildfarm/worker/OutputDirectory.java
```diff
@@ -0,0 +1,125 @@
...
+        continue;
+      }
+      String prefixedFile = outputDir.substring(prefix.length(), outputDir.length() - 1);
+      while (prefixedFile.length() > 0) {
+        int separatorIndex = prefixedFile.indexOf('/');
+        if (separatorIndex == 0) {
+          throw new IllegalArgumentException("double separator in output directory");
+        }
+
+        String directoryName = separatorIndex == -1 ? prefixedFile : prefixedFile.substring(0, separatorIndex);
```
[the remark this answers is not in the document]
werkt: too right, this actually highlighted a problem in my iteration - I was clipping the trailing slash off of the initial prefixedFile. Fix forthcoming.

src/main/java/build/buildfarm/worker/ReportResultStage.java
```diff
@@ -81,60 +96,69 @@ private DigestUtil getDigestUtil() {
...
+
+    for (Digest digest : digests) {
+      Chunker chunker;
+      Path file = digestToFile.get(digest);
+      if (file != null) {
+        chunker = new Chunker(file, digest);
+      } else {
+        chunker = digestToChunkers.get(digest);
+        if (chunker == null) {
+          String message = "FindMissingBlobs call returned an unknown digest: " + digest;
```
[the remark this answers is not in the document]
werkt: agreed, a null chunker will just be omitted in a fix.

src/main/java/build/buildfarm/worker/OutputDirectory.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private static OutputDirectory parseDirectories(Iterable<String> outputDirs) {
+    OutputDirectory outputDirectory = new OutputDirectory();
+    Stack<OutputDirectory> stack = new Stack<>();
+
+    OutputDirectory currentOutputDirectory = outputDirectory;
+    String prefix = "";
+    for (String outputDir : outputDirs) {
+      while (!outputDir.startsWith(prefix)) {
```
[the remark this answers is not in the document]
werkt: I believe you were referring to these below in your comments about wanting to have these values in the test: I'll have it generate this list with a file 'a/file', in addition to the other two directories, but since 'a' is a parent of 'a/b/c', that violates our preconditions and the action would be rejected prior to this

PR #134, opened 2018-03-23
Correct setting of file permissions on Windows.

excitoon:
Unfortunately one can't set file permissions with `nio` on Windows.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -25,6 +25,7 @@ import com.google.devtools.remoteexecution.v1test.Directory;
 import com.google.devtools.remoteexecution.v1test.DirectoryNode;
 import com.google.devtools.remoteexecution.v1test.FileNode;
 import io.grpc.StatusRuntimeException;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.FileVisitResult;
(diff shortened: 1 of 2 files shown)


excitoon:
Rebased and fixed.

src/test/java/build/buildfarm/worker/CASFileCacheTest.java
```diff
@@ -49,8 +49,7 @@
   private Map<Digest, ByteString> blobs;
 
   protected CASFileCacheTest(Configuration config) {
-    this.config = config.toBuilder()
-        .build();
+    this.config = config;
```
excitoon: Well, actually, I thought it may be useful for future improvements of the test.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -347,12 +346,7 @@ public Path put(Digest digest, boolean isExecutable, Digest containingDirectory)
   }
 
   private static void setPermissions(Path path, boolean isExecutable) throws IOException {
-    ImmutableSet.Builder<PosixFilePermission> perms = new ImmutableSet.Builder<PosixFilePermission>()
-      .add(PosixFilePermission.OWNER_READ);
-    if (isExecutable) {
-      perms.add(PosixFilePermission.OWNER_EXECUTE);
-    }
-    Files.setPosixFilePermissions(path, perms.build());
+    new File(path.toString()).setExecutable(false, isExecutable);
```
ianoc: @excitoon this looks like this breaks linux? this change doesn't seem to set executable permissions on files extracted from the cache at all. I think this should look like `.setExecutable(isExecutable, true)`

excitoon: Yeah, you're right. Somehow I thought first argument is for other users and second is for owner.

PR #136, opened 2018-03-26
M Fix operation paging, reinforce with tests

werkt:
The memory instance would not progress beyond the initial page due to an
incorrect test against the token itself rather than the deserialized
OperationIteratorToken's name, and the paged test was initialized to
true, making the paging loop break on the first entry anyway.

Added tests to guard operation list retrieval.

### DIFF
--- a/src/test/java/build/buildfarm/instance/memory/MemoryInstanceTest.java
+++ b/src/test/java/build/buildfarm/instance/memory/MemoryInstanceTest.java
@@ -22,6 +22,7 @@ import build.buildfarm.common.DigestUtil;
 import com.google.common.collect.ImmutableList;
 import com.google.longrunning.Operation;
 import com.google.protobuf.Duration;
+import com.google.protobuf.util.Durations;
 import java.util.HashMap;
 import java.util.Map;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


src/test/java/build/buildfarm/instance/memory/MemoryInstanceTest.java
```diff
@@ -0,0 +1,161 @@
...
+    MockitoAnnotations.initMocks(this);
+    outstandingOperations = new HashMap<>();
+    MemoryInstanceConfig memoryInstanceConfig = MemoryInstanceConfig.newBuilder()
+        .setListOperationsDefaultPageSize(1024)
+        .setListOperationsMaxPageSize(16384)
+        .setTreeDefaultPageSize(1024)
+        .setTreeMaxPageSize(16384)
+        .setOperationPollTimeout(Duration.newBuilder()
+            .setSeconds(10)
+            .setNanos(0))
```
ola-rozenfeld: nit: you never need to set default values explicitly.

src/test/java/build/buildfarm/instance/memory/MemoryInstanceTest.java
```diff
@@ -0,0 +1,161 @@
...
+    ImmutableList.Builder<Operation> operations = new ImmutableList.Builder<>();
+
+    String nextToken = instance.listOperations(
+        /* pageSize=*/ 1,
+        /* pageToken=*/ "",
+        /* filter=*/ "",
+        operations);
+    // we should not be at the end
+    assertThat(nextToken).isNotEqualTo("");
+    assertThat(operations.build().size()).isEqualTo(1);
```
ola-rozenfeld: It seems you can merge this test case with the next one without making them less clear (you expect to get 1, and then another one).

PR #137, opened 2018-04-05
Produce docker containers for both server and worker

timperrett:
Hi there!

Wanted to deploy build farm using containers, figured i'd just toss up a PR to do the containerization work as it might be useful for others. Present thinking is that it would make sense to put in a custom entrypoint that can either take some params as env vars (e.g. name of config file), or otherwise always expect files mounted at a particular location as it is now (e.g. `/config/server.config`). Obviously both approaches come with downsides :-) 

This is a first pass knocked up this evening, so i'd welcome some feedback 😄 
#

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 bazel-*
+.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 bazel-*
-.DS_Store
--- a/src/main/java/build/buildfarm/BUILD
+++ b/src/main/java/build/buildfarm/BUILD
@@ -107,6 +107,26 @@ java_binary(
     ],
 )
 
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
+
+container_image(
+    name = "server",
+    base = "@java_base//image",
+    # leverage the implicit target of the buildfarm-server to get a fat jar.
+    # this is simply a workaround for the fact that we have so many dependencies,
+    # so we'd want some wrappy script. This seemed more straightforward.
+    # https://docs.bazel.build/versions/master/be/java.html#java_binary_implicit_outputs
+    files = [
+        "//src/main/java/build/buildfarm:buildfarm-server_deploy.jar"
+    ],
+    cmd = [
+        "buildfarm-server_deploy.jar",
+        "/config/server.config",
+        "--port",
+        "8098"
+    ],
+)
+
 java_library(
     name = "stub-instance",
     srcs = glob(["instance/stub/*.java"]),
--- a/src/main/java/build/buildfarm/BUILD
+++ b/src/main/java/build/buildfarm/BUILD
@@ -110,7 +110,7 @@ java_binary(
 load("@io_bazel_rules_docker//container:container.bzl", "container_image")
 
 container_image(
-    name = "server",
+    name = "server.container",
     base = "@java_base//image",
     # leverage the implicit target of the buildfarm-server to get a fat jar.
     # this is simply a workaround for the fact that we have so many dependencies,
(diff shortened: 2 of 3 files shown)


src/main/java/build/buildfarm/BUILD
```diff
@@ -179,3 +199,19 @@ java_binary(
...
+
+container_image(
+    name = "worker.container",
+    base = "@java_base//image",
+    # leverage the implicit target of the buildfarm-server to get a fat jar.
+    # this is simply a workaround for the fact that we have so many dependencies,
+    # so we'd want some wrappy script. This seemed more straightforward.
+    # https://docs.bazel.build/versions/master/be/java.html#java_binary_implicit_outputs
+    files = [
+        "//src/main/java/build/buildfarm:buildfarm-worker_deploy.jar",
```
werkt: likewise, shorten the target

src/main/java/build/buildfarm/BUILD
```diff
@@ -179,3 +199,19 @@ java_binary(
...
+    ],
+    cmd = [
+        "buildfarm-worker_deploy.jar",
+        "/config/worker.config",
+    ],
+)
```
werkt: newline

.gitignore
```diff
@@ -1 +1,2 @@
 bazel-*
+.DS_Store
```
werkt: what is this?

RNabel: This is a macos-only file placed in every directory -- usually added when running `git add .` instead of something more fine-grained.

timperrett: correct - OSX is just adding these to every dir, so I exclude them by default to ensure they do not inadvertently make their way into the checked-in codebase.

werkt: this is an OS specific artifact, and not one incurred by building the software, please remove it.

src/main/java/build/buildfarm/BUILD
```diff
@@ -2,6 +2,8 @@ package(
     default_visibility = ["//src:__subpackages__"],
 )
 
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
```
werkt: https://docs.bazel.build/versions/master/skylark/build-style.html#file-structure

I appreciate you moving it, but please refer to the above for ordering: load should precede `package`

.gitignore
oscar-stripe: it would be really nice to keep it since otherwise OSX users will see this and accidentally add the file repeatedly.

What are your guidelines for what should be allowed in .gitignore? This is a very common entry to add.

andyscott: FWIW this ignore isn't added to bazelbuild/bazel, bazelbuild/rules_kotlin, or bazelbuild/rules_go.

werkt:
I missed the upfront request for feedback, I'm used to only seeing commit summaries in PR contents.

The container definition itself looks fine, I'm personally a fan of providing custom mounts (and this meshes well with kubernetes config maps as well) for config files and just targeting a specific location universally - my current (manual) container deployment of buildfarm uses the same. The only thing I probably would have changed would be to drop the `--port` specification, since the config allows the specification of the listen port, but given how container port mapping works, I suppose it doesn't matter.

I haven't had a chance to try this out myself, but its definitely on my TODO list, and will become integral to my deployment.

andyscott:
@timperrett I haven't tried this yet, but it's on my radar to try in a few weeks (once we finish our current iteration of Bazel caching work).

timperrett:
@werkt cool, thank you :-) Looking forward to sending more contributions - is there a list of stuff that needs doing?

werkt:
@timperrett - the issues list has a couple of items that might be easily done - I will try to tag appropriately - and I'm going to put up a project under this repo with the goal of 100% code coverage.

PR #146, opened 2018-05-29
Remove DigestUtil from Chunker interface

werkt:
Remove the requirement that the Chunker use a DigestUtil instance to
produce an empty chunk. Suppressed several places where empty blobs were
being interpreted or uploaded to filter those behaviors, including
findMissingBlobs, action validation, ByteStreamUploader, and others.


src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -286,28 +286,31 @@ private void validateActionInputDirectory(
...
-    path.push(inputRootDigest);
-
-    Directory root = expectDirectory(inputRootDigest);
-    validateActionInputDirectory(root, path, new HashSet<>(), inputDigests);
+  private void validateActionInputDirectory(
+      Digest directoryDigest,
+      Stack<Digest> path,
+      Set<Digest> visited,
+      ImmutableSet.Builder<Digest> inputDigests) {
+    if (directoryDigest.getSizeBytes() != 0 || Directory.getDefaultInstance().toByteString().size() != 0) {
```
ola-rozenfeld: Wait, an empty dir digest means an empty directory, because Directory.getDefaultInstance().toByteString() serializes to an empty string. And if it didn't, I don't quite understand this condition -- wouldn't it mean an empty digest is not a valid directory? Also, don't you want to add the empty digest to visited?
Also, I think it be safer to optimize for empty digests as close to the upload/download as possible -- in the getBlob function, for instance, instead of finding every place that uses downloads directly or indirectly. That way ensures things don't break when anyone anywhere asks to get an empty blob from CAS.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -322,17 +325,19 @@ protected void validateAction(Action action) {
...
     filesUniqueAndSortedPrecondition(action.getOutputDirectoriesList());
-    Command command;
-    try {
-      command = Command.parseFrom(getBlob(commandDigest));
-    } catch (InvalidProtocolBufferException ex) {
-      Preconditions.checkState(
-          false,
-          INVALID_DIGEST);
-      return;
+    if (commandDigest.getSizeBytes() != 0 || Command.getDefaultInstance().toByteString().size() != 0) {
```
ola-rozenfeld: I think an empty command is a good reason for an INVALID_ARGUMENT.

src/main/java/build/buildfarm/instance/stub/Chunker.java
```diff
@@ -171,7 +167,7 @@ public Chunk next() throws IOException {
 
     if (digest.getSizeBytes() == 0) {
       data = null;
-      return emptyChunk;
+      return new Chunk(Digest.getDefaultInstance(), ByteString.EMPTY, 0);
```
ola-rozenfeld: It is a bit of a hack, IIUC. Basically, this works because your usage of Chunker now ensures that Chunkers with empty data never get uploaded, so it doesn't matter what the digest hash is, as long as size_bytes is 0. But I don't have a better idea :-)

src/test/java/build/buildfarm/server/BuildFarmServerTest.java
```diff
@@ -306,19 +306,19 @@ public void actionWithExcessiveTimeoutFailsValidation()
...
+    Digest rootBlobDigest = digestUtil.compute(root);
     Action action = Action.newBuilder()
-        .setCommandDigest(digestUtil.compute(commandBlob))
-        .setInputRootDigest(digestUtil.compute(rootBlob))
+        .setCommandDigest(commandBlobDigest)
+        .setInputRootDigest(rootBlobDigest)
         .build();
     ByteStreamUploader uploader = new ByteStreamUploader("memory", inProcessChannel, null, 60, Retrier.NO_RETRIES, null);
 
     uploader.uploadBlobs(ImmutableList.of(
```
ola-rozenfeld: So this line now doesn't do anything, right?

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: Yeah, this is unnecessarily complicated, and was a hedge for the first run for me against getDefaultInstance() serializing to an empty string. If an empty string was not a serialization of the default instance, I wanted it to be proven invalid by the Directory.parseFrom. I'll simplifiy it and make sure that we're safe for upload/download and that we guard/optimize on either side of the interface.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: I'll buy that as well, since we don't believe any representation of a message and its [lack of] contents are valid arbitrarily.

src/main/java/build/buildfarm/instance/stub/Chunker.java
werkt: Yeah, I didn't want to be checking for valid permutations of empty hashes everywhere, and forcing everything to know about that instance-configured method

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -325,19 +333,22 @@ protected void validateAction(Action action) {
...
-      } catch (InvalidProtocolBufferException ex) {
-        Preconditions.checkState(
-            false,
-            INVALID_DIGEST);
-        return;
-      }
-      environmentVariablesUniqueAndSortedPrecondition(
-          command.getEnvironmentVariablesList());
+
+    Preconditions.checkState(commandDigest.getSizeBytes() != 0, INVALID_DIGEST);
```
ola-rozenfeld: A bit redundant, this would be handled by line 350 as well.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -132,8 +132,14 @@ public InputStream newStreamInput(String name) {
 
   @Override
   public Iterable<Digest> findMissingBlobs(Iterable<Digest> digests) {
-    return Iterables.filter(
-        digests, digest -> digest.getSizeBytes() != 0 && !contentAddressableStorage.contains(digest));
+    ImmutableList.Builder<Digest> missingBlobs = new ImmutableList.Builder<>();
```
ola-rozenfeld: Can all of getBlob, putBlob, and findMissingBlobs have the respective 0 case workaround in the AbstractServerInstance? IIUC, that will ensure that no matter which CAS implementation is used in the config, the invariant that 0-length blobs always virtually exist will be maintained. Basically, I'm just trying to insert this at a level such that the code is both guaranteed to exist and won't need to be duplicated.

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
werkt: Agreed, was thinking too much in a null pointer mindset.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: Further agreed, I am not 100% sure why I had the MemoryInstance hold the findMissingBlobs impl, but perhaps I just missed it my initial refactor.

PR #151, opened 2018-06-06
Remove deprecated usage of native http/git rules

edbaunton:
Per https://groups.google.com/forum/#!msg/bazel-discuss/dO2MHQLwJF0/2OXHjMAaAAAJ this should be mechanical.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,8 @@
 workspace(name = "build_buildfarm")
 
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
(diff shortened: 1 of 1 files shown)


edbaunton:
I am curious to see if the CI passes as locally I ran into the error:

`Not a regular file: /Users/user/bazel-buildfarm/external/BUILD.googleapis`

Which looks like a bug in `0.14` (it's inserting 'external') that has since been fixed (but I can't find any evidence of the bug ever being reported); unless anyone knows what I might be doing wrong?

werkt:
Yup, CI breaks - build_file may expect a repo specifier: `"@//:BUILD.googleapis"` perhaps

PR #152, opened 2018-06-07
S Autoconfigure worker gRPC context with SSL for target port 443

pcj:
Allows worker to connect to a server process over SSL


pcj:
I'm thinking this PR probably does not make a ton of sense to be merged.  It allows a worker to connect to a server sitting behind a TLS-terminated ingress, but that's really not a real-world use case.

Nevertheless I figure I'd share it anyway and leave it here as a tombstone for other people that might want to take bits and pieces of it for other use-cases.

PR #157, opened 2018-06-13
S Guard against NullPointer Command/Action fetch

werkt:
Handle a nonexistent command or action fetch.

### DIFF
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -481,11 +481,6 @@ public abstract class AbstractServerInstance implements Instance {
         ExecuteOperationMetadata.Stage.EXECUTING;
   }
 
-  protected boolean isComplete(Operation operation) {
-    return expectExecuteOperationMetadata(operation).getStage() ==
-        ExecuteOperationMetadata.Stage.COMPLETED;
-  }
-
   abstract protected boolean matchOperation(Operation operation);
   abstract protected void enqueueOperation(Operation operation);
 
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -445,9 +445,13 @@ public abstract class AbstractServerInstance implements Instance {
   }
 
   protected Action expectAction(Operation operation) {
+    ByteString actionBlob = getBlob(
+        expectExecuteOperationMetadata(operation).getActionDigest());
+    if (actionBlob == null) {
+      return null;
+    }
     try {
-      return Action.parseFrom(getBlob(
-          expectExecuteOperationMetadata(operation).getActionDigest()));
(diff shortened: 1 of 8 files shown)


src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -445,9 +445,13 @@ protected ExecuteOperationMetadata expectExecuteOperationMetadata(
   }
 
   protected Action expectAction(Operation operation) {
```
ola-rozenfeld: I don't see you doing anything upstream to handle the possible null value returned here, like you do for command in validateAction. Maybe it would be simpler for all expectX functions to either return the X or throw a Precondition failure. That way you only have to handle it in one place.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -46,10 +46,16 @@
...
+    ByteString commandBlob = workerContext.getBlob(operationContext.action.getCommandDigest());
+    Command command = null;
+    if (commandBlob != null) {
+      try {
+        command = Command.parseFrom(commandBlob);
+      } catch (InvalidProtocolBufferException ex) {
+      }
+    }
+
+    if (command == null) {
```
ola-rozenfeld: Is there a way to pass the error upstream? Ideally with the missing digest. Ditto with the action.

werkt: Currently, the path for this goes like this: The error stage drops the operationContext on the floor, the requeuer on the server notices the lack of updates and queues the operation again (without a check for validity, adding that in an update), and the cycle repeats. There aren't any great channels available for this, except for the server logs, which I'm woefully behind in terms of supplying any content for. Were you thinking of something more structured? Otherwise I'd like to hold off on this until I have put better logging into place.

src/main/java/build/buildfarm/server/OperationQueueService.java
```diff
@@ -88,12 +88,18 @@ public void put(
...
-    responseObserver.onCompleted();
+    try {
+      boolean ok = instance.putOperation(operation);
+      Code code = ok ? Code.OK : Code.UNAVAILABLE;
+      responseObserver.onNext(com.google.rpc.Status.newBuilder()
+          .setCode(code.getNumber())
+          .build());
+      responseObserver.onCompleted();
+    } catch (IllegalStateException e) {
+      responseObserver.onError(Status.FAILED_PRECONDITION
```
ola-rozenfeld: We also want to send a PreconditionFailure message in the status error_detail field, with a Violation of type MISSING and the right digest where appropriate (according to the API spec), so that the client can reupload+retry that.
I think the simplest way of doing that would be to raise and handle a CacheNotFoundException.

werkt: Ah, yeah.  I was going to switch to making a PreconditionViolationException or some such to do those aggregates and lump as many issues as possible into one response

PR #158, opened 2018-06-14
XS Avoid CASFileCache ByteStreamService interactions for empty blobs

pcj:


pcj:
OK, I'll close this one.

PR #162, opened 2018-06-15
S Handle IOException in fetchDirectory with rollback

werkt:
An IOException encountered during fetchDirectory should be met with a
rollback of a putDirectory operation. References to inputs are
decremented, and the target directory is removed, recursively.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -242,7 +242,7 @@ public class CASFileCache {
   private void expireDirectory(Digest digest) throws IOException {
     DirectoryEntry e = directoryStorage.remove(digest);
     purgeDirectoryFromInputs(digest, e.inputs);
-    CASFileCache.removeDirectory(getDirectoryPath(digest));
+    removeDirectory(getDirectoryPath(digest));
   }
 
   /** must be called in synchronized context */
@@ -318,7 +318,7 @@ public class CASFileCache {
         purgeDirectoryFromInputs(digest, inputs);
         decrementReferencesSynchronized(inputs, ImmutableList.<Digest>of());
       }
-      CASFileCache.removeDirectory(path);
+      removeDirectory(path);
       throw e;
     }
 
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -189,7 +189,8 @@ public class CASFileCache {
     return root.resolve(filename);
   }
 
-  private Path getDirectoryPath(Digest digest) {
+  @VisibleForTesting
+  public Path getDirectoryPath(Digest digest) {
     return root.resolve(digestFilename(digest));
   }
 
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -222,18 +226,21 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
       }
     }
-    removeDirectory(path);
+  }
+
+  /** must be called in synchronized context */
+  private void expireDirectory(Digest digest) throws IOException {
+    DirectoryEntry e = directoryStorage.remove(digest);
+    purgeDirectoryFromInputs(digest, e.inputs);
+    CASFileCache.removeDirectory(getDirectoryPath(digest));
```
ola-rozenfeld: nit: why reference this by module name when it's the same module?

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -272,21 +279,33 @@ private void fetchDirectory(
...
 
     ImmutableList.Builder<Path> inputsBuilder = new ImmutableList.Builder<>();
-    fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
+    try {
+      fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
+    } catch (IOException e) {
+      ImmutableList<Path> inputs = inputsBuilder.build();
+      synchronized (this) {
+        purgeDirectoryFromInputs(digest, inputs);
+        decrementReferencesSynchronized(inputs, ImmutableList.<Digest>of());
```
ola-rozenfeld: Wait, so who calls directoryStorage.remove(digest)? It should be called, right?

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -272,21 +279,33 @@ private void fetchDirectory(
...
+        incrementReferences(e.inputs);
+        return path;
+      }
     }
 
     ImmutableList.Builder<Path> inputsBuilder = new ImmutableList.Builder<>();
-    fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
+    try {
+      fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
+    } catch (IOException e) {
```
ola-rozenfeld: What if there's another IOException while deleting the directory? You probably don't want it to hide the original exception.

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: I had it in my head that the linter threw a warning on this, guess it's just for !same-module

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: we haven't yet added this digest to the directoryStorage at this point, this is all that is required to effect the rollback. The only call to `directoryStorage.remove` is in `expireDirectory`

ola-rozenfeld: /facepalm
yep, you're right, sorry!

src/main/java/build/buildfarm/worker/CASFileCache.java
werkt: Agreed. Added a catch with a printStackTrace and rerouted.

PR #163, opened 2018-06-15
XS Reduce fileCache putDirectory sync region

werkt:
The synchronized region for putDirectory is reduced to prevent inclusion
of CAS fetches.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -270,7 +270,6 @@ public class CASFileCache {
     }
   }
 
-  /** must be called in synchronized context */
   private void fetchDirectory(
       Digest containingDirectory,
       Path path,
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -290,21 +290,23 @@ public class CASFileCache {
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -272,21 +272,23 @@ private void fetchDirectory(
...
+    synchronized (this) {
+      DirectoryEntry e = directoryStorage.get(digest);
+      if (e != null) {
+        incrementReferences(e.inputs);
+        return path;
+      }
     }
 
     ImmutableList.Builder<Path> inputsBuilder = new ImmutableList.Builder<>();
     fetchDirectory(digest, path, digest, directoriesIndex, inputsBuilder);
```
ola-rozenfeld: So the comment on the fetchDirectory function is not true? Why, what changed? It does look like it needs sync to me.

werkt: The comment is now extraneous - I don't see any issues with calling fetchDirectory from an cache-wide-unsynchronized context. There is, however a race when two misses on the same key occur, and when similar things happen to files via put - I will have to add a path-based synchronization from a separate unit of work.

werkt: Will do. The synchronization is largely academic here, since we only do puts of files or directories from the single threaded input stage in the operationqueue worker. That's about to shift substantially...

PR #173, opened 2018-07-01
Remove the worker container image.

EdSchouten:
The container image for the worker process that we currently ship is
effectively useless. It's based on a distroless container image, meaning
it doesn't even include a shell interpreter. Because of this, it is not
able to spawn any subprocesses.

People who want to run workers through Docker should create their own
container images with the development tools they need.

### DIFF
--- a/src/main/java/build/buildfarm/BUILD
+++ b/src/main/java/build/buildfarm/BUILD
@@ -202,20 +202,3 @@ java_binary(
     ],
     visibility = ["//visibility:public"],
 )
-
-container_image(
-    name = "worker.container",
-    base = "@java_base//image",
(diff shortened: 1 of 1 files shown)


werkt:
There is a case for this to say that the worker *shouldn't* come with anything on the image by default - all actions should have declared inputs that would be shipped to the execution environment on the worker. Otherwise, you don't have hermetic actions. You are free to declare your own worker image constructs with other contents that still run the workers.

werkt:
Closing this after a year, the container is mostly useless though, so it should probably be removed and relocated (possibly with a guide book for how to do a deploy incorporating tini and process-wrapper)

PR #178, opened 2018-08-19
Update remote execution API to v2

werkt:


ola-rozenfeld:
Also: I screwed up, sorry, you'll need https://github.com/bazelbuild/bazel/pull/5934 as well. Only now discovered I forgot to define the java package in the semver.proto (our Go server didn't complain...).

werkt:
Forgive my ignorance, but how are you supporting 10MB batch sizes if the GRPC limit is 4MB? just over capped?

ola-rozenfeld:
No, that's a good question -- I remember looking into this limit thing further and discovering it wasn't trivial, but I don't remember what I found. I'll get back to you on this! We might be doing it wrong.
In any case, 4mb will work for sure.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -450,18 +462,15 @@ private boolean satisfiesRequirements(Platform platform, Operation operation) {
...
-      if (operation == null || operation.getDone()) {
-        // watcher did not process completed state
-        return false;
-      }
+    Operation operation = getOperation(operationName);
+    if (!watcher.test(operation)) {
+      // watcher processed completed state
+      return true;
+    }
+    if (operation == null || operation.getDone()) {
```
ola-rozenfeld: How can this even happen? I see the watcher returns falls for these cases.

src/main/java/build/buildfarm/server/ExecutionService.java
```diff
@@ -29,6 +32,58 @@ public ExecutionService(Instances instances) {
...
+  private Predicate<Operation> createWatcher(StreamObserver<Operation> responseObserver) {
+    return (operation) -> {
+      if (Context.current().isCancelled()) {
+        return false;
+      }
+      try {
+        if (operation != null) {
+          responseObserver.onNext(operation);
+        }
+        if (operation == null || operation.getDone()) {
```
ola-rozenfeld: if you move the operation == null condition into line 37, then you don't need to check for it in the rest of the function and the code gets slightly simpler.

src/main/java/build/buildfarm/worker/MatchStage.java
```diff
@@ -64,7 +65,19 @@ private boolean fetch(Operation operation) {
 
     try {
       action = Action.parseFrom(actionBlob);
-    } catch (InvalidProtocolBufferException ex) {
+    } catch (InvalidProtocolBufferException e) {
+      return false;
+    }
+
+    ByteString commandBlob = workerContext.getBlob(action.getCommandDigest());
+    if (actionBlob == null) {
```
ola-rozenfeld: commandBlob

src/main/java/build/buildfarm/worker/UploadManifest.java
```diff
@@ -163,29 +163,16 @@ public void addContent(ByteString content, CASInsertionPolicy policy, Consumer<B
   }
 
   private void addFile(Path file, long size, CASInsertionPolicy policy) throws IOException {
```
ola-rozenfeld: Do you still need to pass the size externally? I think DigestUtil should handle computing the digest correctly, regardless of whether the file is a symlink or not.

src/test/java/build/buildfarm/instance/memory/MemoryInstanceTest.java
```diff
@@ -225,9 +211,9 @@ public void watchOperationAddsWatcher() throws InterruptedException {
     watchers.put(operation.getName(), operationWatchers);
 
     Predicate<Operation> watcher = (Predicate<Operation>) mock(Predicate.class);
+    when(watcher.test(eq(operation))).thenReturn(true);
```
ola-rozenfeld: only reason to mock a predicate is if you're going to verify it. Otherwise could just use a lambda.

src/test/java/build/buildfarm/server/BuildFarmServerTest.java
```diff
@@ -306,15 +309,20 @@ public void cancellingExecutingOperationFailsPoll()
...
+  }
 
-    executeAction(actionWithExcessiveTimeout);
+  private Digest createSimpleAction() throws RetryException, InterruptedException {
+    return createAction((action) -> action);
   }
 
-  private Action createSimpleAction() throws RetryException, InterruptedException {
-    DigestUtil digestUtil = new DigestUtil(DigestUtil.HashFunction.SHA256);
+  private Digest createAction(Function<Action, Action> onAction) throws RetryException, InterruptedException {
```
ola-rozenfeld: Consider refactoring to accepting the action builder as a parameter instead and using it as a "starter" in line 332, instead of Action.newBuilder().

ola-rozenfeld:
Oh, and to the 4MB question: um, yes, it is totally 4MB, I checked it fails if I try to stuff more into it. Don't know now what made us change it to 10, but this will not work, and we should fix it :-)

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -402,21 +406,30 @@ protected boolean matchOperation(Operation operation) {
...
       while (!matched && !queuedOperations.isEmpty()) {
         Operation operation = queuedOperations.remove(0);
-        if (satisfiesRequirements(platform, operation)) {
-          matched = true;
-          if (onMatch.test(operation)) {
-            onDispatched(operation);
-          } else if (!requeueOnFailure) {
+        Command command = expectCommand(operation);
+        if (command == null) {
+          cancelOperation(operation.getName());
```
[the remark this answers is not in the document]
werkt: Low/load-induced probability race - this is the long tail operation queue, where operations may have been sitting for an indeterminate amount of time, waiting for a worker to be ready to execute.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -298,25 +299,31 @@ public String getTree(
...
         .get()
         .getTree(GetTreeRequest.newBuilder()
             .setInstanceName(getName())
             .setRootDigest(rootDigest)
             .setPageSize(pageSize)
             .setPageToken(pageToken)
             .build());
-    directories.addAll(response.getDirectoriesList());
-    return response.getNextPageToken();
+    // new streaming interface doesn't really fit with what we're trying to do here...
```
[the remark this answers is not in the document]
werkt: The effect of the new streaming interface on this is actually less than you think - I'm benefitting here because what used to take multiple calls to getTree with distinct page tokens can now be fed completely, ending on a single empty nextPageToken to complete the sequence. The only call to this method is in worker/operationqueue/Worker.java, where the intended effect was to fetch the entire tree, which this is, as you say, completely in spec for.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: I preferred to be decoupled from any watcher implementation, in particular one that would be unconcerned with the null-ness or done-ness of the operation it is watching. In those cases, I can do nothing further in this function.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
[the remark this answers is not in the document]
ola-rozenfeld: > what used to take multiple calls to getTree with distinct page tokens can now be fed completely

Yes, that's what I don't see here -- at which point are you replacing a loop of getTree calls with a single getTree call?  The only service implementation I see still calls .onNext() just once once before .onCompleted().

src/main/java/build/buildfarm/server/ExecutionService.java
werkt: But in the null operation case, I still need to complete the responseObserver stream. Moving it out of the try block here will also prevent coverage by the grpc induced cancellation/expiration handlers present in the catches.

ola-rozenfeld: Oh, sorry, you're right! Disregard.

src/main/java/build/buildfarm/worker/UploadManifest.java
werkt: nope, cleaning that up.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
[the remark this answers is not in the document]
werkt: The only call to this particular function is
src/test/java/build/buildfarm/server/BuildFarmServerTest.java:314
which is unmodified from the previous implementation because the instance interface hasn't changed (it just manifests as a single call in the case of a successful stream, and expects that a non-empty nextPageToken will be returned for any retriable error).

Gotcha regarding the service implementation, yeah I didn't change that to take advantage. Updating now.

src/main/java/build/buildfarm/server/ContentAddressableStorageService.java
```diff
@@ -129,13 +130,18 @@ public void getTree(
...
-        request.getRootDigest(), pageSize, request.getPageToken(), directories);
 
-    responseObserver.onNext(GetTreeResponse.newBuilder()
-        .addAllDirectories(directories.build())
-        .setNextPageToken(nextPageToken)
-        .build());
+    String pageToken = request.getPageToken();
+    do {
+      String nextPageToken = instance.getTree(
+          request.getRootDigest(), pageSize, pageToken, directories);
```
ola-rozenfeld: You want to clear directories between every onNext. I don't see https://github.com/uber/bazel-buildfarm/blob/working/src/main/java/build/buildfarm/instance/AbstractServerInstance.java#L183 doing that.

Also: while this is better than before, it could be made even more efficient by refactoring instance.getTree to return everything at once instead of repeatedly calling it here. The problem with calling it repeatedly is that you're repeatedly fetching the entire tree, and then slicing out just one portion of it. It's still O(n^2) overall, even though it's all in memory so you could ignore it, I guess.

werkt: Missed the directories builder reset, thanks.

You should take another look at my implementation of things here - I'm not refetching the entire tree on each call to MemoryInstance::getTree - my page token contains a path hierarchy of digests and my restarts are logarithmic.

PR #187, opened 2018-09-21
Make memory collections thread safe

werkt:
Ensure that requeuers, outstanding operations, and other derivative
collections in MemoryInstance are thread safe per their specifications.

Fixes #184, #185

PR #189, opened 2018-10-05
(XS) Use a synchronized sorted map for Outstanding Ops

werkt:
Prevents unsafe access to the outstanding operations, which can yield
inconsistent responses to execute requests - null operations passed to
watcher routines which result in no-response stream response
completions.


werkt:
Closing this, duplicate of #187

PR #191, opened 2018-10-23
M Execution logging

werkt:
With an addition to the README.md to use an example logging
configuration to provide some feedback on executions.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -16,10 +16,11 @@ package build.buildfarm.worker;
 
 import static com.google.common.collect.Iterables.filter;
 import static com.google.common.collect.Iterables.transform;
-import static com.google.common.collect.Lists.newArrayList;
 import static build.buildfarm.v1test.ExecutionPolicy.PolicyCase.WRAPPER;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
 
 import build.buildfarm.v1test.ExecutionPolicy;
+import com.google.common.base.Stopwatch;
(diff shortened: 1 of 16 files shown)


src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -190,15 +210,14 @@ private Code executeCommand(
...
-    int exitValue = -1;
     Process process;
     try {
       process = processBuilder.start();
       process.getOutputStream().close();
     } catch(IOException ex) {
       ex.printStackTrace();
       // again, should we do something else here??
-      resultBuilder.setExitCode(exitValue);
+      resultBuilder.setExitCode(exitCode);
```
ola-rozenfeld: Sorry, I'm having a code-read fail -- can the exitCode here ever be not -1?

werkt: It cannot, the exitCode assignments after the default initialization all occur after this. lmk if you think that merits different behavior here

ola-rozenfeld: Oh, then I would very much prefer you just say -1 here, to not make the reader think about what the value might be.

werkt: Pedantically, I used the exitCode here to match the releaseExecutor information propagated to the ExecuteActionStage so that there would not be any discontinuity between what is [logged](https://github.com/bazelbuild/bazel-buildfarm/pull/191/files/5009aa01810a4f392e314db27561f5d4af2efa10#diff-830c4529fc2f6fa022b5fad4505c0507R164) and what is recorded in the resultBuilder, but I see the point.

PR #194, opened 2018-11-11
M Requeue explicitly as a result of stage failures

werkt:
Each stage must be capable of requeuing an operation that has failed
during processing in an incomplete execution.

### DIFF
--- a/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/AbstractServerInstance.java
@@ -88,7 +88,7 @@ public abstract class AbstractServerInstance implements Instance {
   private static final String MISSING_INPUT =
       "A requested input (or the `Command` of the `Action`) was not found in the CAS.";
 
-  private static final String INVALID_DIGEST = "A `Digest` in the input tree is invalid.";
+  public static final String INVALID_DIGEST = "A `Digest` in the input tree is invalid.";
 
   private static final String INVALID_FILE_NAME =
       "One of the input `PathNode`s has an invalid name, such as a name containing a `/` character"
@@ -631,8 +631,16 @@ public abstract class AbstractServerInstance implements Instance {
     if (action == null) {
       return null;
     }
+    return expectCommand(action.getCommandDigest());
+  }
+
+  protected Command expectCommand(Digest commandDigest) {
+    ByteString commandBlob = getBlob(commandDigest);
+    if (commandBlob == null) {
+      return null;
+    }
     try {
-      return Command.parseFrom(getBlob(action.getCommandDigest()));
+      return Command.parseFrom(commandBlob);
     } catch (InvalidProtocolBufferException ex) {
       return null;
     }
@@ -786,35 +794,33 @@ public abstract class AbstractServerInstance implements Instance {
     return putOperation(operation);
   }
 
-  @Override
+  @VisibleForTesting
   public boolean requeueOperation(Operation operation) throws InterruptedException {
     String name = operation.getName();
     ExecuteOperationMetadata metadata = expectExecuteOperationMetadata(operation);
     if (metadata == null) {
       // ensure that watchers are notified
+      String message = String.format(
+          "Operation %s does not contain ExecuteOperationMetadata",
+          name);
       errorOperation(name, com.google.rpc.Status.newBuilder()
           .setCode(com.google.rpc.Code.FAILED_PRECONDITION.getNumber())
+          .setMessage(message)
           .build());
-      throw new IllegalStateException(
-          String.format(
-              "Operation %s does not contain ExecuteOperationMetadata",
-              name));
+      return false;
     }
-
-    // valid metadata should mean that we revalidate in dispatched
-    Preconditions.checkState(
-        metadata.getStage() == Stage.QUEUED,
-        String.format("Operation %s stage is not QUEUED", name));
 
     Digest actionDigest = metadata.getActionDigest();
 
     Action action = expectAction(actionDigest);
 
     try {
+      Preconditions.checkState(action != null, MISSING_INPUT);
       validateAction(action);
     } catch (IllegalStateException e) {
       errorOperation(name, com.google.rpc.Status.newBuilder()
           .setCode(com.google.rpc.Code.FAILED_PRECONDITION.getNumber())
+          .setMessage(e.getMessage())
           .build());
       return false;
     }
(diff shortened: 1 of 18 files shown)


src/main/java/build/buildfarm/server/OperationQueueService.java
```diff
@@ -63,6 +64,8 @@ public OperationQueueService(Instances instances) {
           responseObserver.onError(ex);
```
ola-rozenfeld: I'm suddenly confused -- what can ever throw a StatusRuntimeException here? If it's the responseObserver itself, then calling its onError method will not work either.

In any case -- if we had an error here, why do we want to requeue the operation? Wouldn't it just be the exact same error the next time around?

This question applies in general to the whole premise of the requeing logic. If you want to do it as a retry layer inside Buildfarm, shouldn't you have a hard cap on them to not get into an infinite loop? And in general -- Bazel already has retries, so if you just return an INTERNAL_ERROR when something goes wrong on the server, Bazel will already retry with exponential decay (which is more likely to resolve the error) and likely hitting a different server and worker processes, which is also more likely to resolve the error. I'm wary of multiple nested retries, in general -- they often lead to way more executions than was expected.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
```diff
@@ -396,7 +397,13 @@ public boolean putOperation(Operation operation) throws InterruptedException {
             .setNanos(actionTimeout.getNanos() + delay.getNanos())
             .build();
         // this is an overuse of Watchdog, we will never pet it
-        Watchdog operationTimeoutDelay = new Watchdog(timeout, () -> expireOperation(operation));
+        Watchdog operationTimeoutDelay = new Watchdog(timeout, () -> {
+          try {
+            expireOperation(operation);
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
ola-rozenfeld: Where is the interrupted flag tested?

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -327,34 +328,14 @@ public void execute(
...
+    Operation operation = operationQueueBlockingStub.get()
+        .take(TakeOperationRequest.newBuilder()
         .setInstanceName(getName())
         .setPlatform(platform)
         .build());
-    boolean successful = onMatch.test(operation);
-    if (!Thread.currentThread().isInterrupted() && !successful && requeueOnFailure) {
-      requeue(operation);
-    }
+    boolean success = onMatch.test(operation);
```
ola-rozenfeld: so the return value is not used?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -214,9 +214,6 @@ message WorkerConfig {
   // policy for process output files
   CASInsertionPolicy file_cas_policy = 13;
 
-  // whether the worker should attempt to requeue
-  // an operation when an unexpected failure
-  // occurs during an execution
   bool requeue_on_failure = 14;
```
ola-rozenfeld: So this is now deprecated, yes?

src/main/java/build/buildfarm/instance/AbstractServerInstance.java
```diff
@@ -777,23 +778,34 @@ public void cancelOperation(String name) throws InterruptedException {
...
       throw new IllegalStateException(
           String.format(
               "Operation %s does not contain ExecuteOperationMetadata",
               name));
     }
 
+    // valid metadata should mean that we revalidate in dispatched
+    Preconditions.checkState(
+        metadata.getStage() == Stage.QUEUED,
+        String.format("Operation %s stage is not QUEUED", name));
```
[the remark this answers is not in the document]
werkt: agreed, I'll clean this up.

src/main/java/build/buildfarm/server/OperationQueueService.java
werkt: The SRE happens when the client has disconnected while waiting for a response. It only happens with CANCELLED status in this case, which is why I tested for all other results, and funnelled them to the responseObserver, because the caller's context (the instance) is not, strictly speaking, equipped or expecting to deal with grpc SREs - similar to doing one thing, it only expects to hear about success or failure to dispatch.

You're right about the requeue here though - It belongs back in the match implementation, which handles the long-lived operations that should redetermine whether a mismatched operation that fails to plaform or dispatch should be requeued.

Per the hard cap or fail fast mechanism consideration, the error recovery personality of bazel has obviously been fairly erratic, and the requeing logic functions more akin to error-correction for a worker-initiated requeue or a watcher requeue, or invariance for possibly long-period event sequences where a queue entry may no longer be valid (without having a worker to experimentally determine such). I would be concerned about setting a hard cap here where the requeues/op may become routine in, for instance, diverse operation/worker platform environments.

src/main/java/build/buildfarm/instance/memory/MemoryInstance.java
werkt: It isn't, and the param is already an InterruptingRunnable, reducing.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: nope. the stub client just takes from the queue, with the same semantics of the match interface, and leaves any failure recovery to the caller. publicising the requeue method would extend the instance interface unnecessarily, calling someone else's requeue method would add a mutual dependency for other requeuers. I might be convinced that this deserves a `take` method with runnables instead of predicates, but this was convenient.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
werkt: yup, reserving

src/main/java/build/buildfarm/instance/stub/StubInstance.java
werkt: Switched this to using a runnable for now

werkt:
Could use a re-review on this when you're able, @ola-rozenfeld

src/main/java/build/buildfarm/server/OperationQueueService.java
ola-rozenfeld: Is there anything now that prevents reproducibly failing operations (from a future bug, for example) to be requeued forever? Would such an operation be dropped at any point?

werkt: Nothing intrinsic to the requeuing behavior, no. The operation will be dropped systemically via expiration on the underlying cas delegate if there is sufficient inflow, and we're also obviously bound by the lifetime of the server itself.

PR #231, opened 2019-03-07
Fix for --incompatible_depset_is_not_iterable

### LINKED ISSUE #230: --incompatible_depset_is_not_iterable
Buildfarm fails when we set `--incompatible_depset_is_not_iterable` in Bazel: https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/31#bd9cdfb2-f1a2-4aba-99fc-d762b48caf91

Updating your dependencies might be enough to fix it.

werkt:
Fixes #230


third_party/grpc_java/java_grpc_library.bzl
```diff
@@ -18,7 +18,7 @@ def _gensource_impl(ctx):
   srcdotjar = ctx.actions.declare_file(ctx.label.name + "_src.jar")
 
   srcs = [f for dep in ctx.attr.srcs for f in dep[ProtoInfo].direct_sources]
-  includes = [f for dep in ctx.attr.srcs for f in dep[ProtoInfo].transitive_imports]
+  includes = [f for dep in ctx.attr.srcs for f in dep[ProtoInfo].transitive_imports.to_list()]
```
laurentlb: Ideally, you could avoid flattening the depset (https://docs.bazel.build/versions/master/skylark/performance.html#use-depsets) to reduce memory usage.

1. The inputs argument of actions.run can accept a depset.
2. For arguments, it's not as easy, but ctx.actions.args() should work.

I'd suggest to do at least 1. But in any case, this fix is enough for me as it fixes the issue.

werkt: Challenge accepted. protobuf and grpc deps upgraded. Using the grpc-java version of java_grpc_library.bzl now, should have fewer issues with it falling out of date.

PR #238, opened 2019-03-21
Avoid reading more than remaining in bytestream

werkt:


ulfjack:
Is there a particular reason for this? IIUC, you're basically duplicating what read() already does.

werkt:
Yes, there's a point here. If a limit is passed that is less than the available bytes, we don't want to advance the stream past what that limit requested. On the other side of the minimization, we don't want to read with a length parameter longer than the buffer

PR #248, opened 2019-04-23
Don't test on ubuntu1804_java11.

philwo:
That platform will default to JDK 11 soon, because it's the default for that distribution.


ulfjack:
LGTM

PR #270, opened 2019-06-06
HTTP Proxy fixes

moroten:
When trying to use a [HTTP cache](https://github.com/buchgr/bazel-remote/) in combination with BuildFarm, I started to experiment by letting my the CAS configuration point to a grpc->http proxy. I found the proxy implementation in this bazel-buildfarm repository, added in #144, but soon discovered problems. The problems are a mixture of bugs and adaptions of behavior to get Bazel to work.

After all, I abandoned the trials but would like to contribute with my findings. Please feel free to pick and polish the changes as I do not intend to do so myself.


werkt:
Going to close this, It was a great resource to land improvements to the http proxy, thanks again!

PR #272, opened 2019-07-09
Update the Java gRPC rules to keep them from breaking.

lberki:
They relied on accidental behavior in Bazel which is about to change.


lberki:
The commit in question is here:

https://github.com/bazelbuild/bazel/commit/2203a9ff4894d7c1b699c25d5be85b866954b929

Apologies for the no-op description (I originally conceived it for testing what would break with it). It removes the ability to reference `.proto` files with their repository-relative path even if they use `strip_import_prefix` or `import_prefix`.

PR #275, opened 2019-07-10
Update the patch to make gRPC Java work.

lberki:
A change to the fix was requested by the gRPC folks, so let's keep it tidy and replicate the change here.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,9 +28,9 @@ http_archive(
     patches = [
         "//third_party/io_grpc_grpc_java:054def3c63.patch",
         "//third_party/io_grpc_grpc_java:0959a846c8.patch",
-        "//third_party/io_grpc_grpc_java:1111913510.patch",
         "//third_party/io_grpc_grpc_java:952a767b9c.patch",
         "//third_party/io_grpc_grpc_java:c63752789e.patch",
+        "//third_party/io_grpc_grpc_java:1111913510.patch",
     ],
     sha256 = "f5d0bdebc2a50d0e28f0d228d6c35081d3e973e6159f2695aa5c8c7f93d1e4d6",
     strip_prefix = "grpc-java-1.19.0",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -28,9 +28,9 @@ http_archive(
     patches = [
         "//third_party/io_grpc_grpc_java:054def3c63.patch",
         "//third_party/io_grpc_grpc_java:0959a846c8.patch",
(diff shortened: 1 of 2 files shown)


lberki:
The pull request I'm merging here is https://github.com/grpc/grpc-java/pull/5967 .

werkt:
Just for clarification here, neither this nor its predecessor are intended to fix the "warning: directory does not exist" message that you've referenced, correct? I agree with your comment on that issue that it's pretty annoying, and saw that it still happens in the build logs from this.

WORKSPACE
```diff
@@ -28,9 +28,9 @@ http_archive(
     patches = [
         "//third_party/io_grpc_grpc_java:054def3c63.patch",
         "//third_party/io_grpc_grpc_java:0959a846c8.patch",
+        "//third_party/io_grpc_grpc_java:1111913510.patch",
```
meteorcloudy: Why changing the order of the patch file? The modification to the patch file made it compatible with `--incompatible_use_native_patch`, but changing the order of this patch file broke applying c63752789e.patch.

meteorcloudy: Why changing the order of this patch file? The modification to the content made it compatible with `--incompatible_use_native_patch`, but changing the order of patch files will break applying c63752789e.patch with the following error message:
```
com.google.devtools.build.lib.syntax.EvalException: Error applying patch C:/tools/msys64/home/pcloudy/workspace/bazel-buildfarm/third_party/io_grpc_grpc_java/c63752789e.patch: Incorrect Chunk: the chunk content doesn't match the target
**Original Position**: 1

**Original Content**:
# "repository" here is for Bazel builds that span multiple WORKSPACES.
def _path_ignoring_repository(f):
    if len(f.owner.workspace_root) == 0:

**Revised Content**:
load("@bazel_tools//tools/jdk:toolchain_utils.bzl", "find_java_runtime_toolchain", "find_java_toolchain")

# "repository" here is for Bazel builds that span multiple WORKSPACES.
def _path_ignoring_repository(f):
    if len(f.owner.workspace_root) == 0:
```

meteorcloudy:
Ping? What's the progress of this change? The patch files are still broken with `--incompatible_use_native_patch`.

werkt:
> Ping? What's the progress of this change? The patch files are still broken with `--incompatible_use_native_patch`.

@lberki can you take a look at the comments here? Fine to approve if you two can come to an agreement.

meteorcloudy:
Thanks! I can now confirm this works with `--incompatible_use_native_patch`! If everything else looks good, it's ready to merge now.

lberki:
k

PR #276, opened 2019-07-10
Fix patch file for --incompatible_use_native_patch

meteorcloudy:
Bazel will enable native patch implementation in repository rules, which is more strict on the correctness of the patch file content. Because fac9edbce1.patch is not generated based on grpc-java-1.19.0, the patch content is actually incorrect.

https://github.com/grpc/grpc-java/blob/42facbd535c9d373bd75153e5b980f26c2b5fb40/java_grpc_library.bzl#L7

Related https://github.com/bazelbuild/bazel/issues/8316
Test failure: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1085#86d2913d-646f-4f81-97ce-9fbabfc8d8a3

### DIFF
--- a/third_party/io_grpc_grpc_java/fac9edbce1.patch
+++ b/third_party/io_grpc_grpc_java/fac9edbce1.patch
@@ -12,7 +12,7 @@ diff --git a/java_grpc_library.bzl b/java_grpc_library.bzl
 index 7c1b1f2b8..2c00ee956 100644
 --- a/java_grpc_library.bzl
 +++ b/java_grpc_library.bzl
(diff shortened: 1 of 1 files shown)


werkt:
Can you add the gate (presumably just `--incompatible_use_native_patch` to .bazelrc) to ensure that we don't regress?

meteorcloudy:
@werkt That's a good idea, but this flag is not in release Bazel yet. I'm just fixing issues in advance.

werkt:
@lberki is already requesting a modification in #275 to that patch, please connect with him to ensure that it is now a compliant modification.

meteorcloudy:
Thanks for the notice, I commented in the other PR.

PR #289, opened 2019-10-07
Force TLS for workers

keith:
Depending on your deploy configuration connecting to the main service
over plain text may not be acceptable.


keith:
As I said this depends on your deploy config, so this might not be an acceptable change? If so is there some way we could make this configurable? In our infra this is required

PR #297, opened 2019-11-20
Add windows ci

werkt:


philwo:
Hey @werkt,

sorry that I never followed up on this. Can we try to rebase this to master and see if the error still occurs? If yes, let's see that we can find out what causes that error. I can try to repro on a Windows VM.

werkt:
@philwo looks like we've regressed even further, some substitution is now failing

werkt:
Ah, excellent. It looks like we've gotten to testing, where we're doing some jffi stuff to get posix native behavior (so further than we were stuck before).

I'll see about getting those tests to pass on windows, not sure if the whole hard link thing is going to work there, but you never know.

werkt:
@philwo heh, looks like the windows filesystem module in bazel has a problem with me marking my directories under tmp as read only: https://buildkite.com/bazel/buildfarm-male-farmer/builds/1001#86514436-e6f6-44d5-a59c-a83452a68672

PR #352, opened 2020-05-01
add support for dynamic conversion of slot range to hashtag

luxe:
Add a method that dynamically calculates a hashtag for a given slot range.  
We can consider replacing the static lookup table with this new functionality after testing it in a dev environment.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisSlotToHash.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisSlotToHash.java
@@ -20,6 +20,7 @@ import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import redis.clients.jedis.util.JedisClusterCRC16;
 
 ///
 /// @class   RedisSlotToHash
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/common/redis/RedisSlotToHash.java
```diff
@@ -46,7 +47,40 @@
   /// @note    Suggested return identifier: hashtag.
   ///
   public static String correlate(long slotNumber) {
+    Preconditions.checkState(slotNumber >= 0);
```
werkt, line 50: consolidate this range check into (slotNumber >= 0 && slotNumber < HASHSLOTS).

src/main/java/build/buildfarm/common/redis/RedisSlotToHash.java
```diff
@@ -46,7 +47,40 @@
...
+  /// @brief   Dynamically convert a range of slot number into string that
+  ///          hashes to a slot within the range.
+  /// @details Dynamically generates hashtags and tests them for valid slot
+  ///          number. Slower than static lookup, but less code.
+  /// @param   start The starting slot range number to find a hashable string for.
+  /// @param   end   The ending slot range number to find a hashable string for.
+  /// @return  The string value to be used in a key's hashtag.
+  /// @note    Suggested return identifier: hashtag.
+  ///
+  public static String dynamicCorrelate(long start, long end) {
```
werkt, line 64: Recommend `correlateRange` for name here

PR #353, opened 2020-05-01
exit worker with error message when publicName is empty

luxe:
The worker should fail to start if it cannot parse the URL that it is being told to put in as its public name.

### DIFF
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -53,7 +53,6 @@ import build.buildfarm.worker.Pipeline;
 import build.buildfarm.worker.PipelineStage;
 import build.buildfarm.worker.PutOperationStage;
 import build.buildfarm.worker.ReportResultStage;
-import build.buildfarm.worker.WorkerContext;
 import com.google.common.base.Strings;
 import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableList;
@@ -551,8 +550,7 @@ public class Worker extends LoggingMain {
     }
     if (!Strings.isNullOrEmpty(options.publicName)) {
       builder.setPublicName(options.publicName);
-    }
-    else{
+    } else {
       logger.log(SEVERE, "worker's public name should not be empty");
       System.exit(0);
     }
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -551,16 +550,18 @@ private static ShardWorkerConfig toShardWorkerConfig(Readable input, WorkerOptio
     }
     if (!Strings.isNullOrEmpty(options.publicName)) {
       builder.setPublicName(options.publicName);
+    } else {
+      logger.log(SEVERE, "worker's public name should not be empty");
+      System.exit(0);
```
werkt, line 555: This should not exit with a good (0) status. Recommend that you throw a runtime exception here with the failure instead. An outermost exception handler around this call should be responsible for logging.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -543,54 +542,60 @@ protected void onShutdown() throws InterruptedException {
...
+                session,
+                toShardWorkerConfig(
+                    new InputStreamReader(configInputStream),
+                    parser.getOptions(WorkerOptions.class)));
+      }
+      worker.start();
+      worker.blockUntilShutdown();
+      System.exit(0); // bullet to the head in case anything is stuck
+    } catch (Exception e) {
+      Worker.logger.log(SEVERE, e.toString());
```
werkt, line 598: I would prefer not to fully indent a method with a try block - prefer a separate method call in this case, and there are some other issues:
There's no need to user the `Worker.` here, you're in a static method.
The logging message should not be the string rendering of the exception. `logger.log(SEVERE, "exception caught", e)` should suffice here
I'd recommend `System.exit(1)`, or to rethrow the exception (probably the former, given that you're logging explicitly).

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -543,27 +542,39 @@ protected void onShutdown() throws InterruptedException {
...
   private static void printUsage(OptionsParser parser) {
     logger.log(INFO, "Usage: CONFIG_PATH");
-    logger.log(INFO,
-        parser.describeOptions(
-            Collections.emptyMap(), OptionsParser.HelpVerbosity.LONG));
+    logger.log(
+        INFO, parser.describeOptions(Collections.emptyMap(), OptionsParser.HelpVerbosity.LONG));
   }
 
   public static void main(String[] args) throws Exception {
```
werkt, line 566: One last thing now - main no longer throws exceptions

PR #358, opened 2020-05-04
Add operation metadata metrics publishing

80degreeswest:


src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -116,6 +120,17 @@ private static BuildFarmServerConfig toBuildFarmServerConfig(
...
   }
 
+  private static MetricsPublisher getMetricsPublisher(MetricsConfig metricsConfig) {
+    switch (metricsConfig.getMetricsDestination()) {
+      default:
+        return new LogMetricsPublisher(metricsConfig);
+      case "aws":
+        return new AwsMetricsPublisher(metricsConfig);
+      case "gcp":
+        throw new UnsupportedOperationException("Not Implemented.");
```
werkt, line 130: Let the implementation throw on bad usage, you've provided the class already

src/main/java/build/buildfarm/BUILD
```diff
@@ -74,6 +74,77 @@ java_library(
...
+    name = "common-metrics",
+    srcs = glob([
+        "common/metrics/AbstractMetricsPublisher.java",
+        "common/metrics/MetricsPublisher.java",
+    ]),
+    deps = [
+        "@maven//:com_google_guava_guava",
+        "@maven//:com_google_protobuf_protobuf_java",
+        "@maven//:com_google_protobuf_protobuf_java_util",
+        "//src/main/protobuf:build_buildfarm_v1test_buildfarm_java_proto",
```
werkt, line 87: deps ordering here and below.

src/main/java/build/buildfarm/common/metrics/aws/AwsMetricsPublisher.java
```diff
@@ -0,0 +1,105 @@
...
+    } catch (Exception e) {
+      logger.log(Level.WARNING, String.format("Could not publish request metadata to SNS for %s.", operation.getName()), e);
+    }
+  }
+
+  private AmazonSNSAsync initSnsClient() {
+    logger.log(Level.INFO, "Initializing SNS Client.");
+    return
+      AmazonSNSAsyncClientBuilder.standard()
+        .withRegion(Regions.US_EAST_1)
```
werkt, line 83: This needs to be config'd

src/main/java/build/buildfarm/BUILD
80degreeswest: We should look at the rest of the build files as well as the same ordering exists in other places.

src/main/java/build/buildfarm/common/metrics/gcp/GcpMetricsPublisher.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class GcpMetricsPublisher extends AbstractMetricsPublisher {
+  private static final Logger logger = Logger.getLogger(GcpMetricsPublisher.class.getName());
+
+  public GcpMetricsPublisher(MetricsConfig metricsConfig) {
+    super(metricsConfig.getClusterId());
+  }
+
+  public GcpMetricsPublisher() {
+    super();
```
werkt, line 29: this super call shouldn't be necessary

src/main/java/build/buildfarm/common/metrics/log/LogMetricsPublisher.java
```diff
@@ -0,0 +1,55 @@
...
+    super(metricsConfig.getClusterId());
+    if (!metricsConfig.getLogMetricsConfig().getLogLevel().isEmpty()) {
+      logLevel = Level.parse(metricsConfig.getLogMetricsConfig().getLogLevel());
+    } else {
+      logLevel = Level.FINEST;
+    }
+  }
+
+  public LogMetricsPublisher() {
+    super();
```
werkt, line 39: Shouldn't be necessary

PR #361, opened 2020-05-05
Pass platform execution properties for queue selection

luxe:
**OperationQueue API change**
This modifies the  `push` / `dequeue` methods of the `OperationQueue` so that they now take in a list of "platform execution properties".  These properties are used internally by the OperationQueue to determine the eligible provisioned queue to push/dequeue from (think "cpu provisioned queue" / "gpu provisioned queue").  

**Backplane API change** + callers
These OperationQueue methods are called from within the backplane, which in some cases, needed its own APIs extended to pass in execution properties from various contexts.  The worker context will now give platform properties when performing a dequeue (these platform properties are already a part of the worker's configuration).

**Updated grpc**
The OperationQueue can now divide operations by their platform properties.  This is reflected in the grpc status call and proto message.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
+++ b/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
@@ -108,7 +108,7 @@ public class OperationQueue {
   /// @note    Suggested return identifier: name.
   ///
   public String getDequeueName() {
-    return queues.get(0).queue().getDequeueName();
+    return "operation_dequeue";
   }
   ///
   /// @brief   Get dequeue name.
--- a/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
+++ b/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
@@ -14,9 +14,13 @@
 
 package build.buildfarm.instance.shard;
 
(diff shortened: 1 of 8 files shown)


src/main/java/build/buildfarm/instance/shard/OperationQueue.java
```diff
@@ -100,41 +106,118 @@ public long size(JedisCluster jedis) {
...
+  /// @details We use the platform execution properties of a queue entry to
+  ///          determine the appropriate queue. If there no eligible queues, an
+  ///          exception is thrown.
+  /// @param   provisions Provisions to check that requirements are met.
+  /// @return  The chosen queue.
+  /// @note    Suggested return identifier: queue.
+  ///
+  private BalancedRedisQueue chooseEligibleQueue(List<Platform.Property> provisions) {
+    for (ProvisionedRedisQueue pQueue : queues) {
+      if (pQueue.isEligible(toMultimap(provisions))) {
```
werkt, line 201: queue or provisionedQueue here

luxe:
By deleting my forked repository, I seemed to have orphaned the pull-request and can no longer update it.  My mistake.  I'll land these changes and make improvements in a follow-up diff.

PR #362, opened 2020-05-07
Add CAS indexer for worker re-indexing

luxe:
### Summary:
Over time, buildfarm workers become inaccessible due to termination via scaling or unexpected errors.  The redis CAS keys need their references to these workers re-indexed in order to properly reflect the currently active pool of workers.  This is primarily an optimization, as stale references to workers can lead to performance issues.  We integrate an indexer to run periodically as part of buildfarm's normal operation.

The indexer runs on the server, and a global lock is used within redis.

### Note:
This functionality was ported from an internal python script to Java.


luxe:
going to add a lock so not every worker performs the re-index at the same time

werkt:
I'd also suggest that workers probably not do this indexing in general - servers are better suited to it.

luxe:
This was an old diff where I lost the working branch.  It will be easier to make a new pull request.

PR #364, opened 2020-05-07
Fix instruction for external repository usage

dennisgranath:
While trying to follow the instruction for setting up buildfarm as external repository I could not make it work without doing this change to the repository name.


dennisgranath:
@googlebot I signed it!

werkt:
Thanks for pointing this out, and the contribution, I noticed that there were a stack of problems beyond just this one related to the name misalignment, and I've fixed them up with #365.

PR #366, opened 2020-05-08
Add health-check-probe for use with k8s

dennisgranath:
Adds https://github.com/grpc-ecosystem/grpc-health-probe to the container image so we can run grpc healtcheck in k8s.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -1,4 +1,4 @@
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_jar")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_jar", "http_file")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
 RULES_JVM_EXTERNAL_TAG = "3.0"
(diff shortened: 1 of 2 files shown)


werkt:
So we've already added health check natively to the buildfarm server: https://github.com/bazelbuild/bazel-buildfarm/blob/master/src/main/java/build/buildfarm/server/BuildFarmServer.java#L83, and this seems like something that you should be using composition of docker targets to do.

Does this container addition provide something different than the service internally?

dennisgranath:
I was planning on running the server in k8s and then we could add a readiness-probe which would run this binary to do the health check. But of course we could add this at our end. My reasoning was that adding it here might make it easier provide docs on k8s deployment at a later stage? Btw. how do you deploy buildfarm server?

dennisgranath:
We normally do grpc health checks like this in our k8s deployments:
        readinessProbe:
            exec:
              command: ["/grpc_health_probe", "-addr=:8990"]

werkt:
> I was planning on running the server in k8s and then we could add a readiness-probe which would run this binary to do the health check. But of course we could add this at our end. My reasoning was that adding it here might make it easier provide docs on k8s deployment at a later stage? Btw. how do you deploy buildfarm server?

Yeah, buildfarm already has this support - if you add the port of the buildfarm service to the exec check in k8s, the health check will verify not just that the container is up, but that buildfarm's grpc server is.

Regarding deploy, we also use k8s, just haven't explored these exec health checks (have others in place).

dennisgranath:
Ok! thanks for the info and for taking your time. I guess we can close this!

werkt:
No problem. If the health check service needs to be augmented to provide specific details as to the subsystems within buildfarm, happy to accept any PRs there.

PR #375, opened 2020-05-13
Add buildfarm run script to examples

80degreeswest:
Add a bf-run script to stand up a full operational buildfarm cluster locally for testing.
Usage example from the root directory:
examples/bf-run start
examples/bf-run stop

### DIFF
--- a/examples/bf-run
+++ b/examples/bf-run
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+set -e
+
+# Start Buildfarm Cluster
+start_buildfarm () {
+  # Run Redis Container
+  docker run -d --name buildfarm-redis -p 6379:6379 redis:5.0.9
+
+  # Run Buildfarm Server Container
+  docker run -d --name buildfarm-server -v $(pwd)/examples:/var/lib/buildfarm-server -p 8980:8980 --network host \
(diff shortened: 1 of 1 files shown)


examples/bf-run
```diff
@@ -0,0 +1,43 @@
...
+# Start Buildfarm Cluster
+start_buildfarm () {
+  # Run Redis Container
+  docker run -d --name buildfarm-redis -p 6379:6379 redis:5.0.9
+
+  # Run Buildfarm Server Container
+  docker run -d --name buildfarm-server -v $(pwd)/examples:/var/lib/buildfarm-server -p 8980:8980 --network host \
+  80dw/buildfarm-server:latest /var/lib/buildfarm-server/shard-server.config.example -p 8980
+
+  # Run Buildfarm Shard Worker Container
```
SrodriguezO, line 14: Out of curiosity, how come the buildfarm worker must run in a privileged container? And why does it need `host` networking?

80degreeswest: Privileged container is required if you are utilizing exec_properties. Otherwise it can be left off. Host networking is used to be able to reference locally running redis by localhost as in the example configs. If you specify a redis instance as an actual IP address this can also be left off. The idea here was to stand up a fully functioning Buildfarm cluster with the least amount of effort possible.

SrodriguezO: Awesome. Thanks for explaining! :)

PR #383, opened 2020-05-19
enforce formatting in CI

luxe:
This adds source code formatting to the CI.  Any formatting issues will surfaced in buildkite as followed:  
The log also contains the individual files that were not formatted properly.
buildkite_error


werkt:
Will this indicate a failed check from the github side?

luxe:
yeah, changes to presubmit.yml caused this pull-request to fail.  The commits show that it only began passing after formatting.

PR #392, opened 2020-05-27
set write permissions on CASFileCache data

luxe:
This isolates all of the calls in which we write files to the cache.  We can decide to disable write permissions on particular files here.  If we disable write permissions on the directories in the cache, we won't be able to add any more files to them without proactively enabling/disabling write permissions every time.  This is something we could consider doing if its important to protect the cache.  

Hoping this diff elicits further discussion on the issue.  I noticed the cache did not need modified for deleting invalid entries on startup.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CASFileCache.java
+++ b/src/main/java/build/buildfarm/worker/CASFileCache.java
@@ -1508,7 +1508,10 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     decrementReferencesSynchronized(inputFiles, inputDirectories);
   }
 
-  // The different contexts in which cache files are written to disk
+  /**
+   * The different contexts in which cache files are written to disk. Files are made non-writable in
+   * order to prevent operations from deleting them.
+   */
   private void putDirectoryHardLink(Path newHardLink, Path path) throws IOException {
     Files.createLink(newHardLink, path);
     new File(newHardLink.toString()).setWritable(false, false);
@@ -1516,12 +1519,15 @@ public abstract class CASFileCache implements ContentAddressableStorage {
 
   private void putDirectoryFile(Path path, FileNode fileNode) throws IOException {
     Files.createFile(path);
-    setPermissions(path, fileNode.getIsExecutable());
     new File(path.toString()).setWritable(false, false);
+    setPermissions(path, fileNode.getIsExecutable());
   }
 
   private void putDirectory(Path path) throws IOException {
-    Files.createDirectory(path);
+    if (!Files.exists(path)) {
+      File dir = new File(path.toString());
+      dir.mkdir();
+    }
   }
 
   private void putRootHardLink(Path newHardLink, Path path) throws IOException {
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -1508,6 +1508,30 @@ public synchronized void decrementReferences(
     decrementReferencesSynchronized(inputFiles, inputDirectories);
   }
 
+  /**
+   * The different contexts in which cache files are written to disk. Files are made non-writable in
+   * order to prevent operations from deleting them.
+   */
+  private void putDirectoryHardLink(Path newHardLink, Path path) throws IOException {
+    Files.createLink(newHardLink, path);
+    new File(newHardLink.toString()).setWritable(false, false);
```
werkt, line 1517: There appears to be no difference here with putRootHardLink below. These names are a little confusing to boot

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -1508,6 +1508,30 @@ public synchronized void decrementReferences(
...
+  }
+
+  private void putDirectoryFile(Path path, FileNode fileNode) throws IOException {
+    Files.createFile(path);
```
werkt, line 1521: This is an empty file creation? I would name it as such, and only pass the boolean executable

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -1508,6 +1508,30 @@ public synchronized void decrementReferences(
...
+    new File(path.toString()).setWritable(false, false);
+    setPermissions(path, fileNode.getIsExecutable());
+  }
+
+  private void putDirectory(Path path) throws IOException {
+    Files.createDirectory(path);
```
werkt, line 1527: A single line function here with access to base method at least deserves a comment as to its interface utility.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -1508,6 +1508,30 @@ public synchronized void decrementReferences(
...
+  }
+
+  private void putRootHardLink(Path newHardLink, Path path) throws IOException {
+    Files.createLink(newHardLink, path);
+    new File(newHardLink.toString()).setWritable(false, false);
```
werkt, line 1532: The false, false, params here are confusing, use the keyword pattern to denote meaning.

src/main/java/build/buildfarm/worker/CASFileCache.java
```diff
@@ -2436,6 +2459,7 @@ private CancellableOutputStream putOrReferenceGuarded(
                     expiredFuture,
                     (expiredKey) -> {
                       try {
+                        new File(expiredKey.toString()).setWritable(true, true);
```
werkt, line 2462: Double check the behavior here, file delete (unlink) only requires perms on the dir, and java may not perform any consideration for a readonly file deletion.

luxe:
> Is directory permissioning coming in a separate PR, or is it supposed to be here too?

Wanted to throw this diff up to discuss that in more detail.

These functions were to isolate all the file creation so I could debug things a little easier.  We can definitely get rid of them when I'm more confident about how to set permissions in all situations.   I we disable write on directories then we can't write files to those directories later.  

Do you think we should proactively enable/disable write permissions on directories every time the Cache needs to write a file to a directory?

luxe:
back to draft.  I'll try toggling write permissions on directories

luxe:
Taking a better approach.

PR #393, opened 2020-05-27
Remove unparsable workers

80degreeswest:


luxe:
What does the workerStub do?  I can't tell if this is possible to get stuck in an infinite loop or not.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -601,11 +601,21 @@ private void findMissingBlobsOnWorker(
       Executor executor,
       SettableFuture<Iterable<Digest>> missingDigestsFuture,
       RequestMetadata requestMetadata) {
-    String worker = workers.removeFirst();
+    Instance workerStub = null;
+    String worker;
+    do {
+      worker = workers.removeFirst();
+      try {
+        workerStub = workerStub(worker);
```
werkt, line 609: the obvious name collision here is a problem, and in general, I think that we should be making the backplane responsible for the validity of these contents - it should never return an unparseable worker from its workers presentation.

luxe:
it sounds like we aren't adding the check here.  closing the diff since its very old.  please resubmit if needed

PR #394, opened 2020-05-27
ensure OperationQueue can be configured through proto

luxe:
Previously, the OperationQueue was configured directly inside the ShardBackplane.  
This diff extends the configuration to our proto configs.


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1425,6 +1425,7 @@ protected void validatePlatform(
                       "property '%s' value was not a valid integer: %s",
                       property.getName(), property.getValue()));
         }
+      } else if (property.getName().equals("gpu")) {
```
werkt, line 1428: This seems extremely narrow - observe the provisioned queues to see what should be ignored here, rather than hardcoding gpu

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -486,26 +488,56 @@ private void startFailsafeOperationThread() {
...
+    // This will ensure the expected behavior for the paradigm in which all work is put on the same
+    // queue.
+    if (config.getProvisionedQueues().getQueuesList().isEmpty()) {
+      ProvisionedRedisQueue defaultQueue =
+          new ProvisionedRedisQueue(
+              config.getQueuedOperationsListName(), hashtags, LinkedHashMultimap.create());
+      allprovisionedQueues.add(defaultQueue);
+    }
+
+    this.operationQueue = new OperationQueue(allprovisionedQueues);
```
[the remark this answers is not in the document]
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
luxe: A little better now.  The validation is checked by the operationQueue so it can be consistent with how it decides eligibility

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1425,6 +1439,7 @@ protected void validatePlatform(
                       "property '%s' value was not a valid integer: %s",
                       property.getName(), property.getValue()));
         }
+      } else if (validForOperationQueue) {
```
werkt, line 1442: add a comment here that we're skipping things and that they are valid, no action is correct here

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1425,6 +1440,13 @@ protected void validatePlatform(
                       "property '%s' value was not a valid integer: %s",
                       property.getName(), property.getValue()));
         }
+
```
werkt, line 1443: Random whitespace

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardInstance.java

PR #399, opened 2020-06-03
Improve error message when precondition failed

jacobmou:


src/main/java/build/buildfarm/common/Actions.java
```diff
@@ -36,13 +36,23 @@ public static String invalidActionMessage(Digest actionDigest) {
...
   }
 
+  public static String invalidActionVerboseMessage(
+      Digest actionDigest, PreconditionFailure failure) {
+    return format(
+        "Action %s is invalid: %s",
+        DigestUtil.toString(actionDigest),
+        failure.getViolationsList().stream()
+            .map(Violation::getDescription)
+            .reduce("", (message, description) -> message + description + " "));
```
werkt, line 46: The list of violations can get *very* long - we need to ensure that this is trimmed to something reasonable: 3 violations with an ellipses is my preference.

PR #409, opened 2020-06-08
Configurize Worker CAS Shard Membership

luxe:
In order to support GPU workers, we need the ability for a worker to avoid registering themselves in the backplane.  We also need to prevent the following backplane actions:
 - puts
 - expiration

When a gpu worker finishes an operation the output should not be cached locally.  
Instead it will send the output to another CAS member for storage.  
This is implemented via logic in `insertFile`.


src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -488,6 +499,16 @@ private void insertBlob(Digest digest, ByteString content) throws InterruptedExc
   }
 
   private void insertFile(Digest digest, Path file) throws IOException, InterruptedException {
+
+    if (keepInsertFiles) {
+      insertFileToLocalStorage(digest, file);
+    } else {
+      insertFileToCasMember(digest, file);
```
werkt, line 506: Generalize these to consume a single Write

werkt: That will also handle the only other comment below, which will ensure that try-with-resources is used on the FeedbackOutputStream

examples/gpu/shard-cpu-worker.config.example
```diff
@@ -0,0 +1,241 @@
+# the listening port of the shard-worker grpc server
```
werkt, line 1: Drop this file and others in gpu, include comments in shard-worker example at base

examples/shard-worker.config.example
```diff
@@ -5,6 +5,11 @@ port: 8981
 # by all shard cluster servers
 public_name: "localhost:8981"
 
+# This determines whether the shard instance will participate in the backplane
+# and cache it's operation outputs in its own cache file CAS.
+# When not caching locally, the worker will send its output to another worker's CAS.
```
werkt, line 10: prefer 'whether the worker is a shard of the cas. When not a shard, it will upload into the Sharded CAS'

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -259,7 +259,7 @@ public void write(int b) throws IOException {
   @Override
   public synchronized boolean isReady() {
     if (writeObserver == null) {
-      return false;
+      return true;
     }
```
werkt, line 263: This should no longer be true, and it should be an error to call this before getOutput.

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -291,10 +291,11 @@ public boolean isComplete() {
 
   @Override
   public FeedbackOutputStream getOutput(
-      long deadlineAfter, TimeUnit deadlineAfterUnits, Runnable onReadyHandler) {
+      long deadlineAfter, TimeUnit deadlineAfterUnits, Runnable onReadyHandler) throws IOException {
```
werkt, line 294: Add a guard to prevent this from being called more than once

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -600,7 +600,7 @@ public Write getBlobWrite(Digest digest, UUID uuid, RequestMetadata requestMetad
     String resourceName =
         ByteStreamUploader.uploadResourceName(
             getName(), uuid, HashCode.fromString(digest.getHash()), digest.getSizeBytes());
-    return getWrite(resourceName, digest.getSizeBytes(), /* autoflush=*/ false, requestMetadata);
+    return getWrite(resourceName, digest.getSizeBytes(), /* autoflush=*/ true, requestMetadata);
```
werkt, line 603: This should no longer require autoflush

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -90,18 +96,22 @@
...
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 class ShardWorkerContext implements WorkerContext {
   private static final Logger logger = Logger.getLogger(ShardWorkerContext.class.getName());
 
+  private final com.google.common.cache.LoadingCache<String, Instance> workerStubs;
```
werkt, line 114: why the fqcn here? See the below comment about how I don't think this class should be made aware of workerStubs at all.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -122,6 +132,7 @@
   private final boolean limitExecution;
   private final boolean limitGlobalExecution;
   private final boolean onlyMulticoreTests;
+  private final boolean keepInsertFiles;
```
werkt, line 135: I would prefer if this class were given a Write-supplier that did the 'right' thing - there doesn't need to be any functionality split for this class to see.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -182,12 +194,14 @@
...
     this.limitGlobalExecution = limitGlobalExecution;
     this.onlyMulticoreTests = onlyMulticoreTests;
+    this.keepInsertFiles = keepInsertFiles;
     Preconditions.checkState(
         !limitGlobalExecution || limitExecution,
         "limit_global_execution is meaningless without limit_execution");
     Preconditions.checkState(
         !onlyMulticoreTests || limitExecution,
         "only_multicore_tests is meaningless without limit_execution");
+    workerStubs = WorkerStubs.create(getDigestUtil());
```
werkt, line 204: You don't want to create a separate workerStubs cache here - the worker already has one for doing remote cas downloads.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -726,10 +883,10 @@ public void uploadOutputs(
     for (String outputFile : outputFiles) {
       uploadOutputFile(resultBuilder, outputFile, actionRoot, preconditionFailure);
     }
-
     for (String outputDir : outputDirs) {
       uploadOutputDirectory(resultBuilder, outputDir, actionRoot, preconditionFailure);
     }
+
```
werkt, line 889: odd whitespace moves here...

examples/shard-worker.config.example
```diff
@@ -116,6 +120,22 @@ redis_shard_backplane_config: {
...
+  #      name: "queue1"
+  #      platform: {
+  #        properties: [{name: "key" value: "value"}]
+  #      }
+  #    },
+  #    {
+  #        name: "queue2"
+  #    }
+  #  ]
+  #}
```
werkt, line 138: Place this in line with the example I've provided at the end of the shard-worker with a viable provisioned_queues

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -259,7 +259,7 @@ public void write(int b) throws IOException {
   @Override
   public synchronized boolean isReady() {
     if (writeObserver == null) {
-      return false;
+      throw new RuntimeException("isReady should not be called before getOutput");
```
werkt, line 262: Prefer `checkNotNull` with the message here - the ISE that it throws is better than RE.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -477,30 +476,19 @@ public Duration getMaximumActionTimeout() {
   }
 
   private void insertBlob(Digest digest, ByteString content) throws InterruptedException {
+
```
werkt, line 479: whitespace

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -477,30 +476,19 @@ public Duration getMaximumActionTimeout() {
...
     if (digest.getSizeBytes() > 0) {
-      Blob blob = new Blob(content, digest);
-      execFileSystem.getStorage().put(blob);
+      try {
+        writer.get().insertBlob(digest, content);
+      } catch (IOException e) {
+      }
```
werkt, line 484: Don't swallow this exception

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -477,30 +476,19 @@ public Duration getMaximumActionTimeout() {
...
-      if (!write.isComplete()) {
-        write.reset(); // we will not attempt retry with current behavior, abandon progress
-        if (e.getCause() != null) {
-          Throwables.propagateIfInstanceOf(e.getCause(), InterruptedException.class);
-        }
-        throw e;
-      }
-    }
+
+    System.out.println(file.toString());
```
werkt, line 490: Debug output

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -203,6 +241,19 @@ public Worker(String session, ServerBuilder<?> serverBuilder, ShardWorkerConfig
...
 
+    // Create the appropriate writer for the context
+    CasWriter writer;
+    if (config.getOmitFromCas()) {
+      writer = new RemoteCasWriter();
+    } else {
+      writer = new LocalCasWriter();
+    }
+
+    Supplier<CasWriter> suppliedWriter =
```
werkt, line 252: `Suppliers.ofInstance(writer)`

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -439,15 +676,23 @@ public void stop() throws InterruptedException {
...
       throw Status.fromThrowable(e).asRuntimeException();
     }
   }
 
   private void onStorageExpire(Iterable<Digest> digests) {
     try {
-      backplane.removeBlobsLocation(digests, config.getPublicName());
+
+      // if the worker is a CAS member, it can send/modify blobs in the backplane.
+      if (!config.getOmitFromCas()) {
```
werkt, line 693: This should be a check for boolean state (and we can flip the logic internally to be positive: i.e. `isCasShard`) rather than a dereference of the config. Generally it makes things more testable.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -100,6 +116,28 @@
   private final ShardBackplane backplane;
   private final LoadingCache<String, Instance> workerStubs;
 
+  class LocalCasWriter implements CasWriter {
```
werkt, line 119: This feels awkward - with the only caller of insertFile/BlobLocally and insertFile/BlobToCasMember being these classes, they should retain those methods (and their helpers), rather than the outer class.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -100,6 +117,118 @@
...
+      } catch (IOException e) {
+        if (!write.isComplete()) {
+          write.reset(); // we will not attempt retry with current behavior, abandon progress
+          if (e.getCause() != null) {
+            Throwables.propagateIfInstanceOf(e.getCause(), InterruptedException.class);
+          }
+          throw e;
+        }
+      }
+    }
```
werkt, line 143: The replication is strong with this one. We already have an input stream supplier iirc - reduce the duplicity between this and the blob insert with that

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -258,6 +397,101 @@ public Worker(String session, ServerBuilder<?> serverBuilder, ShardWorkerConfig
...
+      }
+      return worker;
+    }
+  }
+
+  private Instance workerStub(String worker) {
+    try {
+      return workerStubs.get(worker);
+    } catch (ExecutionException e) {
+      logger.log(Level.SEVERE, "error getting worker stub for " + worker, e);
```
werkt, line 490: nit: `e` is mostly useless here, prefer `e.getCause()`

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -439,15 +673,23 @@ public void stop() throws InterruptedException {
...
     try {
-      backplane.removeBlobsLocation(digests, config.getPublicName());
+
+      // if the worker is a CAS member, it can send/modify blobs in the backplane.
+      if (isCasShard) {
+        backplane.removeBlobsLocation(digests, config.getPublicName());
+      }
     } catch (IOException e) {
       throw Status.fromThrowable(e).asRuntimeException();
     }
```
werkt, line 695: nit: if around try. even better for either not registering this callback for !isCasShard, or another method which has the exception translation.

src/test/java/build/buildfarm/common/grpc/StubWriteOutputStreamTest.java
```diff
@@ -158,4 +159,33 @@ public void resetIsRespectedOnSubsequentWrite() throws IOException {
...
+
+    boolean callbackTimedOut = false;
+    try (OutputStream out =
+        write.getOutput(
+            1,
+            MICROSECONDS,
+            () -> {
+              try {
+                Thread.sleep(1000);
+              } catch (Exception e) {
```
werkt, line 183: nit: InterruptedException

src/test/java/build/buildfarm/common/grpc/StubWriteOutputStreamTest.java
```diff
@@ -158,4 +159,33 @@ public void resetIsRespectedOnSubsequentWrite() throws IOException {
...
+              }
+            })) {
+    } catch (Exception e) {
```
werkt, line 186: nit: more exception type information, both on timeout detection and the catch

PR #415, opened 2020-06-11
Add invocation blacklist to reject entire builds

werkt:
The blacklist must be maintained independently for now - seek to add control for this in a separate service control plane.

### DIFF
--- a/examples/shard-server.config.example
+++ b/examples/shard-server.config.example
@@ -87,6 +87,16 @@ instances {
       # than a catastrophic failure.
       action_blacklist_prefix: "ActionBlacklist"
 
+      # A redis key prefix for all blacklisted invocations, suffixed
+      # with the tool invocation id. Requests on behalf of an invocation
+      # which is blacklisted should be rejected where it is identified via
+      # its RequestMetadata
+      # To meet API standards, a request which matches this condition
+      # receives a transient UNAVAILABLE response, which, in bazel's
+      # case, can induce a fallback to non-remote recovery, rather
+      # than a catastrophic failure.
+      invocation_blacklist_prefix: "InvocationBlacklist"
+
       # The ttl maintained for action_blacklist entries.
       action_blacklist_expire: 3600 # 1 hour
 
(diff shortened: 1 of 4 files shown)


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -214,6 +214,7 @@ message RedisShardBackplaneConfig {
   int32 action_cache_expire = 4;
   string action_blacklist_prefix = 28;
   int32 action_blacklist_expire = 29;
+  string invocation_blacklist_prefix = 30;
```
luxe, line 217: do we want to use `repeated`?

examples/shard-server.config.example
```diff
@@ -87,6 +87,16 @@ instances {
...
 
+      # A redis key prefix for all blacklisted invocations, suffixed
+      # with the tool invocation id. Requests on behalf of an invocation
+      # which is blacklisted should be rejected where it is identified via
+      # its RequestMetadata
+      # To meet API standards, a request which matches this condition
+      # receives a transient UNAVAILABLE response, which, in bazel's
+      # case, can induce a fallback to non-remote recovery, rather
+      # than a catastrophic failure.
+      invocation_blacklist_prefix: "InvocationBlacklist"
```
luxe, line 98: because "unavailable" is the closest thing we have to "go away" response?  
In this context, I'm also guessing its also wrong to respond with "failed condition" or something similar.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
werkt: Negative, why would you want to specify multiple invocation blacklist prefixes?

examples/shard-server.config.example
werkt: This is the response that is compatible with --remote_local_fallback. This reminds me that we need to have a 'fallback-friendly' option and some more definitive responses here for non-fallback compatible buildfarms.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: oops. misread this as actual blacklisted invocations

examples/shard-server.config.example
luxe: thx. I need to read up on all the possible responses that are expected for the various remote apis

PR #417, opened 2020-06-15
Set configurable ownership privs for exec outputs

werkt:
Specify a user principal for ownership of directories created in the
service of exec roots - this is distinguished from ownership by
CASFileCache (determined by uid of the worker) - so that executions may
only write to directories manifested by output trees.
Refactors CASFileCache permission setting into utility library which
supports global execution of non-exec-owner files.
Generalized operations for Files and Directories now throw an error
without a matching attribute view.
BUILD file linted by buildifier

### DIFF
--- a/examples/shard-worker.config.example
+++ b/examples/shard-worker.config.example
@@ -212,3 +212,9 @@ redis_shard_backplane_config: {
   # When in doubt, leave this enabled.
   subscribe_to_backplane: true
 }
+
+# Create exec trees containing directories that are owned by
+# this user
+# The default (empty) value does not change the owner
+#exec_owner: "nobody"
+exec_owner: ""
--- a/src/main/java/build/buildfarm/BUILD
+++ b/src/main/java/build/buildfarm/BUILD
@@ -418,10 +418,11 @@ java_library(
         ":stub-instance",
         ":worker",
         ":worker-cgroup",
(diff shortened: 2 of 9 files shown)


werkt:
> How do we intend to configure this?

Not sure what 'we' you mean here. This is mechanism, which enables policy defined for application locally. The purpose is to be able to control this.

>     1. Leave `exec_owner` empty?  In which case, we will not set ownership on the directories (making this diff sort of a no op in terms of behavior)

That is correct, the default is a no op.

>     2. Do set an `exec_owner`.  Would that owner would be same across all of the workers?  is it root?

Setting this is execution policy. It doesn't matter what owner is selected on a given worker, as it cannot affect any other worker.

> Also, I'm trying to compare the file changes invoked between the CasFileCache and ExecFileSystem now. Files created in the CasFileCache will be write disabled,
> but when the ExecFileSystem gets setup with the hardlinks and everything, then all of those files get their ownership changed as well. Does that mean the same files are getting both owernship and write disabled now? Are ownership/write changes 1-to-1 across the affected files?

The ownership does not change for files, only directories. Only newly created directories exist under exec roots, and symlinks are not followed for the visitation.

Regarding the read-only changes, there is only refactorization from the previous scenario - only directories created as CFC entries (i.e. symlink *targets*) have their write bit removed. The set of DirectoryEntry directories and the exec dir directories have no intersections.

PR #422, opened 2020-06-17
Add CAS indexer for worker re-indexing

luxe:
### Summary:
Over time, buildfarm workers become inaccessible due to termination via scaling or unexpected errors.  The redis CAS keys need their references to these workers re-indexed in order to properly reflect the currently active pool of workers.  This is primarily an optimization, as stale references to workers can lead to performance issues.  We integrate an indexer to run periodically as part of buildfarm's normal operation.

The indexer runs on the server, and a global lock is used within redis.

### Note:
This functionality was ported from an internal python script to Java.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/ShardInstance.java
+++ b/src/main/java/build/buildfarm/instance/shard/ShardInstance.java
@@ -74,6 +74,7 @@ import build.buildfarm.instance.AbstractServerInstance;
 import build.buildfarm.instance.ExcessiveWriteSizeException;
 import build.buildfarm.instance.Instance;
 import build.buildfarm.v1test.ExecuteEntry;
+import build.buildfarm.v1test.IndexSettings;
 import build.buildfarm.v1test.OperationIteratorToken;
 import build.buildfarm.v1test.OperationsStatus;
 import build.buildfarm.v1test.ProfiledQueuedOperationMetadata;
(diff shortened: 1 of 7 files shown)


cristifalcas:
@werkt can we have this one merged? We want to test buildfarm in gke with HPA and this seems like a thing we will hit a lot.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -418,6 +426,26 @@ public void start() {
...
     }
+
+    if (config.getRunIndexer()) {
+      indexerService.scheduleAtFixedRate(
+          () -> {
+            try {
+
+              // ensure only one scheduler performs the re-indexing.
+              if (backplane.takeIndexerLock()) {
+                backplane.runIndexer();
```
werkt, line 437: There are two failure cases not considered here: the first is easy - once the lock is acquired, a finally should ensure that even if runIndexer() throws an exception, that the indexer lock is released.

The second is the abnormal termination of a process, which means that there needs to be an upper bound on the lock acquisition time. Ideally that should be tied to the lifetime of some redis connection, but in a cluster, that will be difficult. Preferentially there should be a) a required poll of the activity in the indexer to extend the lock so that the timeout can remain short, and b) some clear indication which node is performing an index currently, and queryability as to its progress

PR #432, opened 2020-06-25
fix FileDirectoriesIndex sqlite insert bug and add test

jacobmou:
Fix a bug in sqlite insert sql and simple test cases to validate its functionalities.

### DIFF
--- a/src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
+++ b/src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
@@ -1,4 +1,4 @@
-// Copyright 2018 The Bazel Authors. All rights reserved.
+// Copyright 2020 The Bazel Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
+++ b/src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
@@ -35,19 +35,16 @@ import org.junit.runners.JUnit4;
 public class FileDirectoriesIndexTest {
   private final DigestUtil DIGEST_UTIL = new DigestUtil(HashFunction.SHA256);
 
-  private final String jdbcIndexUrlBase = "jdbc:sqlite:memory:";
+  private final String jdbcIndexUrl= "jdbc:sqlite::memory:";
   private Path root;
   private FileDirectoriesIndex directoriesIndex;
-  private String system;
 
-  protected FileDirectoriesIndexTest(Path root, String system) {
+  protected FileDirectoriesIndexTest(Path root) {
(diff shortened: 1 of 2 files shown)


src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
```diff
@@ -0,0 +1,89 @@
...
+    for (String entry : entries) {
+      Set<Digest> digests = directoriesIndex.removeEntry(entry);
+      assertThat(digests.size() == 1 && digests.contains(directory)).isTrue();
+    }
+
+    // put again to test remove by directory
+    Files.write(path, coolBlob.toByteArray());
+    directoriesIndex.put(directory, entries);
+    directoriesIndex.remove(directory);
+    assertThat(Files.notExists(directoriesIndex.path(directory))).isTrue();
```
werkt, line 87: This doesn't test the other side of this, whether the index thinks that it actually removed the entry index membership

werkt: Given the previous bug, such a test should fail without this change.

jacobmou: There are something i'm missing here, because you're right, the test cannot catch the previous bug, which confused me.

The bug is, (directory, entry) get inserted as (entry, directory) into the database, which means when we do SELECT/DELETE on entry, we cannot find the rows, therefore no directory get deleted. 
If that's the case, the testcase on line 79-80 should already catch it, since the returned set come from a sql SELECT on entry. If directory is used as entry, nothing should be contained in the Set.

Apparently, I'm wrong, because the test pass either way. But why?

werkt: The directories files were removed, because they are correctly identified.

You need to assert that prior to the remove, the directories are mapped to their entries via `directoryEntries` (they will not in the case of the bug).

You need a separate test to perform `removeEntry`, and assert that the returned `Set<Digest>` contains the expected directories.

jacobmou: issue found and fixed!

src/test/java/build/buildfarm/cas/FileDirectoriesIndexTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    // insert
+    directoriesIndex.put(directory, entriesBuilder.build());
+    entries = directoriesIndex.directoryEntries(directory);
+    assertThat(entries).contains(digest.getHash());
+
+    // remove entry-wise
+    for (String entry : entries) {
+      Set<Digest> digests = directoriesIndex.removeEntry(entry);
+      assertThat(digests.size() == 1 && digests.contains(directory)).isTrue();
```
werkt, line 76: nit, write this as a containment assert:

`assertThat(digests.size()).containsExactly(directory);`

PR #443, opened 2020-06-30
Serialize bytestream proxy callbacks to a single thread in StubInstance.

AngusDavis:
This commit moves all interaction with the server response stream and
client stub onto a single thread. This allows us to have guarantees
that onNext and onReady will not execute in an interleaved fashion.
Critical sections could have been added, but blocking either the client
stub to the shard worker or server stream could have impacts on
performance of other streams being handled on those transports.

For #440

### DIFF
--- a/src/main/java/build/buildfarm/instance/stub/StubInstance.java
+++ b/src/main/java/build/buildfarm/instance/stub/StubInstance.java
@@ -113,6 +113,8 @@ import java.util.Iterator;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -132,6 +134,7 @@ public class StubInstance implements Instance {
   private final TimeUnit deadlineAfterUnits;
   private final Retrier retrier;
   private final @Nullable ListeningScheduledExecutorService retryService;
+  private final ExecutorService byteStreamExecutorService;
   private boolean isStopped = false;
   private final int maxBatchUpdateBlobsSize = 3 * 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -309,6 +313,11 @@ public void stop() throws InterruptedException {
     if (retryService != null && !shutdownAndAwaitTermination(retryService, 10, TimeUnit.SECONDS)) {
       logger.log(Level.SEVERE, format("Could not shut down retry service for %s", identifier));
     }
+    byteStreamExecutorService.shutdown();
```
werkt, line 316: Prefer the use of `shutdownAndAwaitTermination` as above

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -132,6 +134,7 @@
   private final TimeUnit deadlineAfterUnits;
   private final Retrier retrier;
   private final @Nullable ListeningScheduledExecutorService retryService;
+  private final ExecutorService byteStreamExecutorService;
```
werkt, line 137: Provide the member level initialization here of `newSingleThreadExecutor()`, rather than in the constructor, as it's not dependent upon parameter injection

AngusDavis: Done

src/main/java/build/buildfarm/instance/stub/StubInstance.java
AngusDavis: Done.

AngusDavis:
> Looks good, I wonder if there are other issues with the serverside implementation of BS::Read lying in wait

Thanks for merging! So far ByteStream seems to be working well for us with this change.

werkt:
No problem, thanks for the contribution, and I definitely saw the race as described by you. I could easily see this done with the critical section as well, but I dont think a thread per client is too bad for now - maybe with thousands of workers it could be problematic, but I dont know of anyone at that scale.

PR #445, opened 2020-07-03
Specify debug image base for instrumentation tools

werkt:


codesuki:
Breaks https://github.com/bazelbuild/bazel-buildfarm/blob/master/src/main/java/build/buildfarm/BUILD#L359

PR #459, opened 2020-07-09
enable test

luxe:


werkt:
This still a WIP?

luxe:
yeah, need to fix IOUtils on windows before enabling these tests

PR #464, opened 2020-07-10
use ConcurrentMap for directoryStorage

jacobmou:
Use ConcurrentMap for directory storage to reduce lock retention. By using the following data
```
Storage Count: 3872751
Directory Count: 242054
Current Size: 1718GB
```
this change can reduce 13-43% of time cost in computeDirectories(...).


luxe:
nice

PR #465, opened 2020-07-10
check write completion and show digest

luxe:
If an exception is thrown when transferring a file, check if the write has completed, and show the digest of the file.


src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -182,7 +182,7 @@ private void writeToCasMember(Digest digest, InputStream in)
       Instance casMember = workerStub(workerName);
       Write write = getCasMemberWrite(digest, workerName);
 
-      streamIntoWriteFuture(in, write, digest.getSizeBytes()).get();
+      streamIntoWriteFuture(in, write, digest, digest.getSizeBytes()).get();
```
werkt, line 185: nit, you don't need to pass size if you're already passing digest

PR #467, opened 2020-07-14
add unreferenced file entry to WorkerProfileService

jacobmou:
to gain a better understanding of the file entry pool.


luxe:
the more information the better.  I don't really understand what unreferenced entry count means though- can you explain?  
some entries are referenced, and others are not?  why is this the case.

jacobmou:
> the more information the better. I don't really understand what unreferenced entry count means though- can you explain?
> some entries are referenced, and others are not? why is this the case.

Entries represent files in CAS. Referenced Entries are Entries currently used by certain directories, and should not be removed/expired, and unreferenced entries are not used and can be candidates for Entry LRU evictions.

This is added to help us to understand the distribution of all entries. This might be useful for future entires lru optimization in the future.

werkt:
> Entries represent files in CAS. Referenced Entries are Entries currently used by certain directories, and should not be removed/expired, and unreferenced entries are not used and can be candidates for Entry LRU evictions.

Clarification on this: Referenced Entries are entries that are **currently in use by an Operation in the Pipeline**, not just (or specifically) entries in directories - Referenced Directories transitively refer to their constituent Entries, making a referenced directory id shorthand for a group of entries.

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2729,6 +2734,7 @@ void commit() throws IOException {
 
   @VisibleForTesting
   public static class Entry {
+    public static AtomicLong entryCount = new AtomicLong(0);
```
werkt, line 2737: This shouldn't be state maintained statically in the Entry list - all CASFileCaches will share this statistic if so. It needs to be a (final) instance member of the CFC, and maintained as such.

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -177,6 +178,7 @@ public void onRemoval(
 
   private transient long sizeInBytes = 0;
   private transient Entry header = new SentinelEntry();
+  private final AtomicLong unreferencedEntryCount = new AtomicLong(0);
```
werkt, line 181: Do you need an AtomicLong here? A volatile might suffice, as all list operations are already guarded by `synchronized(this)`

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -188,10 +190,14 @@ public synchronized long size() {
     return sizeInBytes;
   }
 
-  public long storageCount() {
+  public long totalEntryCount() {
```
werkt, line 193: `totalEntryCount` sounds like there's component groups to entries: `entryCount` should be fine

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -1442,6 +1448,9 @@ private void processRootFile(
             onPut.accept(fileEntryKey.getDigest());
             synchronized (CASFileCache.this) {
               e.decrementReference(header);
```
werkt, line 1450: Have `decrementReference` return the count and check it in one swoop here

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -1602,6 +1611,7 @@ private int decrementInputReferences(Iterable<String> inputFiles) {
       e.decrementReference(header);
       if (e.referenceCount == 0) {
         entriesDereferenced++;
+        unreferencedEntryCount.incrementAndGet();
```
werkt, line 1614: Same as above

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2454,6 +2469,9 @@ private synchronized boolean referenceIfExists(String key) throws IOException {
     }
 
     e.incrementReference();
```
werkt, line 2471: Likewise. I'm also leery here of the check for '1' - maybe return a boolean to indicate whether a referenced state change occurred

jacobmou:
@werkt ready for review again! Thanks!

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2803,11 +2800,14 @@ public void incrementReference() {
                   + ") and is being incremented");
         }
         unlink();
+        becomeReferenced = true;
       }
       referenceCount++;
+      return becomeReferenced;
```
werkt, line 2806: A little awkward: prefer `return referenceCount++ == 0;`

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2821,9 +2821,12 @@ public void decrementReference(Entry header) {
               + " to "
               + (referenceCount - 1));
       referenceCount--;
+      boolean becomeUnreferenced = false;
       if (referenceCount == 0) {
         addBefore(header);
+        becomeUnreferenced = true;
       }
+      return becomeUnreferenced;
```
werkt, line 2829: Prefer: `if (--referenceCount == 0) {` without the boolean. return true in the block, false otherwise. The early return is kind of weird, but you can use your judgement there.

PR #468, opened 2020-07-14
Explain eligibility on ProvisionedQueue

luxe:
In order to improve error reporting/debugging of queue selection in the future:
- add a method to the `ProvisionedRedisQueue`  that explains why the properties were eligible or not.
- add tests


src/main/java/build/buildfarm/common/redis/ProvisionedRedisQueue.java
```diff
@@ -127,4 +175,39 @@ public boolean isEligible(SetMultimap<String, String> properties) {
   public BalancedRedisQueue queue() {
     return queue;
   }
+  ///
+  /// @brief   Filter the provisions into separate sets by checking for the
+  ///          existence of wildcards.
+  /// @details This will organize the incoming provisions into separate sets.
+  /// @param   filterProvisions The filtered provisions of the queue.
+  ///
+  private void filterProvisionsByWildcard(SetMultimap<String, String> filterProvisions) {
```
werkt, line 184: make this static and return the filtered provisions to assign to a final requiredProvisions - definalizing this is a really bad idea from a computability point of view.

luxe:
ready for re-review

PR #490, opened 2020-08-03
Allow custom connection timeout and connection attempts for jedis cluster

80degreeswest:
Addresses https://github.com/bazelbuild/bazel-buildfarm/issues/484

The default connection timeout is 2000ms and the max number of attempts is 5. When network hiccups occur, the max number of connections gets exhausted and the container shuts down. This change will give us the ability to increase the thresholds so that we can better control connection to Redis.


80degreeswest:
> great!
> 
> in case its not obvious from CI, you just need to run `./.bazelci/format.sh` before landing

Ah yes, done.

PR #491, opened 2020-08-05
Coerce io into Interrupt on destroyExecDir

werkt:
Interpret ClosedByInterruptException at the CAS layer as an
InterruptedException when interacting with the DirectoriesIndex.
Avoid wrapping IOExceptions in RuntimeExceptions (SQLExceptions are to
be dealt with separately).

### DIFF
--- a/src/main/java/build/buildfarm/cas/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/CASFileCache.java
@@ -1586,13 +1586,20 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     return getFileName(digest, isExecutable);
   }
 
-  private void decrementReference(String inputFile) {
-    decrementReferences(ImmutableList.of(inputFile), ImmutableList.of());
+  private synchronized void decrementReference(String inputFile) throws IOException {
+    decrementReferencesSynchronized(ImmutableList.of(inputFile), ImmutableList.of());
   }
 
   public synchronized void decrementReferences(
-      Iterable<String> inputFiles, Iterable<Digest> inputDirectories) {
-    decrementReferencesSynchronized(inputFiles, inputDirectories);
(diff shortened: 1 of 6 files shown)


werkt:
Fixes #485 as well

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2140,11 +2148,15 @@ Digest getDigest() {
...
 
     ListenableFuture<Void> deindexFuture =
-        transform(
+        transformAsync(
             expireFuture,
             result -> {
-              directoriesIndex.remove(digest);
-              return null;
+              try {
+                directoriesIndex.remove(digest);
```
ericfelly, line 2155: what's the purpose of change the transform to async here?

src/main/java/build/buildfarm/cas/CASFileCache.java
```diff
@@ -2189,34 +2201,41 @@ Digest getDigest() {
...
                 disableAllWriteAccess(path);
-              } catch (IOException chmodException) {
-                logger.log(
-                    Level.SEVERE,
-                    "error while disabling write permissions on " + path,
-                    chmodException);
-                return immediateFailedFuture(chmodException);
+              } catch (IOException e) {
+                logger.log(Level.SEVERE, "error while disabling write permissions on " + path, e);
+                return immediateFailedFuture(e);
```
ericfelly, line 2214: same as above: if all results are "immediate" you should be able to reuse a regular non-async transform()

src/main/java/build/buildfarm/cas/CASFileCache.java
werkt: To redirect checked IOExceptions into the future. `transform` routines are runnable and may not throw them

ericfelly: Good point!

PR #494, opened 2020-08-19
Add admin grpc service

80degreeswest:
Add an admin grpc service to perform various buildfarm management tasks, such as terminate hosts, reboot instances, etc.

### DIFF
--- a/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
+++ b/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
@@ -1,0 +1,59 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.common.admin.aws;
+
+import build.buildfarm.common.admin.Admin;
+import com.amazonaws.services.ec2.AmazonEC2;
+import com.amazonaws.services.ec2.AmazonEC2ClientBuilder;
+import com.amazonaws.services.ec2.model.TerminateInstancesRequest;
+import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement;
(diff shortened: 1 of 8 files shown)


src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Map;
+import java.util.logging.Logger;
+
+public class AwsAdmin implements Admin {
+  private static final Logger logger = Logger.getLogger(Admin.class.getName());
+  private final AmazonEC2 ec2;
+  private final AWSSimpleSystemsManagement ssm;
+
+  public AwsAdmin() {
+    ec2 = AmazonEC2ClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
```
80degreeswest, line 37: TODO: Config

src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
```diff
@@ -0,0 +1,55 @@
...
+    ssm = AWSSimpleSystemsManagementClientBuilder.standard().withRegion(Regions.US_EAST_1).build();
```
80degreeswest, line 38: TODO: Config

src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
```diff
@@ -0,0 +1,55 @@
...
+    logger.info(String.format("Terminated host: %s", hostId));
+  }
+
+  @Override
+  public void stopContainer(String hostId, String containerName) {
+    String stopContainerCmd = "docker ps | grep " + containerName + " | awk '{print $1 }' | xargs -I {} docker stop {}";
+    Map<String, List<String>> parameters = new HashMap<String, List<String>>();
+    parameters.put("commands", Collections.singletonList(stopContainerCmd));
+    ssm.sendCommand(new SendCommandRequest().withDocumentName("AWS-RunShellScript").withInstanceIds(hostId).withParameters(parameters));
+    logger.info(String.format("Stopped container: %s on host: %s", containerName, hostId));
```
luxe, line 53: should we confirm that the command succeeded, and that the container actually stopped?  I guess buildfarm admin has always just done it this way too.

80degreeswest:
> I understand the scheduler getting the grpcAdmin endpoints. But how does the awsAdmin come into play? How will its methods get called?

AdminService determines what is called based on the config (deployment_environment)

src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
80degreeswest: Nope, we should just fire off the event and forget about it. We do send the status back so we can listen for errors if we want to on the client side. The client can handle refreshing of the status on the front end.

PR #496, opened 2020-08-21
Add getHosts grpc service

80degreeswest:
Add a getHosts grpc service which will return scheduler or worker hosts running in the cloud environment (this PR is for AWS implementation only).

### DIFF
--- a/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
+++ b/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
@@ -101,7 +101,7 @@ public class AwsAdmin implements Admin {
     }
     resultBuilder.addAllHosts(hosts);
     resultBuilder.setNumHosts(hosts.size());
-    logger.info(String.format("Got %d hosts for filter: %s", hosts.size(), filter));
+    logger.fine(String.format("Got %d hosts for filter: %s", hosts.size(), filter));
     return resultBuilder.build();
   }
 
--- a/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
+++ b/src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
@@ -15,16 +15,28 @@
 package build.buildfarm.common.admin.aws;
 
 import build.buildfarm.common.admin.Admin;
+import build.buildfarm.v1test.GetHostsResult;
+import build.buildfarm.v1test.Host;
 import com.amazonaws.services.ec2.AmazonEC2;
 import com.amazonaws.services.ec2.AmazonEC2ClientBuilder;
+import com.amazonaws.services.ec2.model.DescribeInstancesRequest;
+import com.amazonaws.services.ec2.model.DescribeInstancesResult;
+import com.amazonaws.services.ec2.model.Filter;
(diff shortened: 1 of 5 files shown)


src/main/java/build/buildfarm/common/admin/aws/AwsAdmin.java
```diff
@@ -56,4 +68,45 @@ public void stopContainer(String hostId, String containerName) {
...
+          hostBuilder.setState(e.getState().getName());
+          hostBuilder.setType(e.getInstanceType());
+          hostBuilder.setUptimeMinutes(uptime);
+          hosts.add(hostBuilder.build());
+        }
+      }
+    }
+    resultBuilder.addAllHosts(hosts);
+    resultBuilder.setNumHosts(hosts.size());
+    logger.info(String.format("Got %d hosts for filter: %s", hosts.size(), filter));
```
werkt, line 104: This looks like debug logging

src/main/java/build/buildfarm/server/AdminService.java
```diff
@@ -63,6 +65,23 @@ public void stopContainer(StopContainerRequest request, StreamObserver<Status> r
...
+      GetHostsResult result = null;
+      if (adminController != null) {
+        result =
+            adminController.getHosts(
+                request.getFilter(), request.getAgeInMinutes(), request.getStatus());
+      }
+      responseObserver.onNext(result);
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      logger.severe("Could not get hosts." + e);
```
werkt, line 80: This is a pattern here and throughout this file: You don't want to string-concat the message and 'e', you want to send the full stacktrace through to the logger with the verbose `.log(<level>, <message>, e)` format.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -39,6 +43,42 @@ message StopContainerRequest {
...
+
+  string dns_name = 6;
+
+  int64 uptime_minutes = 7;
+
+  int64 num_cores = 8;
+
+  string type = 9;
+
+  string lifecycle = 10;
```
werkt, line 79: this looks like it should be an enum, rather than a freeform string.

80degreeswest: AWS defines it as a string here. For most use cases this will be a single value or an empty string so I'd rather keep it as a string.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -39,6 +43,42 @@ message StopContainerRequest {
   string container_name = 3;
 }
 
+message GetHostsRequest {
+  string filter = 1;
+
+  int32 age_in_minutes = 2;
+
+  string status = 3;
```
luxe, line 51: what are some examples of status?  wondering where all those values are.

80degreeswest: in aws, it's spot and scheduled.

80degreeswest: in aws, it's running, terminated, etc.

PR #501, opened 2020-08-24
Store client uptime in Redis

80degreeswest:
The client (server/worker) uptime will be used for display purposes on the Buildfarm Admin dashboard.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -553,7 +553,7 @@ public class RedisShardBackplane implements ShardBackplane {
 
     // Record client start time
     client.call(
-      jedis -> jedis.set("uptime/" + clientPublicName, Long.toString(new Date().getTime())));
+        jedis -> jedis.set("uptime/" + clientPublicName, Long.toString(new Date().getTime())));
   }
 
   @Override
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -553,7 +553,7 @@ public class RedisShardBackplane implements ShardBackplane {
 
     // Record client start time
     client.call(
-        jedis -> jedis.set("uptime/" + clientPublicName, Long.toString(new Date().getTime())));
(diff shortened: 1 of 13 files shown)


80degreeswest:
> Make this an implementation detail of backplane.start for redis.
> Don't refer to it as a 'container' start, because it has nothing to do with a container. Backplane 'client' maybe.
> Don't use the `keyName` verbatim - add a prefix. Otherwise you will get collisions and no namespacing.

Got it!

> Set an expiry for the key entry.
> I would use the 'service' name (with the guid that is supposed to follow a single instance).
> Make the backplane refresh periodically, so that you don't a) get expiry when you don't want it, and b) can keep track of whether the thing is reporting in.

I think I'd rather keep it ip address based - we are not keeping instance IDs anywhere in BF and it's also cloud specific.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -549,6 +550,10 @@ public void start() throws IOException {
     if (config.getRunFailsafeOperation()) {
       startFailsafeOperationThread();
     }
+
+    // Record client start time
+    client.call(
+        jedis -> jedis.set("uptime/" + clientPublicName, Long.toString(new Date().getTime())));
```
luxe, line 556: I'd consider uptime the entire duration.  start_time might be less confusing and you can calculate uptime from it.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -786,7 +786,7 @@ public void run() {
 
   public void start() throws InterruptedException {
     try {
-      backplane.start();
+      backplane.start(config.getPublicName());
```
luxe, line 789: Not related to this diff, but I kind of wish we saved the entire config in redis so we could query anything about a worker/scheduler.  
key: ip / value: config

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
80degreeswest: Good point, I'll update it.

src/main/java/build/buildfarm/worker/shard/Worker.java
80degreeswest: Sounds like a new story for the backlog!

PR #503, opened 2020-08-25
Fix formatter script for OSX

80degreeswest:
The formatter script fails on OSX. This PR makes it compatible.


luxe:
might need to do it on 26: `if [[ "$@" == "--check" ]]`. not sure.

80degreeswest:
> might need to do it on 26: `if [[ "$@" == "--check" ]]`. not sure.

That one does not seem to be an issue.

PR #505, opened 2020-08-27
Add uptime grpc service to Admin

80degreeswest:


src/main/java/build/buildfarm/common/ShardBackplane.java
```diff
@@ -295,4 +296,7 @@ boolean pollOperation(QueueEntry queueEntry, ExecutionStage.Value stage, long re
 
   @ThreadSafe
   Boolean validQueueProperties(List<Platform.Property> provisions);
+
+  @ThreadSafe
```
luxe, line 300: seems like we make every method threadsafe here. do you know why?  we're only doing a read with jedis, but maybe it has something to do with only 1 jedis being used at a time?

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1375,4 +1377,17 @@ public OperationsStatus operationsStatus() throws IOException {
...
+  @Override
+  public GetClientStartTimeResult getClientStartTime(String clientKey) throws IOException {
+    try {
+      return client.call(
+          jedis ->
+              GetClientStartTimeResult.newBuilder()
+                  .setClientStartTime(Timestamps.fromMillis(Long.parseLong(jedis.get(clientKey))))
+                  .build());
+    } catch (NumberFormatException nfe) {
+      return GetClientStartTimeResult.newBuilder().build();
```
luxe, line 1390: seems fine.  I guess the time is 0 and we can identify any issues from that.

PR #514, opened 2020-09-09
Update shard-worker.config.example

uajith:
## To fix the issues related to 
```sh
com.google.protobuf.TextFormat$ParseException: 10:1: Input contains unknown fields and/or extensions:
```
And
```
provisioned_queues while starting the worker using   bf-run start command.


uajith:
shard server is not working with this configuration

PR #517, opened 2020-09-17
Configurable grpc timeouts for stub instances

luxe:
Stub instances have a hard-coded timeout of 1 minute.  
This may not be sufficient for `findMissingBlobs` depending on the size of the input.  
This diff is to ensure that the grpc timeout can be set from config.

### DIFF
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -286,7 +286,8 @@ public class Worker extends LoggingMain {
         break;
     }
 
-    workerStubs = WorkerStubs.create(digestUtil);
+    workerStubs =
+        WorkerStubs.create(digestUtil, config.getShardWorkerInstanceConfig().getGrpcTimeout());
 
     ExecutorService removeDirectoryService =
         newFixedThreadPool(
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -71,7 +71,9 @@ import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.devtools.common.options.OptionsParser;
 import com.google.longrunning.Operation;
 import com.google.protobuf.ByteString;
+import com.google.protobuf.Duration;
 import com.google.protobuf.TextFormat;
+import com.google.protobuf.util.Durations;
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
 import io.grpc.Status;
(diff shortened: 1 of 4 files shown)


80degreeswest (approved):

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -286,7 +286,8 @@ public Worker(String session, ServerBuilder<?> serverBuilder, ShardWorkerConfig
         break;
     }
 
-    workerStubs = WorkerStubs.create(digestUtil);
+    workerStubs =
```
80degreeswest, line 289: We should have a default timeout in case the config is missing one.

jlisee:
Is this `getMissingBlobs` or `findMissingBlobs`?

luxe:
> Is this `getMissingBlobs` or `findMissingBlobs`?

I mistyped.  `findMissingBlobs` is correct.  There is no "get" version.

src/main/java/build/buildfarm/worker/shard/Worker.java
luxe: 2 minute default if not configured

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -387,6 +388,25 @@ public static int KBtoBytes(int sizeKb) {
     return sizeKb * 1024;
   }
 
+  private static Duration GetGrpcTimeout(ShardWorkerConfig config) {
```
80degreeswest, line 391: method names should start with lowercase - getGrpcTimeout.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -387,6 +388,25 @@ public static int KBtoBytes(int sizeKb) {
...
+    // return the configured
+    if (config.getShardWorkerInstanceConfig().hasGrpcTimeout()) {
+      Duration configured = config.getShardWorkerInstanceConfig().getGrpcTimeout();
+      if (configured.getSeconds() > 0 || configured.getNanos() > 0) {
+        return configured;
+      }
+    }
+
+    // return a default
+    Duration defaultDuration = Durations.fromSeconds(120);
```
80degreeswest, line 402: Let's leave default as 60 here.

src/main/java/build/buildfarm/worker/shard/Worker.java
luxe: fixed in followup diff

src/main/java/build/buildfarm/worker/shard/Worker.java
luxe: fixed in followup diff

PR #518, opened 2020-09-17
allow skipping the CASFileCache load

luxe:
### Goal:
We want to experiment with the idea of starting workers without an initial cache.  
Loading the cache can be slow- upwards of 20 minutes depending on its size.
The faster a worker starts, the sooner it can take on work.  
We have issues currently where we can't scale to meet demands because of the slow startup time.
In the future, buildfarm-admin may tell workers to skip loading their cache based on the context of existing scaling issues.
^ we'll see based on experiments of having this functionality available

### Diff:
For starters, we customize the CASFileCache to allow skipping the load phase. 
If loading the cache is skipped, we instead delete those files.

### Test:
 - Added a test that shows load skipping.  
 - Edited the unit tests to report if loading occured or not.

### DIFF
--- a/src/main/java/build/buildfarm/CASTest.java
+++ b/src/main/java/build/buildfarm/CASTest.java
@@ -19,6 +19,7 @@ import static com.google.common.util.concurrent.MoreExecutors.newDirectExecutorS
 
 import build.bazel.remote.execution.v2.Digest;
 import build.buildfarm.cas.CASFileCache;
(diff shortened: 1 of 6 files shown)


jacobmou (approved):

luxe:
Note:  This should not affect production.  We make skipping cache load optional, but for all instances other than the CasTest program, loading the cache is still done.

PR #519, opened 2020-09-18
[refactor] Use Durations to simplify StubInstance

luxe:
### Problem:  
Buildfarm mixes `Duration` and `Deadline` types throughout the codebase.  
This is because timeouts are configured as a protobuf Duration, but later used to mean a grpc Deadline.  
In other places of the codebase, we pass around a value & a timepoint which eventually represents one of the more concrete types mentioned above.

### Refactor:  
 - we change StubInstance to expect Duration types instead of value + timepoints.
 - we provide utilities to easily convert between Duration/Deadline as needed.

### DIFF
--- a/src/main/java/build/buildfarm/common/Time.java
+++ b/src/main/java/build/buildfarm/common/Time.java
@@ -28,22 +28,33 @@ public class Time {
 
   ///
   /// @brief   Convert a protobuf duration to a grpc deadline.
-  /// @details Deadline will be represented in seconds.
+  /// @details Deadline will have nanosecond precision.
   /// @param   duration A protobuf duration.
   /// @return  A converted grpc deadline.
   /// @note    Suggested return identifier: deadline.
   ///
   public static Deadline toDeadline(Duration duration) {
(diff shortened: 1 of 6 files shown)


src/main/java/build/buildfarm/common/Time.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+  ///
+  /// @brief   Convert a grpc deadline to a protobuf duration.
+  /// @details Duration will be represented in seconds.
+  /// @param   deadline A converted grpc deadline.
+  /// @return  A protobuf duration.
+  /// @note    Suggested return identifier: duration.
+  ///
+  public static Duration toDuration(Deadline deadline) {
+    return Durations.fromSeconds(deadline.timeRemaining(TimeUnit.SECONDS));
```
werkt, line 47: Make this load from nanos, regardless of use, you don't want to throw away portions of a second.

werkt (approved):

src/main/java/build/buildfarm/common/Time.java
luxe: fix

PR #541, opened 2020-10-07
update bazel to 3.6.0, update rules_docker, rules_jvm_external, com_g…

C00IHandLuke:
…oogle_protobuf, io_grpc_grpc_java, and remote_apis.

add buildifier as it blocks a commit if buildifier doesn't pass so now you can also run it with bazel

Updating bazel to 3.6.0 required some of the deps to be updated, I tried to keep the style similar but if there is a better way let me know.


luxe:
thanks Luke!  I enabled container builds as part of the CI so we could more confidently apply your changes.  I'm going to do just the docker rules first because of https://github.com/bazelbuild/bazel-buildfarm/issues/553

C00IHandLuke:
Sounds good, I'll come back and clean this up then.

Thanks for setting the tags in the ci for the containers!

C00IHandLuke:
Hi @luxe, everything should be up to date now and working with the ci flags you added. If there is anything else blocking for this let me know, thanks

PR #550, opened 2020-10-14
add file directory index in memory implementation and make choosing of in memory and in sqlite configurable

jacobmou:
1) Make using of sqlite for file <-> directory bidirectional mapping configurable;
2) add in-memory implementation for the bi-directional mapping.

Test result:
1) This can reduce the worker startup time by ~30%, but will increase memory cost significantly. 
2) run rNA build against it


src/main/java/build/buildfarm/cas/MemoryFileDirectoriesIndex.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public MemoryFileDirectoriesIndex(Path root) {
+    super(root);
+    data = new ConcurrentHashMap<>();
+  }
+
+  @Override
+  public void close() {
+    throw new UnsupportedOperationException(
+        "MemoryFileDirectoriesIndex doesn't support close() operation.");
```
luxe, line 51: wonder if close() should be an empty function.  that way the client can call it regardless of what the inherited object is. like how start() is.

src/main/java/build/buildfarm/cas/MemoryFileDirectoriesIndex.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Use ConcurrentHashMap in memory to store bidirectional entry directory mapping. Comparing to
+ * using sqlite, this should reduce worker startup time a lot, but will also cause high memory
+ * usage.
+ */
+class MemoryFileDirectoriesIndex extends DirectoriesIndex {
+  private static final Charset UTF_8 = Charset.forName("UTF-8");
+
+  private final ConcurrentHashMap<String, Set<String>> data;
```
luxe, line 41: Does `Set` need to be threadsafe also?

luxe (approved):
Is there a max memory limit on the indexing?  Or do we assume the memory is sufficient based on CAS size?

Since we can configure it on/off easily, I'm good to land!

src/main/java/build/buildfarm/cas/MemoryFileDirectoriesIndex.java
jacobmou: Yeah! Good catch! usecases like `data.get(entry).add(digest);`  will be a problem!

jacobmou:
> Is there a max memory limit on the indexing? Or do we assume the memory is sufficient based on CAS size?
> 
> Since we can configure it on/off easily, I'm good to land!

We just `assume` memory is enough, because that's essentially what we do before. The memory issue start to exacerbate after/around 36 hours. With the new java and gc, hopefully, it could be alleviated

PR #556, opened 2020-10-19
80degreeswest/performance fixes

80degreeswest:
Couple of adjustments to increase performance based on benchmarks and refactor logging.


src/main/java/build/buildfarm/server/ContentAddressableStorageService.java
```diff
@@ -112,11 +112,14 @@ public void onSuccess(FindMissingBlobsResponse.Builder builder) {
...
-                      instance.getName(),
-                      request.getBlobDigestsList().size(),
-                      elapsedMicros / 1000.0));
+                  new StringBuilder()
+                    .append("FindMissingBlobs(")
+                    .append(instance.getName())
+                    .append(") for ")
+                    .append(request.getBlobDigestsList().size())
+                    .append(" blobs in ")
+                    .append(elapsedMicros / 1000.0)
```
luxe, line 121: .append("ms")

luxe (approved):
looks good.  Didn't realize how much logging was currently had.  We can probably adjust how much we log in the future since we have other methods of introspection now.

jacobmou:
> looks good. Didn't realize how much logging we currently had. We could probably adjust how much we log in the future since we have other methods of introspection now.

Yeah! Good idea! But we might still need many logs as we probably cannot have datadog on every worker/scheduler.

jacobmou (approved):

luxe:
@80degreeswest land this when you can! Want to get it on master since it changes many files.

src/main/java/build/buildfarm/server/ContentAddressableStorageService.java
80degreeswest: Yes, the nexts step will be to reduce most of INFO type logging to FINE and send metrics based logging elsewhere.

80degreeswest:
> looks good. Didn't realize how much logging we currently had. We could probably adjust how much we log in the future since we have other methods of introspection now.

Yes, the nexts step will be to reduce most of INFO type logging to FINE and send metrics based logging elsewhere.

src/main/java/build/buildfarm/server/ContentAddressableStorageService.java
80degreeswest: I'm going to leave it as is since we have log parsing that will break if we modify the content. This will be going away in favor of actual metrics sending soon anyways.

PR #557, opened 2020-10-20
add CAS Index API to admin service

luxe:
### CAS Indexes
The `RedisShardBackplane` contains CAS keys which point to sets of instance names.  
These instance names indicate where that CAS data can be found.  

For context, here is an example of that data using `redis-cli`: 
```
redis-cli
127.0.0.1:6379> keys Content*
   1) "ContentAddressableStorage:3c061aba22a4b14e84de5f1e6a4b5b7ea1f185da33cc248aba2a81f419573dda/1851"
   2) "ContentAddressableStorage:e4cced2b4dcf56d3023384a90f3f7c09f7bdfe1ba60dce670306ac50a5050f97/3587"
   3) "ContentAddressableStorage:8ca0e48f1a62f3440ef7857b721cb74ec60900d3219d28ec40628850635e9ffd/15010"
   4) "ContentAddressableStorage:271cdbddce4a65bb9a2c84ed57c6a9e05d5f2caea9b79cb800ea3aa4f1882622/3623"
   5) "ContentAddressableStorage:32e95ad3ba8016388dd7af03e93b6c7c24937f4706d3a471af6ff3c8fc6be722/6043"
   6) "ContentAddressableStorage:83bc7b4c34df3b1dc3409f01900acecf103d5d920ecfcfe1f46fa6c03b58cb53/4886"
   ...

127.0.0.1:6379> smembers "ContentAddressableStorage:be14f4ca7cb6decfc3f6dbec569720be27691efa12413b987bc76763b03574b2/2624"
1) "localhost:8983"
2) "localhost:8981"
```
### Re-indexing
When workers leave the cluster, these keys must be updated to inform other workers where the data can be received.  
Without cleaning up these indexes on departed workers, buildfarm will become slower as it will take workers longer to find valid workers with the needed CAS data.  The current solution is to periodically run an indexer which is being done outside of buildfarm.  There was a [previous attempt](https://github.com/bazelbuild/bazel-buildfarm/pull/422) to start running this indexer periodically inside buildfarm.  Now we want the reindexing to be done by buildfarm admin through the admin service:  
Buildfarm Worker Shutdowns

### Testing Tool
We can use the stub instance to test this grpc functionality.  
1. Run buildfarm with 2 workers.
2. Populate the CAS via compilations
3. Perform Cas Reindexing

`./bazelw run //src/main/java/build/buildfarm:bf-index-worker localhost:8980 shard SHA256 localhost:8981`
```
Total keys re-indexed: 7613 Worker references removed: 3460 CAS keys deleted: 473 CAS lost: 6.213057%
```
`./bazelw run //src/main/java/build/buildfarm:bf-index-worker localhost:8980 shard SHA256 localhost:8982`
```
Total keys re-indexed: 3570 Worker references removed: 3570 CAS keys deleted: 3570 CAS lost: 100.000000%
```


80degreeswest:
> **First,** if the worker is being gracefully shutdown, it knows that it will no longer be available to provide CAS information. It can unregister itself from the CAS as part of its shutdown process.

What happens if the worker is just being restarted and the CAS is not cleared out as it is with the host shutdown?

src/main/java/build/buildfarm/common/CasIndexResults.java
```diff
@@ -0,0 +1,47 @@
...
+///          remove the CAS key.
+///
+public class CasIndexResults {
+
+  ///
+  /// @field   removedInstances
+  /// @brief   The number of CAS entries the worker was removed from.
+  /// @details This indicates how much CAS data the shard new the worker had.
+  ///
+  public int removedInstances = 0;
```
80degreeswest, line 30: Let's call them hosts here. Instance is probably too AWS specific.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -645,6 +647,60 @@ public boolean removeWorker(String name, String reason) throws IOException {
...
+                String casQuery = config.getCasPrefix() + ":*";
+                ScanParams params = new ScanParams();
+                params.match(casQuery);
+                params.count(defaultScanCount);
+
+                // iterate over all CAS entries via scanning
+                // and remove worker from the CAS keys.
+                String nextCursor = cursorSentinal;
+                do {
+
```
80degreeswest, line 676: Remove newline

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -645,6 +647,60 @@ public boolean removeWorker(String name, String reason) throws IOException {
...
+                    nextCursor = scanResult.getCursor();
+                  }
+                } while (!nextCursor.equals(cursorSentinal));
+              }
+            });
+
+    return results;
+  }
+
+  private void RemoveWorkerFromCasKeys(
```
80degreeswest, line 689: Method names should start with lowercase letter.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -645,6 +647,60 @@ public boolean removeWorker(String name, String reason) throws IOException {
...
+      JedisCluster cluster, CasIndexResults results, List<String> casKeys, String workerName) {
+    for (String casKey : casKeys) {
+      results.totalKeys++;
+      Long wasRemoved = cluster.srem(casKey, workerName);
+      if (wasRemoved == 1) {
```
80degreeswest, line 694: Could just do if (cluster.srem(casKey, workerName) == 1) here since wasRemoved is not used anywhere else.

luxe:
> > **First,** if the worker is being gracefully shutdown, it knows that it will no longer be available to provide CAS information. It can unregister itself from the CAS as part of its shutdown process.
> 
> What happens if the worker is just being restarted and the CAS is not cleared out as it is with the host shutdown?  

Good point.  My expectation is that this will need to evolve as we make "graceful shutdown" and " graceful restart" more explicit in code.  If we get a standalone "Cleanup the CAS for this worker" module, then it could be later utilized via more sophisticated communication with buildfarm admin or a heartbeat monitor that understands this kind of context.  

Right now the worker "gracefully shutdowns" via `stop()`.  We could configure this to decide whether or not to clear indexes, but I would still need some context on deciding when to set that flag.  Thoughts?

src/main/java/build/buildfarm/common/CasIndexResults.java
```diff
@@ -0,0 +1,47 @@
...
+  /// @brief   The number of CAS entries removed due to loss of worker.
+  /// @details This indicates how many CAS entries were held only by the
+  ///          removed worker.
+  ///
+  public int removedKeys = 0;
+
+  ///
+  /// @field   totalKeys
+  /// @brief   The total number of keys processed.
+  /// @details A raito can be made with removed keys to see the percentage of
```
jacobmou, line 43: typo, raito -> ration

src/main/java/build/buildfarm/common/CasIndexResults.java
luxe: fix

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: fix

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: fix

src/main/java/build/buildfarm/common/CasIndexResults.java
luxe: fix

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: fix

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -35,6 +35,24 @@ service Admin {
...
+  
+  rpc ReindexCas(ReindexCasRequest) returns (ReindexCasRequestResults) {
+    option (google.api.http) = { get: "/v1test/admin:reindexCas" };
+  }
+}
+
+message ReindexCasRequest {
+  string instance_name = 1;
+  
+  string host_id = 2;
```
80degreeswest, line 47: What are we using as host ID? A full worker name in the IP:PORT format? Or just IP address? Maybe either a host_name or worker_name would be better here. Also can you add a comment here on what the expected format is.

80degreeswest (approved):

src/main/java/build/buildfarm/IndexWorker.java
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
```
80degreeswest, line 1: Do you think we need this tool since we are exposing this via a grpc service? Would a test be sufficient enough?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
jacobmou: Yes, i keep wondering about the same thing when going through the code, i think it's `ip:port`, like `localhost:8981`.

jacobmou (approved):

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: I was also unsure about the name.  Other "request types" use `host_id` so I followed that pattern, ex:
```
message TerminateHostRequest {
  string instance_name = 1;
  string host_id = 2;
}
```

In this case, `host_id` is the `ip:port` because it correlates to the key format that's in `smembers  ContentAddressableStorage:*`.  I'll document that it should be that format. But also willing to change the name if it can improve clarity.

src/main/java/build/buildfarm/IndexWorker.java
luxe: At the very least, I should add a mock test.  My issue has been that jedis throws weird runtime exceptions when you use the library in subtly incorrect ways. For example, I couldn't iterate over the whole cluster and do a certain key operations, but instead had to iterate over the cluster nodes individually which the mock wouldn't have caught.  The closest thing we have to an integration-level (for stressing grpc/redis) is probably building rNA with dev?  

The client side tool lets me mimic bf-admin and get some more confidence before landing.
I see the concern of having a bunch of random tools that hit grpc endpoints though.  Maybe they could be consolidated better? Or if we could somehow run bf-admin locally, and evoke changes that way, I would have no need for the tools.

luxe: I'm going to land and then followup with the other ideas of testing. We could remove this in the future.

PR #559, opened 2020-10-22
change tool name for CasFileCache

luxe:
We have an existing tool to load the CasFileCache and print it's startup information.  The name of the tool is too vague and I'd like to make other CAS-related tools for debugging.  We can rename it.


jacobmou (approved):
Nice

PR #560, opened 2020-10-22
fix shardWorker config

luxe:
I've also broken these config files.  We'll add unit tests soon: https://github.com/bazelbuild/bazel-buildfarm/pull/561


jacobmou (approved):
thanks!

PR #565, opened 2020-10-27
Speed up startup time by batching Redis calls

walles:
The following changes will reduce the amount of calls to Redis and can accelerate the startup time by a factor of two.

Before this change, one call per cache entry was made to Redis.

With this change in place, we collect a list of all the cache entires during cache scanning, and then make one call with all of them at the end.

### DIFF
--- a/src/main/java/build/buildfarm/cas/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/CASFileCache.java
@@ -95,8 +95,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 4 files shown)


werkt:
Closing this in favor of #566 - we previously had this aggregated, but it got lost in refactor. Thanks for the prompt to go fix it!

walles:
I'm probably missing something here, but how does #566 improve startup time? It looks like just renaming a variable.

This PR (#565) batches all startup Redis calls into one, which I can't see how #566 does.

@werkt, what am I missing?

In our tests, one large call to Redis (this PR) was much faster than millions of small calls (still the case after #566 IIRC).

For reference, the cache sizes we've been experimenting on are 50GB/11M entries.

werkt:
The aggregation takes place at https://github.com/bazelbuild/bazel-buildfarm/blob/fd6255d9f1a2b7638a17c1d2a38a8886e7db569c/src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java#L118-L120, where scanned digests are placed into a list. There is then a single call to the supplied `onDigests` consumer, which translates via https://github.com/bazelbuild/bazel-buildfarm/blob/fd6255d9f1a2b7638a17c1d2a38a8886e7db569c/src/main/java/build/buildfarm/worker/shard/Worker.java#L824-L825 as a call to https://github.com/bazelbuild/bazel-buildfarm/blob/fd6255d9f1a2b7638a17c1d2a38a8886e7db569c/src/main/java/build/buildfarm/worker/shard/Worker.java#L740-L753. The `addBlobsLocation` implementation under https://github.com/bazelbuild/bazel-buildfarm/blob/fd6255d9f1a2b7638a17c1d2a38a8886e7db569c/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java#L855-L867 for redis uses cluster pipelining to reduce command/response latency.

This was all in place previously, and had been accidentally subverted, which is why it only took 'renaming a variable' - in reality, propagating and removing a shadow - to reenable it.

walles:
Makes sense, thank @werkt for the thorough explanation!

Not sure this is thread safe, opened #567 about that.

PR #570, opened 2020-10-28
log an error when proceedToOutput fails

luxe:


src/main/java/build/buildfarm/worker/InputFetcher.java
```diff
@@ -213,6 +213,8 @@ private long fetchPolled(Stopwatch stopwatch) throws InterruptedException {
     } finally {
       if (!completed) {
         try {
+          logger.log(
+              Level.SEVERE, format("error claiming %s input. Must remove execDir", operationName));
```
werkt, line 217: This doesn't need to be in the try block, and should be formatted similar to the other error messages in this module.
The error case itself only occurs on interrupt, which will be unlikely except in the event of a shutdown, or a runtime exception, which will be logged by line 282. What exactly is the reason behind logging this?

luxe: lack of visibility. we were speculating if this code path was hit

PR #572, opened 2020-10-29
configurable max cpu on platform

luxe:
Allow setting the max cpu setting in which buildfarm will accept actions.  
Replaced constant [here](https://github.com/bazelbuild/bazel-buildfarm/pull/572/files#diff-7d364025210bb06a44a529f14efcdf9a0bbbd8e715af51786c4f0ca24310fe52R1408)
```
PlatformValidationSettings {
    max_cpu: 80
}
```


luxe:
too many files changed passing it down the call stack.  i'll do it again and store in a context

PR #575, opened 2020-10-29
show pid commands when erroring on lingering command resources

luxe:


werkt (changes requested):
Leaving a hanging change request here so that I have to review these before merging.

80degreeswest (approved):

PR #576, opened 2020-10-30
disable action failure when cgroup resources remain

luxe:
When a process finishes running under cgroups, it may have lingering resources that it did not cleanup.  
We want to configure whether this is an error or not.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -216,8 +216,7 @@ class Executor {
       // If the main process of a test exits, but some of its children are still running,
       // the test runner should consider the run complete and count it as a success or failure
       // based on the exit code observed from the main process. The test runner may kill any stray
-      // processes.
-      // Tests should not leak processes in this fashion.
+      // processes. Tests should not leak processes in this fashion.
       if (workerContext.shouldErrorOperationOnRemainingResources() && resource.isReferenced()) {
 
         // there should no longer be any references to the resource. Any references will be
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -62,15 +62,10 @@ class Executor {
   private final ExecuteActionStage owner;
   private int exitCode = INCOMPLETE_EXIT_CODE;
   private boolean wasErrored = false;
-  private final boolean errorOperationRemainingResources;
 
   Executor(
-      WorkerContext workerContext,
-      boolean errorOperationRemainingResources,
-      OperationContext operationContext,
-      ExecuteActionStage owner) {
+      WorkerContext workerContext, OperationContext operationContext, ExecuteActionStage owner) {
     this.workerContext = workerContext;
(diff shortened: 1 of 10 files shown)


werkt:
The pattern so far has been to ask the WorkerContext about configuration - it has kept the stages light on state. I'd prefer you add a method there, and call that to ask about the current setting - you may want to change it on the fly, and that will make this more adaptable.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -212,7 +217,7 @@ private long executePolled(
               "", // executingMetadata.getStderrStreamName(),
               resultBuilder);
 
-      if (resource.isReferenced()) {
+      if (resource.isReferenced() && errorOperationRemainingResources) {
```
werkt, line 220: Swap these checks. The method may be a more substantial performance impact than the config check.

werkt (changes requested):

src/main/java/build/buildfarm/worker/Executor.java
luxe: fix

luxe:
> The pattern so far has been to ask the WorkerContext about configuration - it has kept the stages light on state. I'd prefer you add a method there, and call that to ask about the current setting - you may want to change it on the fly, and that will make this more adaptable.
makes more sense.  fixed.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -212,7 +212,14 @@ private long executePolled(
...
 
-      if (resource.isReferenced()) {
+      // From Bazel Test Encyclopedia:
+      // If the main process of a test exits, but some of its children are still running,
+      // the test runner should consider the run complete and count it as a success or failure
+      // based on the exit code observed from the main process. The test runner may kill any stray
+      // processes. Tests should not leak processes in this fashion.
+      // Based on configuration, we will decide whether remaining resources should be an error.
+      if (workerContext.shouldErrorOperationOnRemainingResources() && resource.isReferenced()) {
+
```
werkt, line 222: Check your whitespace here

src/test/java/build/buildfarm/worker/StubWorkerContext.java
```diff
@@ -207,4 +207,9 @@ public IOResource limitExecution(
   public int commandExecutionClaims(Command command) {
     throw new UnsupportedOperationException();
   }
+
+  @Override
+  public boolean shouldErrorOperationOnRemainingResources() {
+    throw new UnsupportedOperationException();
```
werkt, line 213: This call happens for every execution, whether the IOResource is meaningful or not. Have it return the configured value, as you've added it to the operationqueue worker config definition.

werkt (approved):

werkt:
Approved pending addressed comments.

src/test/java/build/buildfarm/worker/StubWorkerContext.java
werkt: Apologies, I misread the context and thought this was the operationqueue worker

PR #580, opened 2020-10-31
Propagate errors in ByteStreamHelper.

AngusDavis:


werkt (changes requested):
It's not clear what issue this is solving. Please expand upon the intent of the change in the description, and provide a test case for what this future set guards against - the intent of setting only the inputStream's exception is that it will be available for stream consumers.

AngusDavis:
> It's not clear what issue this is solving. Please expand upon the intent of the change in the description, and provide a test case for what this future set guards against - the intent of setting only the inputStream's exception is that it will be available for stream consumers.

Ack, I'll work on it (will add to comments as well) but the overall gist is that if an error is raised from a retry before scheduling another retry, then the unbounded `streamReadyFuture.get()` on 141 can hang until interrupted.

PR #584, opened 2020-11-07
creates queues from config

luxe:
Simplify the startup of the shard backplane by providing config to queue creation functions.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -495,7 +495,7 @@ public class RedisShardBackplane implements ShardBackplane {
     failsafeOperationThread.start();
   }
 
-  private SetMultimap<String, String> toMultimap(List<Platform.Property> provisions) {
+  private static SetMultimap<String, String> toMultimap(List<Platform.Property> provisions) {
     SetMultimap<String, String> set = LinkedHashMultimap.create();
     for (Platform.Property property : provisions) {
       set.put(property.getName(), property.getValue());
@@ -510,8 +510,8 @@ public class RedisShardBackplane implements ShardBackplane {
     // multiple clients.
     client = new RedisClient(jedisClusterFactory.get());
 
-    this.prequeue = createPrequeue(config);
-    this.operationQueue = createOperationQueue(config);
+    prequeue = createPrequeue(client, config);
+    operationQueue = createOperationQueue(client, config);
 
     if (config.getSubscribeToBackplane()) {
       startSubscriptionThread();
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -505,18 +505,34 @@ private void startFailsafeOperationThread() {
...
   @Override
   public void start(String clientPublicName) throws IOException {
-
     // Construct a single redis client to be used throughout the entire backplane.
     // We wish to avoid various synchronous and error handling issues that could occur when using
     // multiple clients.
     client = new RedisClient(jedisClusterFactory.get());
 
+    this.prequeue = createPrequeue(config);
+    this.operationQueue = createOperationQueue(config);
```
werkt, line 514: No need for `this.` if it doesn't shadow a parameter.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -505,18 +505,34 @@ private void startFailsafeOperationThread() {
...
+    if (config.getSubscribeToBackplane()) {
+      startSubscriptionThread();
+    }
+    if (config.getRunFailsafeOperation()) {
+      startFailsafeOperationThread();
+    }
+
+    // Record client start time
+    client.call(
+        jedis -> jedis.set("startTime/" + clientPublicName, Long.toString(new Date().getTime())));
```
werkt, line 525: I see that this was just moved, but it really shouldn't be set without an expiration, and the timestamp acquisition is pretty awkward - why not just make it human readable?

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -505,18 +505,34 @@ private void startFailsafeOperationThread() {
...
+  }
+
+  BalancedRedisQueue createPrequeue(RedisShardBackplaneConfig config) throws IOException {
```
werkt, line 528: This looks like it qualifies for static

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -505,18 +505,34 @@ private void startFailsafeOperationThread() {
...
-    this.prequeue =
-        new BalancedRedisQueue(
-            config.getPreQueuedOperationsListName(),
-            getQueueHashes(config.getPreQueuedOperationsListName()));
+    return new BalancedRedisQueue(
+        config.getPreQueuedOperationsListName(),
+        getQueueHashes(config.getPreQueuedOperationsListName()));
+  }
 
+  OperationQueue createOperationQueue(RedisShardBackplaneConfig config) throws IOException {
```
werkt, line 535: Likewise qualified for static.

werkt (approved):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: fixed

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: changed the "hash generation" func to take the client as an argument, so these "queue creation" functions can also take the client and now be marked static.  
that's better.  when things are marked static they become obvious candidates to move out of the backplane in the future.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: fixed

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: maybe the format is best suited for bf-admin? let me see about changing. @80degreeswest is it safe to expire this key?

PR #587, opened 2020-11-11
Describe what to do when no eligible queues are found

thna123459:
Since the default shard-worker.example does not work out-of-the-box,
it is better to provide some documentation when the exception is thrown.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
+++ b/src/main/java/build/buildfarm/instance/shard/OperationQueue.java
@@ -219,7 +219,9 @@ public class OperationQueue {
       }
     }
     throw new RuntimeException(
-        "there are no eligible queues for the provided execution requirements.  One solution to is to configure a provision queue with no requirements which would be eligible to all operations.");
+        "there are no eligible queues for the provided execution requirements."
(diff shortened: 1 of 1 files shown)


luxe (approved):
nice improvement!  Let us know if anything is not clear in the wiki.  

@werkt By the way, I don't think users are able to make pull-requests to the wiki.  We had people internally wanting to do that.  Is there an option you can enable as an admin?

werkt:
I'll see about it. I'm also a little leery of putting in bespoke references to documentation in the RuntimeException here, but there's far more cleanup that we can do on this front besides this.

werkt:
I'd also prefer that the default config work out of the box - how is the circumstance encountered here by default?

There's a page describing a complex process by which the wiki is cloned into its own primary repo, then it can be forked and PR'd like normal, with some mechanism to sync the primary repo back to the project wiki. This seems pretty onerous - for me, I'd just say that we can add individual users, or have documentation issues describing the nature of the desired adds.

thna123459:
Yes, it would be best to have a working configuration out-of-the-box. In my case, I was updating a worker configuration that others had setup a few months ago; it took me some time to realise that the default settings would lead to failures.

PR #589, opened 2020-11-14
CASFileCache cleanup/performance/path strategy

werkt:
Use of json library to generate serialized objects for logging
Short circuit exists check for reset writes
FileStore parameterization to prevent in-loop mount lookups
  This could still use some work, much is replicated in common.io.Utils,
  EvenMoreFiles, and others.
Moved all common IO-related mechanisms to common.io
Moved CASFileCache under its own cfc module for further refactoring
CASFileCache now uses a path entry strategy with configurable depth for
more expedient use on directory-size performance-afflicted filesystems


luxe:
awesome.  I'll review this soon

luxe (commented):
check my understanding: 
Apart from cleaning things up, the main goal is to enable the CasFileCache to save files in nested directories?  Because its an issue for to have many files all saved at root in a flat directory structure.  This is perhaps in response to some of the goals here:  https://github.com/bazelbuild/bazel-buildfarm/issues/568

We can test this out in our dev environment to ensure nothing breaks.

werkt:
> check my understanding:
> Apart from cleaning things up, the main goal is to enable the CasFileCache to save files in nested directories? Because its an issue for to have many files all saved at root in a flat directory structure. This is perhaps in response to some of the goals here: #568
> 
> We can test this out in our dev environment to ensure nothing breaks.

Correct. The unchanged config, with a 0 value in the hex_bucket_levels, should have no behavior changes

> check my understanding:
> Apart from cleaning things up, the main goal is to enable the CasFileCache to save files in nested directories? Because its an issue for to have many files all saved at root in a flat directory structure. This is perhaps in response to some of the goals here: #568
> 
> We can test this out in our dev environment to ensure nothing breaks.

Correct. The unchanged config, with a 0 value in the hex_bucket_levels, should have no behavior changes.

luxe (approved):
conceptually this all makes sense.  We should land since we can also opt-out for now.  We haven't had time to test it personally.

PR #592, opened 2020-11-17
make worker profile be able to profile all workers in a loop and monitor their status

jacobmou:
`workerprofile` is now in a separate class, which can be invoked either locally or remotely to profile workers. 
The command that invoking it would be `bazel-bin/src/main/java/build/buildfarm/bf-workerprofile WorkerProfile shard SHA256 examples/shard-worker.config.example`, with the config file containing correct redis uri.

A few things need to be done later:
- clean up profile related code from bf-cat
- logging profile related message from worker directly, so we can access these logs from honeycomb and set up alert and automatic actions


src/main/java/build/buildfarm/Cat.java
```diff
@@ -607,7 +621,150 @@ static int deadlineSecondsForType(String type) {
...
   }
 
+  /**
+   * Transform worker string from "ip-10-135-31-210.ec2:8981" to "10-135-31-210".
+   *
+   * @param worker
+   * @return
+   */
+  private static String workerStringTransformation(String worker) {
+    String result = worker.split("\\.")[0].substring("ip-".length()).replaceAll("-", ".") + ":8981";
```
werkt, line 631: Like, I know what this is from? but this is not a job for cat. Please come up with your own sed to do this outside, and pass it in as a parameter.

werkt (changes requested):

luxe (approved):
Having this tool available will be hugely beneficial.  I agree with George that we'll want to generalize any specific use cases like the IP addresses.  In terms of all the output that the tool does, I think its best if we just start using it in production and shape anything based on that experience.

src/main/java/build/buildfarm/Cat.java
jacobmou: I click `resolve` because i move the new thing to a separate class. Yeah, I definitely will make this string format change out in the future, but for now I didn't see anything else that would potentially use it. 

Also, not super sure what ` come up with your own sed to do this outside` mean here. This is a IP string format change to make grpc call work, the old worker ip string comes from redis directly, not passed in by cmd args.

src/main/java/build/buildfarm/WorkerProfile.java
```diff
@@ -0,0 +1,354 @@
...
+          }
+        }
+
+        // add new registered workers and profile
+        for (String worker : workers) {
+          if (!workersToChannels.containsKey(worker)) {
+            workersToChannels.put(
+                worker,
+                new StubInstance(
+                    "shard",
```
werkt, line 201: This is a hardcoded instance name.

src/main/java/build/buildfarm/WorkerProfile.java
```diff
@@ -0,0 +1,354 @@
...
+  }
+
+  /**
+   * Transform worker string from "ip-10-135-31-210.ec2:8981" to "10-135-31-210".
+   *
+   * @param worker
+   * @return
+   */
+  private static String workerStringTransformation(String worker) {
+    return worker.split("\\.")[0].substring("ip-".length()).replaceAll("-", ".") + ":8981";
```
werkt, line 70: This is still pretty ugly, and speaks to a problem with DNS, hardcoded ports, and an implementation detail in doling out hostnames: I would honestly rather you have a DNS proxy/adapter in place that resolves these names than it be hardcoded in this tool.

src/main/java/build/buildfarm/worker/shard/WorkerProfileService.java
```diff
@@ -98,4 +105,17 @@ public void getWorkerProfile(
...
   }
+
+  @Override
+  public void getWorkerList(
+      WorkerListRequest request, StreamObserver<WorkerListMessage> responseObserver) {
+    WorkerListMessage.Builder replyBuilder = WorkerListMessage.newBuilder();
+    try {
+      replyBuilder.addAllWorkers(backplane.getWorkers());
+    } catch (IOException e) {
+      e.printStackTrace();
```
werkt, line 116: This error should be relayed to the observer with `onError`

werkt (approved):
Some things there are ugly, including the DNS stuff and some semantics.

PR #604, opened 2020-11-25
Document Redis timeout and maxAttempts settings

thna123459:


luxe (approved):
This looks correct for how JedisClusterFactory is setup

PR #607, opened 2020-12-01
Gitignore more things

walles:
Tell Git to ignore IntelliJ IDEA and Eclipse files.

Having this change merged would make it nicer for us to work with the Buildfarm source code.


werkt (approved):
I suppose these are mostly harmless.

PR #608, opened 2020-12-01
Add cfc blob overwrite protection

werkt:


jacobmou (approved):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2704,6 +2709,10 @@ public void write(byte[] b) throws IOException {
 
       @Override
       public void write(byte[] b, int off, int len) throws IOException {
+        if (written + len > blobSizeInBytes) {
+          throw new IOException(
+              format("attempted overwrite at %d by %d bytes for %s", written, len, writeKey));
```
walles, line 2714: Shouldn't the `by` number be `blobSizeInBytes - (written + len)`?

werkt: The 'by' (and the message itself) captures that the write operation itself is an 'overwrite', with the interpretation of the overage up to the user to compute - I wanted to make it explicit that the numbers don't line up, so you see an 'overwrite at 10 by 20 for deadbeef/12'. Perhaps 'of size' or some language improvement would work out better there, but it is as intended.

walles: My confusion was that writing 30 bytes to a 10 bytes buffer gives:
* `attempted overwrite at 10 by 30 bytes for XXX`, I interpreted this as having written 30 bytes too many

I would have expected the amount of overshoot:
* `attempted overwrite at 10 by 20 bytes for XXX`

Or maybe this?
* `attempted write of 30 bytes to 10 bytes buffer for XXX`

But in the end, no matter the formulation of this message, it does convey there was an overwrite, and anybody who's unsure about the numbers has the source code at hand, so this is no biggie.

PR #610, opened 2020-12-01
Permit stub deadlining of <1s

werkt:


luxe (approved):
ty! at some point we'll hopefully verify full protos in 1 place, so the rest of code can assume things are valid

PR #612, opened 2020-12-03
Extend worker match evaluation

luxe:
### Worker re-queuing 
When a worker gets an operation they must make a decision as to whether they should keep that operation or put it back on the queue.  We wish to customize this decision process and add testing where there previously was not any.  

### Feature request
There is a subtle issue that we are trying to address.
Since the `OperationQueue` has been extended to allow a superset of `execution_properties`, so too must the workers be able to accept a superset of properties on a `queueEntry`.  If a user were to craft an invalid execution property it may make it onto the queue but be constantly rejected by all the workers and never actually execute.  A similar flag to the queue's `allowUnmatched` has been added.  This can be made configurable  in the future.

### Superset rationale
One might wonder why we allow these supersets at all, and the reason is to better accommodate other remote execution solutions who are specifying their own execution_properties in the same repository.  Whether you want to allow supersets or not is something that should be fully customizable so we are not enforcing any kind of paradigm.  If users want to lock down a strict set of allowed execution properties, this is still possible via queue configuration.

### DIFF
--- a/src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
+++ b/src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
@@ -16,7 +16,6 @@ package build.buildfarm.worker.shard;
 
 import static build.buildfarm.cas.ContentAddressableStorage.UNLIMITED_ENTRY_SIZE_MAX;
 import static build.buildfarm.common.Actions.checkPreconditionFailure;
-import static build.buildfarm.common.Actions.satisfiesRequirements;
 import static build.buildfarm.common.Errors.VIOLATION_TYPE_INVALID;
 import static build.buildfarm.common.Errors.VIOLATION_TYPE_MISSING;
 import static java.lang.String.format;
@@ -96,6 +95,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import build.buildfarm.worker.DequeueMatchEvaluator;
+import build.buildfarm.worker.DequeueMatchSettings;
 
 class ShardWorkerContext implements WorkerContext {
   private static final Logger logger = Logger.getLogger(ShardWorkerContext.class.getName());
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -311,7 +312,10 @@ private void matchInterruptible(MatchListener listener) throws IOException, Inte
       // transient backplane errors will propagate a null queueEntry
     }
     listener.onWaitEnd();
-    if (queueEntry == null || satisfiesRequirements(matchProvisions, queueEntry.getPlatform())) {
+
+    DequeueMatchSettings settings = new DequeueMatchSettings();
```
jacobmou, line 316: Every time `workerContext.match()` is called for a new operation, this setting will be created again and in the end, it seems to be a couple of hardcoded booleans, which won't be changed during the lifetime of the worker. Maybe we can make this either a member of `ShardWorkerContext` or a singleton with a factory method so that we dont need to create it repeatedly?

src/test/java/build/buildfarm/worker/DequeueMatchEvaluatorTest.java
```diff
@@ -0,0 +1,300 @@
...
+    assertThat(shouldKeep).isTrue();
+  }
+
+  // Function under test: shouldKeepOperation
+  // Reason for testing: the entry should be kept because the min cores are valid for the worker
+  // properties
+  // Failure explanation: either the property names changed or we evaluate these properties
+  // differently
+  @Test
+  public void shouldKeepOperationValidEqualMinCoresQueueEntry() throws Exception {
```
jacobmou, line 128: I feel this test can be merged with the one above `shouldKeepOperationValidMinCoresQueueEntry`, the QueueEntry settings are identical, the only difference is the worker provision settings.

I also see similar patterns in the following test functions. I feel some of them can be easily merged to make this file much shorter and possibly easier to read.

jacobmou (approved):
Looks good!

Just have some thoughts when reading through the code. Let me know your thought and I'm happy to unlock this.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
luxe: Yeah, we'll need bubble this up all the way to a proto config (ideally).  I'll do this next so we're not creating it in every match call

src/test/java/build/buildfarm/worker/DequeueMatchEvaluatorTest.java
luxe: thanks!  I was able to remove 3 similar tests and condense

luxe:
Thanks for review.  I'm going to follow up with moving the match settings higher into the ShardWorkerContext

PR #613, opened 2020-12-03
Allow individual env vars on exec_properties

luxe:
### problem:
Currently buildfarm allows specifying environment variables through `exec_properties`.   
We made this available because `--action_env` was limited in bazel.  
Given this existing feature there are still some rough edges on the client side.  

### discussion:
See this discussion point about existing functionality and the suggested alternative:
> I can see that this is constrained by exec_properties needing to be a dictionary that maps a string key to a string value. A downside of the current approach is that it does not permit using --remote_default_exec_properties to build up a set of environment variables one at a time. Nor does it facilitate the aggregation of the set of the environment variables from a variety of sources, since the most recent value of env-vars overrides any previous value. Adding a new environment variable to a previous setting would be a bit awkward, as you'd first have to convert the string to a dictionary, then update the dictionary with the new environment variable name and value, and then render the dictionary back into a string. This is possible within Starlark, but not from the command line. And even within Starlark, the code would not have access to a default setting of env-vars provided by --remote_default_exec_properties.
> 
> Another way to do this would be to have exec_properties keys indicate the environment variable names. For instance:
> 
> validated_exec_properties["env-var:MKL_NUM_THREADS"] = "{{limits.cpu.claimed}}"
> validated_exec_properties["env-var:OMP_NUM_THREADS"] = "{{limits.cpu.claimed}}"
> In this format, we could specify the equivalent on the command line:
> 
> --remote_default_exec_properties=env-var:MKL_NUM_THREADS={{limits.cpu.claimed}}
> --remote_default_exec_properties=env-var:OMP_NUM_THREADS={{limits.cpu.claimed}}
> without these being overwritten by the setting of any other environment variables elsewhere.
> This approach, via .bazelrc, would also have the same scope as the default setting of max-cores=1. Specifying the environment variables via the atg wrappers, as in this PR, only covers the code that uses these wrappers, so it will be a subset of all the places that currently get the max-cores=1 setting. I think it would be better to ensure that every place that sees max-cores=1 also gets these environment variable settings.
> 
> This approach would, of course, require buildfarm to scan all the exec_properties keys in order to identify all the environment variables to set. This approach would also eliminate a possible failure mode of supplying a value that did not happen to be a correctly formatted string representing a dictionary.

### solution:
We add this alternative strategy for specifying environment variables.
We will consider removing the other strategy afterward.

### DIFF
--- a/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
+++ b/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
@@ -151,8 +151,8 @@ public class ResourceDecider {
             (property) -> {
               if (property.getName().startsWith(EXEC_PROPERTY_ENV_VAR)) {
                 String keyValue[] = property.getName().split(":", 2);
-                String key = keyValue[0];
-                String value = (keyValue.length > 1) ? keyValue[1] : "";
+                String key = keyValue[1];
+                String value = property.getValue();
                 limits.extraEnvironmentVariables.put(key, value);
               }
             });
--- a/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
+++ b/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
@@ -55,6 +55,14 @@ public class ResourceDecider {
   private static final String EXEC_PROPERTY_ENV_VARS = "env-vars";
 
   ///
+  /// @field   EXEC_PROPERTY_ENV_VAR
+  /// @brief   The exec_property and platform property prefix name for
+  ///          providing an additional environment variable.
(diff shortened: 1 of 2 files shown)


jacobmou:
Is there any test you have done to confirm that the environment variables set through "exec_properties" are correctly parsed and presented here?

jacobmou (approved):

luxe:
> Is there any test you have done to confirm that the environment variables set through "exec_properties" are correctly parsed and presented here?

Apart from those unit tests and some local testing, we have sh_tests in the internal repo that will exercise this.  I'll send you a link as to not expose any urls here

PR #615, opened 2020-12-04
Implement AdminService::DeregisterWorker

luxe:
This provides an API for the Admin Service to deregister a worker from the shard.  
It implements the "Deregister from Redis" box:
Buildfarm Worker Shutdowns

### DIFF
--- a/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
+++ b/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
@@ -39,6 +39,20 @@ service Admin {
   rpc ReindexCas(ReindexCasRequest) returns (ReindexCasRequestResults) {
     option (google.api.http) = { get: "/v1test/admin:reindexCas" };
   }
+  
+  rpc DeregisterWorker(DeregisterWorkerRequest) returns (DeregisterWorkerRequestResults) {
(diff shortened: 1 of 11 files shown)


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -39,6 +39,20 @@ service Admin {
...
+  rpc DeregisterWorker(DeregisterWorkerRequest) returns (DeregisterWorkerRequestResults) {
+    option (google.api.http) = { get: "/v1test/admin:deregisterWorker" };
+  }
+}
+
+message DeregisterWorkerRequest {
+  string instance_name = 1;
+  
+  // format: ip:port
+  string host_id = 2;
```
80degreeswest, line 52: We are mixing hostName and hostId here. Let's keep it consistent. Also, since we are storing worker name in Redis, it may be good to have it be workerName as it comprises of hosts:port as the name here and not really a true host ID or host name.

80degreeswest (approved):

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: changed `host_id` to `worker_name`

PR #595, opened 2020-11-20
allow setting env vars through exec_properties

luxe:
### Environment Variables:
Users need to set additional environment variables through `exec_properties`.  
Changing code / using `--action_env` may be less feasible than specifying them through these exec_properties.  
Additionally, the values of their environment variables may need to be influenced by buildfarm decisions.  

For example,  pytorch tests can still see the underlying hardware through `/proc/cpuinfo`.  
Despite being given 1 core, they see all of the cpus and decide to spawn that many threads. This essentially starves them and gives poor test performance (we may spoof cpuinfo in the future).  Another solution is to use env vars `OMP_NUM_THREADS` and `MKL_NUM_THREADS`.  This could be done in code, but we can't trust that developers will do it consistently or keep it in sync with `min-cores` / `max-cores`.  Allowing these environment variables to be passed the same way as the core settings would be ideal.  

### Standard Example:
This test will succeed when env var TESTVAR is foobar, and fail otherwise.
```
#!/bin/bash
[ "$TESTVAR" = "foobar" ]
```
```
./bazel test  \
--remote_executor=grpc://127.0.0.1:8980 --noremote_accept_cached  --nocache_test_results \
//code/tools/bad_tests/env_test:main
FAIL
```

```
./bazel test --remote_default_exec_properties='env-vars={"TESTVAR": "foobar"}' \
 --remote_executor=grpc://127.0.0.1:8980 --noremote_accept_cached  --nocache_test_results \
//code/tools/bad_tests/env_test:main
PASS
```
### Template Example:
If you give a range of cores, buildfarm has the authority to decide how many your operation actually claims.  
You can let buildfarm resolve this value for you (via [mustache](https://mustache.github.io/)).  
```
#!/bin/bash
[ "$MKL_NUM_THREADS" = "1" ]
```
```
./bazel test  \
--remote_executor=grpc://127.0.0.1:8980 --noremote_accept_cached  --nocache_test_results \
//code/tools/bad_tests/env_test:main
FAIL
```
```
./bazel test  \
--remote_default_exec_properties='env-vars="MKL_NUM_THREADS": "{{limits.cpu.claimed}}"' \
--remote_executor=grpc://127.0.0.1:8980 --noremote_accept_cached  --nocache_test_results //code/tools/bad_tests/env_test:main
PASS
```

### Documentation:
 https://github.com/bazelbuild/bazel-buildfarm/wiki/Execution-Properties

PR #617, opened 2020-12-06
 initialize defaults for ResourceLimits to avoid NPE

luxe:
This is a bugfix for the memory instance introducing during https://github.com/bazelbuild/bazel-buildfarm/pull/595#issuecomment-739483991

Data members should be initialized before they are used.  Using data members without initializing them will result in a NPE.  This problem did not occur in the shard instance because we do initialize and construct all of the members.  This is also fixed and further improved in https://github.com/bazelbuild/bazel-buildfarm/pull/614.  

Reflecting on issue:  
I've been bitten by this a few times now.  Coming from C++ and other languages that default initialize your objects, I keep forgetting that in Java, initializing must be explicit in class definitions. I'll see about a linter for this, and the memory instance is clearly lacking unit tests that would have caught this, so we'll fix that too.

### DIFF
--- a/src/main/java/build/buildfarm/worker/operationqueue/Worker.java
+++ b/src/main/java/build/buildfarm/worker/operationqueue/Worker.java
@@ -706,7 +706,7 @@ public class Worker extends LoggingMain {
 
           @Override
           public ResourceLimits commandExecutionSettings(Command command) {
-            return null;
+            return new ResourceLimits();
           }
         };
 
--- a/src/main/java/build/buildfarm/worker/resources/CpuLimits.java
+++ b/src/main/java/build/buildfarm/worker/resources/CpuLimits.java
@@ -39,26 +39,26 @@ public class CpuLimits {
   /// @details Depending on the server implementation, we may skip applying any
   ///          restrictions to core usage.
   ///
-  public boolean limit;
+  public boolean limit = true;
 
   ///
   /// @field   min
   /// @brief   The minimum CPU cores required.
   /// @details Client can suggest this though exec_properties.
   ///
-  public int min;
+  public int min = 1;
 
   ///
   /// @field   max
   /// @brief   The maximum CPU cores required.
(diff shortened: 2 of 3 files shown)


werkt:
> I've been bitten by this a few times now. Coming from C++ and other languages that default initialize your objects, I keep forgetting that in Java, initializing must be explicit in class definitions.

This is why we use `final` and accessors, something which i recommend you switch these to. Everytime you create a non-final member, think about what that means for nullness

luxe:
> > I've been bitten by this a few times now. Coming from C++ and other languages that default initialize your objects, I keep forgetting that in Java, initializing must be explicit in class definitions.
> 
> This is why we use `final` and accessors, something which i recommend you switch these to. Everytime you create a non-final member, think about what that means for nullness

At least for now, I just want PODs that don't crash when you use them. 

I'll consider using final to avoid it.  It would mean adding constructors and then using some kind of builder pattern when you want to change particular data mems. I could do getters/setters as well but they don't directly solve the NPE issue.  Either include a constructor so the user has to initialize, tag them final so the compiler makes you initialize, or remember to default initialize.

werkt:
> > > I've been bitten by this a few times now. Coming from C++ and other languages that default initialize your objects, I keep forgetting that in Java, initializing must be explicit in class definitions.
> > 
> > 
> > This is why we use `final` and accessors, something which i recommend you switch these to. Everytime you create a non-final member, think about what that means for nullness
> 
> At least for now, I just want PODs that don't crash when you use them.
> 
> I'll consider using final to avoid it. It would mean adding constructors and then using some kind of builder pattern when you want to change particular data mems. I could do getters/setters as well but they don't directly solve the NPE issue. Either include a constructor so the user has to initialize, tag them final so the compiler makes you initialize, or remember to default initialize.

If you're making PODs, just use protobuf messages.

jacobmou (approved):

PR #624, opened 2020-12-10
[WorkerGracefullyShutdown implementation] make worker support gracefully shutdown behavior

jacobmou:
The pr implements another piece of worker gracefully shutdown, which corresponds to `Finish Executions` and `Disable EC2 Scale Protection`. 

Worker Gracefully Shutdown essentially means that an on-demand worker should finish all of its ongoing actions, including actions on any stages, before it gets scaled in. To achieve that, all worker instances/hosts will have the scale in protection when they are started. When a worker host/instnce is chosen to be scaled in, its protection will be disabled when it's confirmed the worker's pipeline is empty.

workergracefullyshutdown

### DIFF
--- a/src/main/java/build/buildfarm/admin/Admin.java
+++ b/src/main/java/build/buildfarm/admin/Admin.java
@@ -29,4 +29,6 @@ public interface Admin {
       Integer maxHosts,
       Integer targetHosts,
       Integer targetReservedHostsPercent);
+
+  void disableHostProtection(String scaleGroupName, String hostId);
 }
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -239,7 +239,6 @@ public class Worker extends LoggingMain {
       return false;
     }
     logger.log(Level.INFO, "The worker Pipeline is empty now!");
-    // turn off auto scaling protection here
     return true;
   }
 
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -112,6 +112,8 @@ public class Worker extends LoggingMain {
   private static final int shutdownWaitTimeInSeconds = 10;
   private final boolean isCasShard;
 
+  private boolean isDeregistered = false;
+
   private final ShardWorkerConfig config;
   private final ShardWorkerInstance instance;
   private final HealthStatusManager healthStatusManager;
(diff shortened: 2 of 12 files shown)


src/main/java/build/buildfarm/admin/Admin.java
```diff
@@ -29,4 +29,6 @@ void scaleCluster(
       Integer maxHosts,
       Integer targetHosts,
       Integer targetReservedHostsPercent);
+
+  void disableHostProtection(String scaleGroupName, String hostId);
```
luxe, line 33: should there be an API to enable host protection?  How does enabling happen?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -1016,3 +1016,18 @@ service WorkerProfile {
...
+}
+
+message DrainWorkerPipelineResults {
+  bool is_pipeline_empty = 1;
+}
+
+
+service ShutDownWorkerGracefully {
+  rpc DrainWorkerPipeline(DrainWorkerPipelineRequest) returns (DrainWorkerPipelineResults) {}
+}
```
luxe, line 1033: Is there a reason this is a separate service instead of part of the AdminService?

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -790,7 +810,7 @@ long nextInterval(long now) {
 
               void registerIfExpired() {
                 long now = System.currentTimeMillis();
-                if (now >= workerRegistrationExpiresAt) {
+                if (now >= workerRegistrationExpiresAt && !isDeregistered) {
```
luxe, line 813: oh interesting.  `workerRegistrationExpiresAt`.  I wonder if that means all workers have an eventual expiration time?

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -224,6 +226,23 @@ public Worker(String session, ShardWorkerConfig config) throws ConfigurationExce
...
+    try {
+      while (!pipeline.isEmpty()) {
+        SECONDS.sleep(10);
+      }
+    } catch (InterruptedException e) {
+      logger.log(Level.SEVERE, "The worker gracefully shutdown is interrupted: " + e.getMessage());
+      return false;
+    }
+    logger.log(Level.INFO, "The worker Pipeline is empty now!");
+    // turn off auto scaling protection here
```
luxe, line 242: do we need to call `disableHostProtection` here?

luxe (approved):
We'll probably learn the most by trying everything out with buildfarm admin.  Feel free to land if you think we are ready to move onto those experiments.

src/main/java/build/buildfarm/admin/Admin.java
jacobmou: I remember Yuriy mentioned that all hosts/instances will have it enabled as default when they are started.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
jacobmou: yeah, that's what i want to do in the beginning.
The grpc call should be invoked from bf-admin or some other place(client side) and worker should be server side, which get informed and stop reregistering itself and finally give a response after everything is done. 

As far as I know, AdminService is actually the grpc server that hooked to bf scheduler right? so it cannot really inform a worker that it should stop reregistering itself?

src/main/java/build/buildfarm/worker/shard/Worker.java
jacobmou: Yes, that the i thing i plan to sync with you and Yuriy later.

Worker registration has a short lifespan, that's why it has to reregister itself in a loop as long as the worker is alive. Another benefit of it, besides the ones we can think of the other day, is that we don't need to deregister a worker when it gets killed because of errors or scale in. That means we do need a separate mechanism to inform a worker that it should stop reregistering itself.

src/main/java/build/buildfarm/worker/shard/Worker.java
jacobmou: oh, my bad, forget to delete the comments. Yeah, that's also my initial thoughts like worker can disable the protection through a grpc call to admin service. But actually, we don't need that because whoever makes the grpc call to inform worker that it should be draining all work from its pipeline and get ready for shutdown will receive the signal and can make call to disable the protection.

PR #625, opened 2020-12-10
Revert "[WorkerGracefullyShutdown implementation] make worker support gracefully shutdown behavior"

jacobmou:
Reverts bazelbuild/bazel-buildfarm#624


80degreeswest (approved):

werkt:
Please provide some clarity on what happened here.

jacobmou:
> Please provide some clarity on what happened here.

Simple version: we merged something and then later decided that it's not what we want and we reverted it.

More Details:
We're working on the worker gracefully shutdown feature. What i did in the reverted pr are: 
a) add new a grpc service to worker so that we can inform a worker that it's chosen for gracefully shutdown and should stop re-registering to redis. Then send a response to the caller after all the on-going actions are done. 
b) add a new function that can disable autoscaling protection of an instance, which is supposed to be called after we know that no actions are left in the worker pipeline stages.

We decided to revert it because we want to make some improvements: 
1) we decide not to use a grpc service to do a), but introducing a separate worker set, say `workers_in_shutdown` in redis, which should contain workers that are chosen for gracefully shutdown. a worker should learn that information by checking that if it is in the new set. A worker should stop registering itself to the regular worker set, but to `workers_in_shutdown`.

2) after executing all actions, the worker should make a grpc call to admin-service to disable its autoscaling protection.

What do you think about our new plan? It'll leverage a new set in redis and need less grpc call. Does it make sense?

PR #626, opened 2020-12-10
Jacobmou/worker gracefully

jacobmou:
Worker gracefully shutdown feature with 2 stops. 
In the updated design, 
1) bfadmin will start by asking redis to return a random worker and then request the worker to gracefully shut down through grpc call. 
2) The worker, once received the request, should stop registering itself and monitor all ongoing actions in its pipeline. The worker will inform AdminService hooked to the bf-server when all the actions are done. 

3) The AdminService is responsible to disable the autoscaling protection of the instance, which is turned on by default.

Screen Shot 2021-01-04 at 3 28 32 PM

The pr implements 2) and 3). The "choosing a worker randomly" is already implemented in separate pr.

### DIFF
--- a/src/main/java/build/buildfarm/admin/Admin.java
+++ b/src/main/java/build/buildfarm/admin/Admin.java
@@ -29,4 +29,6 @@ public interface Admin {
       Integer maxHosts,
       Integer targetHosts,
       Integer targetReservedHostsPercent);
+
+  void disableHostProtection(String scaleGroupName, String hostId);
 }
--- a/src/main/java/build/buildfarm/admin/Admin.java
+++ b/src/main/java/build/buildfarm/admin/Admin.java
@@ -30,5 +30,5 @@ public interface Admin {
       Integer targetHosts,
       Integer targetReservedHostsPercent);
 
-  void disableHostProtection(String scaleGroupName, String hostId);
+  void disableHostScaleInProtection(String instanceName);
 }
--- a/src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
+++ b/src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
@@ -155,6 +155,12 @@ public class AwsAdmin implements Admin {
   @Override
   public void disableHostScaleInProtection(String instanceName) {
 
+    // 1 get instance id from instance private ip
+    AmazonEC2ClientBuilder.standard().
+    // 2 get autoscaling group name from instance id
+    // https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/autoscaling/AmazonAutoScaling.html#putScalingPolicy-com.amazonaws.services.autoscaling.model.PutScalingPolicyRequest-
+    // describeAutoScalingInstances
+
     //logger.log(
     //    Level.INFO,
     //    String.format(
--- a/src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
+++ b/src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
@@ -19,7 +19,6 @@ import build.buildfarm.v1test.GetHostsResult;
 import build.buildfarm.v1test.Host;
 import com.amazonaws.services.autoscaling.AmazonAutoScaling;
 import com.amazonaws.services.autoscaling.AmazonAutoScalingClientBuilder;
-import com.amazonaws.services.autoscaling.model.AutoScalingGroup;
 import com.amazonaws.services.autoscaling.model.InstancesDistribution;
 import com.amazonaws.services.autoscaling.model.MixedInstancesPolicy;
 import com.amazonaws.services.autoscaling.model.SetInstanceProtectionRequest;
@@ -32,6 +31,7 @@ import com.amazonaws.services.ec2.model.DescribeInstancesResult;
 import com.amazonaws.services.ec2.model.Filter;
 import com.amazonaws.services.ec2.model.Instance;
 import com.amazonaws.services.ec2.model.Reservation;
+import com.amazonaws.services.ec2.model.Tag;
 import com.amazonaws.services.ec2.model.TerminateInstancesRequest;
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement;
 import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementClientBuilder;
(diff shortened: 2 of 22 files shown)


src/main/java/build/buildfarm/admin/Admin.java
```diff
@@ -29,4 +29,6 @@ void scaleCluster(
       Integer maxHosts,
       Integer targetHosts,
       Integer targetReservedHostsPercent);
+
+  void disableHostProtection(String scaleGroupName, String hostId);
```
luxe, line 33: can you remind me how host protection gets enabled?  

I think all the instances have host protection automatically by default from aws?  And our only use case is to turn it off when shutting down a worker.  

Is host protection applied to schedulers or only workers?

src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
```diff
@@ -148,4 +150,19 @@ private long getHostUptimeInMinutes(Date launchTime) {
...
+        new SetInstanceProtectionRequest()
+            .withInstanceIds(hostId)
+            .withAutoScalingGroupName(scaleGroupName)
+            .withProtectedFromScaleIn(false);
+    SetInstanceProtectionResult response = scale.setInstanceProtection(disableRequest);
+    logger.log(
+        Level.INFO,
+        String.format(
+            "Disable protection of host: %s in AutoScalingGroup: %s and get result: %s",
+            hostId, scaleGroupName, response.toString()));
```
luxe, line 166: we could check the response and log the error state with Level.SEVERE so its easier to find in cloudwatch, but we can do that later if needed.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -853,4 +867,12 @@ public void deregisterWorker(String workerName) {
             .deregisterWorker(
                 DeregisterWorkerRequest.newBuilder().setWorkerName(workerName).build());
   }
+
+  @Override
+  public DrainWorkerPipelineResults drainWorkerPipeline(String worker) {
+    throwIfStopped();
+    return shutDownWorkerGracefullyBlockingStub
+        .get()
+        .drainWorkerPipeline(DrainWorkerPipelineRequest.newBuilder().build());
```
luxe, line 876: I think `worker` needs added the the `DrainWorkerPipelineRequest`

src/main/java/build/buildfarm/server/AdminService.java
```diff
@@ -180,6 +182,14 @@ public void deregisterWorker(
     }
   }
 
+  @Override
+  public void disableScaleInProtection(
+      DisableScaleInProtectionRequest request,
+      StreamObserver<DisableScaleInProtectionRequestResults> responseObserver) {
+    // adminController.disableHostProtection(instance_id, asg_name);
```
luxe, line 189: "host protection" vs "scale in" protection? Not sure what this is

jacobmou: aws call it scale in protection, because that's what it is. Host protection sounds like a more general protection, but I don't exactly what it is.

src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
jacobmou: In the current implementation, this message will be not available in the grpc response. Worker will just send back response that it received the gracefully shutdown request and it wont wait until it finishes all actions, because they may take more than a couple of minutes. The worker will make a separate grpc call to AdminService that it's ready for losing protection. 

But, yeah, i just talked about this with Yuriy, we can make the caller wait for the response that including whether the worker is ready for shutdown, but that means AdminService is more coupled with workers. So we are not super sure yet.

src/main/java/build/buildfarm/admin/Admin.java
jacobmou: Yes, all worker hosts will have the protection on default when we roll out the graceful shutdown. I'm not super sure the are the default on was.

Scale in protection is an aws instance feature, so it also applies to scheduler instances. But i don't remember we have any plan to turn it on for schedulers.

src/main/java/build/buildfarm/server/AdminService.java
jacobmou: Yeah, i will make the names more consistent here

src/main/java/build/buildfarm/instance/stub/StubInstance.java
jacobmou: The worker ip:port is used to create the ManagedChannel, so we are reaching that worker directly. Do you think we should do something with the `worker`?

luxe:
So every worker has a "ShutDownWorkerGracefully" service, and bf-admin will contact these services directly?  I think this all makes sense.  Best thing to do is start experimenting in a dev environment with bf-admin

luxe (approved):

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -233,10 +244,14 @@ message LogMetricsConfig {
...
   string deployment_environment = 1;
 
+  string cluster_id = 4;
+
   oneof type {
     AwsAdminConfig aws_admin_config = 2;
     GcpAdminConfig gcp_admin_config = 3;
   }
+
+  bool enable_graceful_shutdown = 5;
```
jlisee, line 254: Is this flag in use here, I assume it would be used like `clusterId` is used in `Worker.java` above, right?

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -224,6 +227,61 @@ public Worker(String session, ShardWorkerConfig config) throws ConfigurationExce
...
+    try {
+      while (!pipeline.isEmpty() && timeWaited < timeOut) {
+        SECONDS.sleep(scanRate);
+        timeWaited += scanRate;
+      }
+    } catch (InterruptedException e) {
+      logger.log(Level.SEVERE, "The worker gracefully shutdown is interrupted: " + e.getMessage());
+    } finally {
+      // make a grpc call to disable scale protection
+      String clusterId = config.getAdminConfig().getClusterId();
```
80degreeswest, line 251: We should check for clusterId first here, if it doesn't exist no point in trying to finish work.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
80degreeswest, line 254: I don't see this being checked to make sure it's enabled?

80degreeswest (changes requested):

src/main/java/build/buildfarm/worker/shard/ShutDownWorkerGracefully.java
```diff
@@ -37,6 +46,20 @@ public ShutDownWorkerGracefully(Worker worker) {
...
+    String clusterId = config.getAdminConfig().getClusterId();
+    if (clusterId == null
+        || clusterId.equals("")
+        || !config.getAdminConfig().getEnableGracefulShutdown()) {
+      String errorMessage =
+          String.format(
+              "Current AdminConfig doesn't have cluster_id set or doesn't support shut down worker gracefully, "
+                  + "the worker %s won't be shut down.",
+              config.getPublicName());
+      logger.log(SEVERE, errorMessage);
```
80degreeswest, line 58: This can be a WARNING instead.

80degreeswest (approved):

PR #627, opened 2020-12-11
Support SymlinkNode in input directories

werkt:
Refactor OperationQueueWorkerContext (finally) and improve the stat
logic around directory scanning.

### DIFF
--- a/src/test/java/build/buildfarm/common/io/UtilsTest.java
+++ b/src/test/java/build/buildfarm/common/io/UtilsTest.java
@@ -54,8 +54,8 @@ class UtilsTest {
 
     List<NamedFileKey> files = Utils.listDirentSorted(root, fileStore);
     assertThat(files.size()).isEqualTo(2);
-    Object firstKey = files.get(0).fileKey();
-    Object secondKey = files.get(1).fileKey();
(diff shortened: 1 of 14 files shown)


src/test/java/build/buildfarm/common/io/UtilsTest.java
```diff
@@ -71,8 +71,8 @@ public void fileKeysVerifyDifferentFiles() throws IOException {
 
     List<NamedFileKey> files = Utils.listDirentSorted(root, fileStore);
     assertThat(files.size()).isEqualTo(2);
-    Object firstKey = files.get(0).fileKey();
-    Object secondKey = files.get(1).fileKey();
+    Object firstKey = files.get(0).getFileKey();
+    Object secondKey = files.get(1).getFileKey();
     assertThat(firstKey).isNotEqualTo(secondKey);
```
luxe, line 76: these tests won't run in CI until public is added to `class UtilsTest`

luxe (approved):
Looks like a nice refactor and it should be low risk for us since we don't use the in-memory worker.  This also adds support SymlinkNode which is not something I think we'll see with bazel.  Nothing stands our in the file io changes.

PR #629, opened 2020-12-14
Fix server hanging during startup on Redis error

### LINKED ISSUE #623: Build farm server stuck on Redis errors
We are observing server processes stuck with traces similar to the following:
failure-wait-signal
After such errors, the server process exists only when getting a signal.

These error can be reproduced locally easily by throwing an exception (RuntimeException or IOException in RedisShardSubscription::iterate:
https://github.com/bazelbuild/bazel-buildfarm/blob/master/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java#L69

I am not sure how to have the process terminate or recover when such errors occur. Do you have any advice?

jensgreen:
Fixes #623.

---

There is a race condition in BuildFarmServer.start(): if redis fails
before BuildFarmServer.start() has finished, a callback will end up
triggering BuildFarmServer.stop(), causing the server to hang.

In BuildFarmServer.start(), [instances.start()][1] starts a new thread.

If [RedisShardSubscription.iterate()][2] throws an IOException, the
[onUnsubscribe callback][3] will trigger. In the end, this calls
[BuildFarmServer.stop()][4]. If stop() is called before start() has
finished, the server hangs.

A fix is to make start() synchronized.

[1]: https://github.com/bazelbuild/bazel-buildfarm/blob/4104ed9c97e9cea0b1e5125e3511146de00fc038/src/main/java/build/buildfarm/server/BuildFarmServer.java#L141
[2]: https://github.com/bazelbuild/bazel-buildfarm/blob/4104ed9c97e9cea0b1e5125e3511146de00fc038/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java#L69-L84
[3]: https://github.com/bazelbuild/bazel-buildfarm/blob/4104ed9c97e9cea0b1e5125e3511146de00fc038/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java#L110
[4]: https://github.com/bazelbuild/bazel-buildfarm/blob/4104ed9c97e9cea0b1e5125e3511146de00fc038/src/main/java/build/buildfarm/server/BuildFarmServer.java#L80-L81

### DIFF
--- a/src/main/java/build/buildfarm/server/BuildFarmServer.java
+++ b/src/main/java/build/buildfarm/server/BuildFarmServer.java
@@ -136,7 +136,11 @@ public class BuildFarmServer extends LoggingMain {
     }
   }
 
-  public void start(String publicName) throws IOException {
+  public synchronized void start(String publicName) throws IOException {
+    if (stopping) {
+      logger.log(Level.WARNING, "already stopping, skipping start");
+      return;
+    }
     actionCacheRequestCounter.start();
     instances.start(publicName);
     server.start();
--- a/src/main/java/build/buildfarm/server/BuildFarmServer.java
+++ b/src/main/java/build/buildfarm/server/BuildFarmServer.java
@@ -15,6 +15,7 @@
 package build.buildfarm.server;
 
 import static build.buildfarm.common.io.Utils.formatIOError;
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.util.concurrent.MoreExecutors.shutdownAndAwaitTermination;
 import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
 import static java.util.logging.Level.SEVERE;
@@ -137,10 +138,7 @@ public class BuildFarmServer extends LoggingMain {
   }
 
   public synchronized void start(String publicName) throws IOException {
-    if (stopping) {
-      logger.log(Level.WARNING, "already stopping, skipping start");
-      return;
-    }
+    checkState(!stopping, "must not call start after stop");
     actionCacheRequestCounter.start();
     instances.start(publicName);
     server.start();


jensgreen:
Here is a repro:


https://github.com/jensgreen/bazel-buildfarm/commit/5fd80e1da2c75f4ba89c8e53609116a051849a26

```diff
From 5fd80e1da2c75f4ba89c8e53609116a051849a26 Mon Sep 17 00:00:00 2001
From: Jens Green Olander <jgreen@spotify.com>
Date: Mon, 14 Dec 2020 10:54:34 +0100
Subject: [PATCH] Reproduce server hang on startup when redis fails

---
 .../buildfarm/instance/shard/RedisShardSubscription.java      | 3 +++
 src/main/java/build/buildfarm/server/BuildFarmServer.java     | 4 +++-
 2 files changed, 6 insertions(+), 1 deletion(-)

diff --git src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
index 6bf3e4a1..34b50f7a 100644
--- src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
+++ src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
@@ -66,10 +66,13 @@ class RedisShardSubscription implements Runnable {
     jedis.subscribe(subscriber, subscriptions.get().toArray(new String[0]));
   }
 
   private void iterate(boolean isReset) throws IOException {
     try {
+      if (1 - 1 == 0) {
+        throw new IOException("boom!");
+      }
       client.run(jedis -> subscribe(jedis, isReset));
     } catch (IOException e) {
       Status status = Status.fromThrowable(e);
       switch (status.getCode()) {
         case DEADLINE_EXCEEDED:
diff --git src/main/java/build/buildfarm/server/BuildFarmServer.java src/main/java/build/buildfarm/server/BuildFarmServer.java
index f1f22f0b..ebcdaa12 100644
--- src/main/java/build/buildfarm/server/BuildFarmServer.java
+++ src/main/java/build/buildfarm/server/BuildFarmServer.java
@@ -134,13 +134,15 @@ public class BuildFarmServer extends LoggingMain {
       case "gcp":
         return new GcpMetricsPublisher(metricsConfig);
     }
   }
 
-  public void start(String publicName) throws IOException {
+  // NOTE! Make start synchronized to make the server shut down instead of hang.
+  public void start(String publicName) throws IOException, InterruptedException {
     actionCacheRequestCounter.start();
     instances.start(publicName);
+    Thread.sleep(3000);
     server.start();
     healthStatusManager.setStatus(
         HealthStatusManager.SERVICE_NAME_ALL_SERVICES, ServingStatus.SERVING);
   }
 
-- 
2.25.0


```

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -136,7 +136,11 @@ private static MetricsPublisher getMetricsPublisher(MetricsConfig metricsConfig)
     }
   }
 
-  public void start(String publicName) throws IOException {
+  public synchronized void start(String publicName) throws IOException {
+    if (stopping) {
```
werkt, line 140: This should just be an invalid state - start should not be callable in any context where stopping was able to be set: prefer guava's `checkState` here.

werkt (changes requested):

src/main/java/build/buildfarm/server/BuildFarmServer.java
jensgreen: Good point, will fix!

werkt (approved):

PR #631, opened 2020-12-17
make actionCache an explicit object in backplane

luxe:
The prequeue/queue appear in the backplane as explicit containers.  This was done to make tracing code and extending the queue functionality easier.  We now want to do the same thing for the action cache.  

We create a map abstraction to represent the key/value format of the action cache and re-integrate it into the backplane.
This data structure will be later used for similar patterns found in the backplane.

**from internal convo:**
By "container", I mean the java class `RedisMap`.
By "object" I mean a named instantiation of this map: `RedisMap actionCache = newRedisMap("name");`

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisMap.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisMap.java
@@ -83,6 +83,17 @@ public class RedisMap {
     p.sync();
   }
   ///
+  /// @brief   Get the value of the key.
+  /// @details If the key does not exist, null is returned.
+  /// @param   jedis Jedis cluster client.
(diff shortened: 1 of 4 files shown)


werkt (approved):
Interesting. I like it. Not sure what the gain is, but it seems like worthwhile compartmentailization.

luxe:
> Interesting. I like it. Not sure what the gain is, but it seems like worthwhile compartmentailization.

that's right, its not much of a gain.  I'd like to glance at the data members in the backplane or the constructor and immediately know the data involved in the shard. seeing the prequeue, the queue, an action cache, a cas lookup, a worker list, an operation blacklist, etc etc

PR #632, opened 2020-12-17
backplane: choosing workers for shutdown

luxe:
This is added for our "graceful worker shutdown" functionality.  A scale down service will decide how many workers it wants to shutdown and ask the backplane to provide those worker names.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -92,6 +92,8 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.naming.ConfigurationException;
 import redis.clients.jedis.JedisCluster;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -713,6 +715,28 @@ public void deregisterWorker(String workerName) throws IOException {
     return workerSet;
   }
 
+  // When performing a graceful scale down of workers, the backplane can provide worker names to the
+  // scale-down service. The algorithm in which the backplane chooses these workers can be made more
+  // sophisticated in the future. But for now, we'll give back n random workers.
+  public List<String> SuggestedWorkersForShutdown(int amount) throws IOException {
```
80degreeswest, line 721: Should be suggestedWorkersForShutdown

80degreeswest: Also, maybe the method name should be something like suggestedWorkersToScaleDown and amount should be more descriptive, maybe numWorkers.

80degreeswest (approved):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: both suggestions are better. fixed.

PR #634, opened 2021-01-04
configure max cpu for action validation

luxe:
Users should set an appropriate value for `maxCpu`.
```
ShardInstanceConfig {
    max_cpu: 80
}


jacobmou (approved):

werkt:
Except you didn't add this to the examples, and now everyone that tries to run with min-cores: 4 without this set is running into an error.

PR #635, opened 2021-01-06
Update logging

80degreeswest:
Downgrade non-essential logging and update some log messages to reduce overhead.


luxe (approved):
looks good.  might break a honeycomb dashboard or something, but not really a concern.

80degreeswest:
> looks good. might break a honeycomb dashboard or something, but not really a concern.

Yeah it will but we'll be consolidating metrics anyways and we are not using most of those anyways.

PR #641, opened 2021-01-08
add prometheus metrics for AC and CAS lookup containers

luxe:
add size methods for the ActionCache and CasLookup table.  Send those sizes to prometheus for metrics.  I have some concerns about how we calculate the size, as all entries need loaded into memory to avoid counting duplicates across the cluster.  This probably won't scale on higher container sizes.  We can add a scan that doesn't dedup.  It may not be perfectly accurate, but will avoid the high memory usage.


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -173,6 +174,18 @@
       Gauge.build().name("worker_pool_size").help("Active worker pool size.").register();
   private static final Gauge queueSize =
       Gauge.build().name("queue_size").labelNames("queue_name").help("Queue size.").register();
+  private static final Gauge casLookupSize =
```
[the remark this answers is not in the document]
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -173,6 +174,18 @@
...
+      Gauge.build()
+          .name("cas_lookup_size")
+          .labelNames("cas_lookup_size")
+          .help("CAS lookup size.")
+          .register();
+  private static final Gauge actionCacheLookupSize =
+      Gauge.build()
+          .name("action_cache_lookup_size")
+          .labelNames("action_cache_lookup_size")
```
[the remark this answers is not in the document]

src/main/java/build/buildfarm/common/ScanCount.java
```diff
@@ -0,0 +1,104 @@
...
+  /// @param   scanCount The count per scan.
+  ///
+  private static void addKeys(
+      JedisCluster cluster, Jedis node, Set<String> keys, String query, int scanCount) {
+    // iterate over all entries via scanning
+    String cursor = "0";
+    do {
+      keys.addAll(scan(node, cursor, query, scanCount));
+
+    } while (!cursor.equals("0"));
```
jacobmou, line 78: I'm a little confused here about the `cursor` variable in `addKeys(...)` scope. It seems to me `cursor` never changes in this scop and the do-while loop is not necessary.

luxe: The call to `scan(...,cursor,...)` changes the cursor as an output parameter.  


This is inside the `scan` method:  
```
...
ScanResult scanResult = node.scan(cursor, params);
cursor = scanResult.getCursor(); //<- hopefully not zero until we've reached the end
...
```
---
This code was copied from `WorkerIndexer.java` which does a similar thing.  There may be a cleaner way to write this.  I'd like to hide the cursor details completely and do something that looks more like `ForEach/Map` in both this file and WorkerIndexer.

jacobmou: Yeah, i can see `cursor` is intended to be updated inside `scan()`, but `cursor` in `scan()` is just a copy of `addKey()` `cursor` variable. Java just pass it by value.

I did a simple experiment 
```
    private static void stringMutability() {
        String marker = "0";
        changeString(marker);
        System.out.println(marker);
    }

    private static void changeString(String marker) {
        marker = "1";
    }
```
And the printed string is always `0`.

jacobmou: To really achieve what you want, `cursor`could be returned by `scan()`, which is not practical here, or could be a member variable of the class so that its state can be seen from a different scope.

luxe: Oh you're right.  I tested it with too small of data to see the problem.  I'll do what you did in the other diff.

luxe: fixed

80degreeswest (approved):

jacobmou (approved):

PR #643, opened 2021-01-08
Unify documentation comment style

jacobmou:
There are multiple documentation comment styles existing in bf codebase. Per most java code style guides ([Google java documentation style](https://google.github.io/styleguide/javaguide.html#s7-javadoc) and [Oracle java documentation comment convention](https://www.oracle.com/java/technologies/javase/codeconventions-comments.html)), the following style is recommended.
/**
 *
 */
Besides, github can also recognize this comment style and can highlight key information in the documentation comments:
Screen Shot 2021-01-08 at 5 59 51 PM
Screen Shot 2021-01-08 at 6 00 00 PM

I didn't find a java formatter that can enforce this until now. Once we found one, we should enforce this in CI.

### DIFF
--- a/src/main/java/build/buildfarm/common/CasIndexResults.java
+++ b/src/main/java/build/buildfarm/common/CasIndexResults.java
@@ -14,43 +14,43 @@
 
 package build.buildfarm.common;
 
-///
-/// @class   CasIndexResults
-/// @brief   The result of a worker de-registering itself from CAS entries.
-/// @details This will remove the worker from CAS indexes, and in some cases,
-///          remove the CAS key.
-///
+/**
+ * @class   CasIndexResults
+ * @brief   The result of a worker de-registering itself from CAS entries.
+ * @details This will remove the worker from CAS indexes, and in some cases,
+ *          remove the CAS key.
+ */
 public class CasIndexResults {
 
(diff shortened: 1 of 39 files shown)


luxe (approved):
thanks.  this is mostly my fault.  I have an ide that generates them.  They are debatebly too verbose and annoying.  I'm fine with doing anything that makes the most people happy.  And if it looks better on github, that's a positive.  

If you have other requests like not using @field or @class or something, let me know and I will adjust to that.

jacobmou:
> thanks. this is mostly my fault. I have an ide that generates them. They are debatebly too verbose and annoying. I'm fine with doing anything that makes the most people happy. And if it looks better on github, that's a positive.
> 
> If you have other requests like not using @field or @Class or something, let me know and I will adjust to that.

The comments are great! I just try to make the style more consistent. As Yuriy mentioned earlier, we should add verbose comments when necessary. Wondering what IDE you use? I used Eclipse before, but now stick to Intellij most of the time for java, because it has bazel plugin in and makes things pretty easy.

luxe:
i use sublime with a bunch of custom scripts. maybe i'll give the other 2
you mention another shot some time.

PR #645, opened 2021-01-12
fix PipelineStages empty checks

jacobmou:
Using SuperscalarPipelineStage.isClaimed() to check if they are empty turn out to be wrong. This pr will fix that issue.


luxe (approved):
what does isClaimed do?

werkt:
There are some typos in here. I am not sure about this if it pertains to anything other than graceful shutdown

jacobmou:
> There are some typos in here. I am not sure about this if it pertains to anything other than graceful shutdown

Fixed. Yes, it's just for graceful shutdown.

PR #647, opened 2021-01-12
fix issues in graceful shutdown and add separate client to test grpc calls

jacobmou:
Fix the following issues in worker graceful shutdown:
1 handle the port suffixed worker private dns name format in AdminService;
2 prevent worker from matching any operations from the queue after received prepare for graceful shutdown request.

Besides, a separate class with 3 rpc service client sides are added to test their functionality.


luxe (approved):
okay, I see how matching is prevented during shutdown now and that will prevent workers from taking more operations off the queue.  Thanks for adding tests.  Let's try it out in dev.

PR #653, opened 2021-01-20
Fix two shutdown problems

### LINKED ISSUE #651: Have to kill -9 server after JedisClusterException: CLUSTERDOWN The cluster is down
Hi!

We're running d6cec6256ad19405b2be2bf2d18b3f6ed49dc72d in production.

While redeploying our Redis backend, we often see this exception in our logs:
```
Exception in thread "main" redis.clients.jedis.exceptions.JedisClusterException: CLUSTERDOWN The cluster is down
	at redis.clients.jedis.Protocol.processError(Protocol.java:126)
	at redis.clients.jedis.Protocol.process(Protocol.java:166)
	at redis.clients.jedis.Protocol.read(Protocol.java:220)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:389)
	at redis.clients.jedis.Connection.getStatusCodeReply(Connection.java:271)
	at redis.clients.jedis.Jedis.set(Jedis.java:151)
	at redis.clients.jedis.JedisCluster$1.execute(JedisCluster.java:145)
	at redis.clients.jedis.JedisCluster$1.execute(JedisCluster.java:142)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:106)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:27)
	at redis.clients.jedis.JedisCluster.set(JedisCluster.java:147)
	at build.buildfarm.instance.shard.RedisShardBackplane.lambda$start$3(RedisShardBackplane.java:528)
	at build.buildfarm.common.redis.RedisClient.call(RedisClient.java:118)
	at build.buildfarm.instance.shard.RedisShardBackplane.start(RedisShardBackplane.java:527)
	at build.buildfarm.instance.shard.ShardInstance.start(ShardInstance.java:441)
	at build.buildfarm.server.BuildFarmInstances.start(BuildFarmInstances.java:149)
	at build.buildfarm.server.BuildFarmServer.start(BuildFarmServer.java:143)
	at build.buildfarm.server.BuildFarmServer.serverMain(BuildFarmServer.java:228)
	at build.buildfarm.server.BuildFarmServer.main(BuildFarmServer.java:243)
```

After that the process hangs and has to be killed with `kill -9`. CTRL-C does not work, neither does just `kill`.

Having looked at the code, what seems to happen is that when `RedisShardBackplane.start()` tries to write its start time to Redis, it gets this exception.

Then, the exception is caught as a `RuntimeException` in `ShardInstance.start()`, which tries to handle that by calling `ShardInstance.stop()`.

That stop operation ends up waiting for the subscription thread to finish in `RedisShardBackplane.stop()`.

But the subscription thread has called `RedisShardSubscriber.proceed()`, and is now awaiting data on some socket.

At this point, if you try CTRL-C, you get:
```
^C*** shutting down gRPC server since JVM is shutting down
```

But the process never exits, and you have to use `kill -9` to get rid of it.

# Request

In this situation, it would be better if buildfarm (in order of preference):
1. Retried on multiple cluster nodes with a backoff, so that it can recover once Redis is happy again
2. Terminated with an exit code

# Repro

Apply this patch to d6cec6256ad19405b2be2bf2d18b3f6ed49dc72d and start the server:
```diff
diff --git src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
index 36a6715b..c85d9741 100644
--- src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -79,7 +79,6 @@ import java.time.Instant;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -99,6 +98,7 @@ import redis.clients.jedis.JedisClusterPipeline;
 import redis.clients.jedis.Response;
 import redis.clients.jedis.ScanParams;
 import redis.clients.jedis.ScanResult;
+import redis.clients.jedis.exceptions.JedisClusterException;
 
 public class RedisShardBackplane implements ShardBackplane {
   private static final Logger logger = Logger.getLogger(RedisShardBackplane.class.getName());
@@ -524,8 +524,8 @@ public class RedisShardBackplane implements ShardBackplane {
     }
 
     // Record client start time
-    client.call(
-        jedis -> jedis.set("startTime/" + clientPublicName, Long.toString(new Date().getTime())));
+    // client.call(jedis -> jedis.set("startTime/" + clientPublicName, Long.toString(new Date().getTime())));
+    throw new JedisClusterException("CLUSTERDOWN The cluster is down");
   }
 
   static BalancedRedisQueue createPrequeue(RedisClient client, RedisShardBackplaneConfig config)
```

Also, see this thread dump from the frozen state: [frozen-threaddump.txt](https://github.com/bazelbuild/bazel-buildfarm/files/5835923/frozen-threaddump.txt)

walles:
This PR comes with three improvements.

One is to just name two threads to make debugging easier.

One is to bump `commons-pool2` to a version with a fix for "[POOL-348 The commons-pool-evictor-thread should run as a Deamon](https://issues.apache.org/jira/browse/POOL-348)". This is one half of the fix for #651.

One is to avoid an involuntary re-entry into the `ShardInstance` shutdown code. This fixes the other half of #651.

With these changes in place, the repro in #651 results in the Buildfarm server just printing an exception message and terminating with exit code `1`.

Fixes #651


luxe (approved):
> 1. naming threads  

that's good and harmless

> 2. bump commons-pool2    

Running the threads as daemons kind of confuses me.  
But reading through the issue I see no reason we shouldn't keep this dep up-to-date.
Especially if it solves a known issue.

> 3. avoid an involuntary re-entry  

Oh I didn't know the ShardInstance had these stopped/stopping flags.  Its like a makeshift state machine for lifetime management.  
I'll have to research if there is a smarter way to do these things in general.  I'm familiar with the constructor/destructor paradigm of languages like c++ where these states are managed by scope.  But I know that's not the case in Java so I'm not sure best practices. 

--
Looks fine to me.  It feels like an issue where you're already in an exceptional state and the shutdown logic is not correct.  In production, our workers are taken down with the container, and locally I iterate with a script that basically kills everything. so this seems like a good fix for more reliable SIGINT.

walles:
@werkt @80degreeswest @jacobmou could either of you please have a look as well?

jacobmou (approved):
Using `stopping` to flag the ShardInstance state and avoid re-entry seems good enough here! Thanks!

PR #655, opened 2021-01-21
Document setup of local Redis cluster for development

walles:
Following these instructions will get you a Buildfarm server talking to a local Redis cluster of six nodes.

Takes a minute or so.

Good for testing fault tolerance in both Redis and Buildfarm.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,35 @@ bazel run //src/main/java/build/buildfarm:buildfarm-server -- --debug=5005 $PWD/
 
 ## Developer Information
 
+### Setting up Redis for local testing
+
+This is done using [`examples/development-redis-cluster.sh`](examples/development-redis-cluster.sh).
+
+Tested with Redis `6.0.10`, other versions probably work fine as well.
+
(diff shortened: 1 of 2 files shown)


luxe (approved):
Thank you for doing this.  Internally we have a containerized setup for local development. We should land your work first and merge any ideas later.  

In fact, I use our containerized version less frequently for testing buildfarm because its harder to experiment with cc_ like rules without setting up a hermetic toolchain or modifying the existing containers.  So personally, I've been developing with a similar setup you've provided here.

I'm not sure that a minimal redis cluster necessarily needs 6 instances, but its a good size since that's what the tutorial shows.  I run the `./create-cluster` script that comes with the redis repo and it creates 6 instances too.  So maybe it is the case.  We may want to follow-up with something that does all of this in tmux so its closer to a single command setup.

luxe:
We need another stamp to land.  @jacobmou, take a look?

jacobmou (approved):
Looks good! Thank you for adding this!

PR #658, opened 2021-01-26
integrate bazel sandbox

luxe:
Allow buildfarm to use bazel's sandbox for compatibility with locally running actions.  
This organizes the current execution wrappers so they can be used by workers.
Workers use these in their configuration via "execution_wrapper".

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -31,5 +31,8 @@ genrule(
 
 cc_binary(
     name = "as-nobody.binary",
-    srcs = ["as-nobody.c"],
+    srcs = select({
+        "//config:windows": ["as-nobody-windows.c"],
+        "//conditions:default": ["as-nobody.c"],
+    }),
 )
--- a/BUILD
+++ b/BUILD
@@ -3,3 +3,33 @@ load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
 buildifier(
     name = "buildifier",
 )
+
+# These are execution wrappers that buildfarm may choose to use when executing actions.
+# For their availability on a worker, they should be provided to a java_image as a "runtime_dep".
+# The relevant configuration for workers is the "execution policy".
+# That is where these binaries can be used and stacked.
+genrule(
+    name = "process-wrapper.binary",
+    srcs = ["@bazel//src/main/tools:process-wrapper"],
+    outs = ["process-wrapper"],
+    cmd = "cp $< $@;",
+)
+
+genrule(
+    name = "linux-sandbox.binary",
+    srcs = ["@bazel//src/main/tools:linux-sandbox"],
+    outs = ["linux-sandbox"],
+    cmd = "cp $< $@;",
+)
+
+genrule(
+    name = "tini.binary",
+    srcs = ["@tini//file"],
+    outs = ["tini"],
+    cmd = "cp $< $@ && chmod +x $@",
+)
+
+cc_binary(
+    name = "as-nobody.binary",
+    srcs = ["as-nobody.c"],
+)
--- a/BUILD
+++ b/BUILD
@@ -8,6 +8,8 @@ buildifier(
 # For their availability on a worker, they should be provided to a java_image as a "runtime_dep".
 # The relevant configuration for workers is the "execution policy".
 # That is where these binaries can be used and stacked.
+# Be aware that the process-wrapper and linux-sandbox come from bazel itself.
+# Therefore, users may want to ensure that the same bazel version is sourced here as is used locally.
 genrule(
     name = "process-wrapper.binary",
     srcs = ["@bazel//src/main/tools:process-wrapper"],
--- a/as-nobody.c
+++ b/as-nobody.c
@@ -1,0 +1,30 @@
+#include <pwd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <grp.h>
+
+int
+main(int argc, char *argv[])
+{
+  struct passwd *pw = getpwnam("nobody");
+  if (pw == NULL) {
+    perror("getpwnam");
+    return EXIT_FAILURE;
+  }
+  if (setgroups(0, NULL) < 0) {
+    perror("setgroups");
+    return EXIT_FAILURE;
+  }
+  if (setregid(pw->pw_gid, pw->pw_gid) < 0) {
+    perror("setgroups");
+    return EXIT_FAILURE;
+  }
+  if (setreuid(pw->pw_uid, pw->pw_uid) < 0) {
+    perror("setreuid");
+    return EXIT_FAILURE;
+  }
+  execvp(argv[1], argv);
+  perror("execvp");
+  return EXIT_FAILURE;
+}
--- a/deps.bzl
+++ b/deps.bzl
@@ -74,9 +74,9 @@ def archive_dependencies(third_party):
         # Bazel is referenced as a dependency so that buildfarm can access the linux-sandbox as a potential execution wrapper.
         {
             "name": "bazel",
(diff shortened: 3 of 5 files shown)


jacobmou (commented):
Using bazel process wrapper seems to be a good idea, but there are a few caveats i can see here: 
1) if we are using wrappers from a certain version/commit of bazel, there is still a risk of misalignment of these wrappers, considering people are building from different repos using different version of bazels;
2) The misalignment could also happen during bazel upgrading in certain repos. In worst cases, we might have to update these wrappers to make bazel upgrading work

Considering the wrappers in bazel are not changing very frequently, it probably will be fine. 

With that said, I'm more wondering how do we plan to use these wrappers? I see they are introduced in the pr.

deps.bzl
```diff
@@ -70,6 +70,16 @@ def archive_dependencies(third_party):
             "strip_prefix": "rules_docker-f4822f3921f0c343dd9e5ae65c760d0fb70be1b3",
             "urls": ["https://github.com/bazelbuild/rules_docker/archive/f4822f3921f0c343dd9e5ae65c760d0fb70be1b3.tar.gz"],
         },
+        
+        # Bazel is referenced as a dependency so that buildfarm can access the linux-sandbox as a potential execution wrapper.
+        {
+            "name": "bazel",
+            "sha256": "abdb1118e6b013062ed0d47f08d5311fd9ab506c10c093216df89f2b2be382f2",
+            "strip_prefix": "bazel-6b33bdb1e22514304c0e35ce8e067f2175685245",
+            "urls": ["https://github.com/bazelbuild/bazel/archive/6b33bdb1e22514304c0e35ce8e067f2175685245.tar.gz"],
```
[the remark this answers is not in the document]
luxe: thx! pinned at 3.7.2 since that's what we use internally.

BUILD
```diff
@@ -3,3 +3,36 @@ load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
...
+    srcs = ["@bazel//src/main/tools:process-wrapper"],
+    outs = ["process-wrapper"],
+    cmd = "cp $< $@;",
+)
+
+genrule(
+    name = "linux-sandbox.binary",
+    srcs = ["@bazel//src/main/tools:linux-sandbox"],
+    outs = ["linux-sandbox"],
+    cmd = "cp $< $@;",
```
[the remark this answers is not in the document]
luxe: I did it this way because that's how the internal repository does it.  But you're right, we could move these to a different BUILD file also.  All that matters is that they are referenced correctly when we build the java_image (which we also do at the top level):  
```
java_image(
    name = ...,
    base = ...,
    classpath_resources = [...],
    entrypoint = [...],
    main_class = ...,
    runtime_deps = [
        ":as-nobody",
        ":process-wrapper.binary",
        ":tini.binary",
        "@build_buildfarm//src/main/java/build/buildfarm:shard-worker",
    ],
)
```

I assume these will need adjusted anyhow when we source them from upstream.

luxe:
> 
> With that said, I'm more wondering how do we plan to use these wrappers? I see they are introduced in the pr.

Internally we are already using the process-wrapper and as-nobody wrapper.  I'm bringing them into the public repo, so we can source them from there instead.  Our internal configuration file for workers looks like this:  
```
execution_policies: {
  # default
  wrapper: {
    path: "/app/buildfarm/as-nobody"
  }
}
execution_policies: { # default
  wrapper: {
    path: "/app/buildfarm/process-wrapper"
  }
}
```

I'm also adding the sandbox (which is new) and we may or may not use it due to its speed.  But having it optionally available will be helpful since that's what users run locally.  And it has other useful cli options that we want to try out in the future.  tmpfs_dirs and network limiting come to mind.

> 1. if we are using wrappers from a certain version/commit of bazel, there is still a risk of misalignment of these wrappers, considering people are building from different repos using different version of bazels;
> Considering the wrappers in bazel are not changing very frequently, it probably will be fine.  

Good point.  This could be out-of sync with the actual process-wrapper or sandbox that is being used locally.  Which could cause slightly different behavior when comparing the environments. Hopefully they are standalone programs that don't need to follow bazel versions in a specific way.  Internally, we've been using the process-wrapper at bazel version `3.1.0`, despite moving the repo to `3.9.2`.  So the command line arguments could have changed, but we would need to pass them explicitly anyhow to use them in buildfarm.  So version differences seem to have worked so far in practice.

jacobmou:
> > With that said, I'm more wondering how do we plan to use these wrappers? I see they are introduced in the pr.
> 
> Internally we are already using the process-wrapper and as-nobody wrapper. I'm bringing them into the public repo, so we can source them from there instead. Our internal configuration file for workers looks like this:
> 
> ```
> execution_policies: {
>   # default
>   wrapper: {
>     path: "/app/buildfarm/as-nobody"
>   }
> }
> execution_policies: { # default
>   wrapper: {
>     path: "/app/buildfarm/process-wrapper"
>   }
> }
> ```
> 
> I'm also adding the sandbox (which is new) and we may or may not use it due to its speed. But having it optionally available will be helpful since that's what users run locally. And it has other useful cli options that we want to try out in the future. tmpfs_dirs and network limiting come to mind.
> 
> > 1. if we are using wrappers from a certain version/commit of bazel, there is still a risk of misalignment of these wrappers, considering people are building from different repos using different version of bazels;
> >    Considering the wrappers in bazel are not changing very frequently, it probably will be fine.
> 
> Good point. This could be out-of sync with the actual process-wrapper or sandbox that is being used locally. Which could cause slightly different behavior when comparing the environments. Hopefully they are standalone programs that don't need to follow bazel versions in a specific way. Internally, we've been using the process-wrapper at bazel version `3.1.0`, despite moving the repo to `3.9.2`. So the command line arguments could have changed, but we would need to pass them explicitly anyhow to use them in buildfarm. So version differences seem to have worked so far in practice.

Yes, I'm aware that we are already these wrappers and everything looks right. I'm just a little concerned because if there is any problem caused by wrapper version misalignment, debugging could be hard. Besides, we might want to document this and make sure users are aware of this.

jacobmou (approved):

luxe:
> Yes, I'm aware that we are already these wrappers and everything looks right. I'm just a little concerned because if there is any problem caused by wrapper version misalignment, debugging could be hard. Besides, we might want to document this and make sure users are aware of this.

Ah.  I'll add a comment about this on the dependencies now.  We can think about to make debugging easier.  I'm hoping to make the wrappers themselves optional based exec_properties which which would help debugging individual actions.  I wonder if the bazel client sends its version to buildfarm.. I'll look into that as well.

luxe:
related to https://github.com/bazelbuild/bazel-buildfarm/issues/552

as-nobody.c
```diff
@@ -0,0 +1,30 @@
...
+  }
+  if (setregid(pw->pw_gid, pw->pw_gid) < 0) {
+    perror("setgroups");
+    return EXIT_FAILURE;
+  }
+  if (setreuid(pw->pw_uid, pw->pw_uid) < 0) {
+    perror("setreuid");
+    return EXIT_FAILURE;
+  }
+  execvp(argv[1], argv);
```
philsc, line 27: Shouldn't this be?
```cpp
  execvp(argv[1], argv + 1);
```

werkt: Negative. Per the manpage:

The execv(), execvp(), and execvpe() functions provide an array of pointers to null-terminated strings that represent the argument list available to the new program. The first argument, by convention, should point to the filename associated with the file being executed.

werkt: Ayyy, good call. (sorry if you saw deleted, reread code and suggestion and thought better of it) Fixed up now with quick edit.

PR #666, opened 2021-02-05
Add details regarding jedis fork

### LINKED ISSUE #664: Kindly document the Jedis fork
Buildfarm is using a Jedis fork, rather than [some official release of Jedis](https://github.com/redis/jedis/releases):

<https://github.com/bazelbuild/bazel-buildfarm/blob/2b7c7d420f91e25af80afccb1c18343dc22596a3/deps.bzl#L89-L96>

Could you please add comments in the code above describing the reason for using this fork?

# Background

The reason I'm asking is that we have made [a Jedis PR](https://github.com/redis/jedis/pull/2358) for backing off on cluster connection failures.

At some point, we'd like for Buildfarm to start using a Jedis version with backoff enabled.

And as long as Buildfarm uses a fork of Jedis, it's hard for us to know what a Jedis version bump would require.

werkt:
Fixes #664

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -96,6 +96,18 @@ def buildfarm_dependencies(repository_name = "build_buildfarm"):
         name = params.pop("name")
         maybe(http_archive, name, **params)
 
+    # Enhanced jedis 3.2.0 containing several convenience, performance, and
+    # robustness changes.
+    # Notable features include:
+    #   Cluster request pipelining, used for batching requests for operation
+    #   monitors and CAS index.
+    #   Blocking request (b* prefix) interruptibility, using client
+    #   connection reset.
+    #   Singleton-redis-as-cluster - support treating a non-clustered redis
+    #   endpoint as a cluster of 1 node.
+    # Other changes are redis version-forward treatment of spop and visibility
+    # into errors in cluster unreachable and cluster retry exhaustion.
+    # Details at https://github.com/werkt/jedis/releases/tag/3.2.0-e82e68e2f7
     maybe(
         http_jar,
         "jedis",


luxe:
I'm guessing walles wants to upgrade redis for upstream fixes.  Maybe we can contribute your fork back to upstream as well?

luxe (approved):

werkt:
> I'm guessing walles wants to upgrade redis for upstream fixes. Maybe we can contribute your fork back to upstream as well?

Possibly, but the cluster pipelining thing was requested a while ago and rejected. I could pursue some of them, but I'm guessing nobody is going to be interested in the slf4j removal.

walles:
> I'm guessing nobody is going to be interested in the slf4j removal.

Instead of removing SLF4J from Jedis, Buildfarm should just add `slf4j-jdk14.jar` to its classpath:
<http://www.slf4j.org/faq.html#where_is_binding>

werkt:
> > I'm guessing nobody is going to be interested in the slf4j removal.
> 
> Instead of removing SLF4J from Jedis, Buildfarm should just add `slf4j-jdk14.jar` to its classpath:
> http://www.slf4j.org/faq.html#where_is_binding

That was what we were trying to avoid... also, we don't appear to be at a minimum of java14 everywhere (macos).

Random logging endpoints introduced by dependencies causes difficult to bound messaging.

walles:
> That was what we were trying to avoid... also, we don't appear to be at a minimum of java14 everywhere (macos).

The `jdk14` part of `slf4j-jdk14.jar` refers to "Java 1.4", which is where Java Logging was introduced.

Yay for version numbering schemes :)

PR #668, opened 2021-02-06
Package relocations under subdirectories.

werkt:
No more java_library in //src/main/java/build/buildfarm.
All libraries are now available in subpackages.
ShardBackplane is now Backplane and has its own namespace for
further interface refinement.
WorkerQueues were relocated from instance/memory to instance
OperationsFinder implementation moved to subpackage in advance of
further backplane refinement.
AbstractServerInstance and friends now exist in their own
instance/server subpackage.
TreeIterator has finally been removed from instance.
MatchListener is in its own unit.
No more java_test in //src/test/java/build/buildfarm.
Tests are now all named :tests under their subpackage.
Some minor visibility changes were accomplished.


luxe (approved):
great! I don't see any behavior differences. Thanks for clarifying those redis native tests- I still want to keep those.  Exception handling on that json is better.  Yup, looks good to land.

luxe:
please land soon! So we can fix merge conflicts instead of you.

PR #671, opened 2021-02-09
Add prometheus metrics publishing

80degreeswest:
The following metrics will be exposed initially:

- JVM Metrics
- cpu_queue_size
- gpu_queue_size
- pre_queue_size
- dispatched_operations_size
- worker_pool_size
- action_results
- execution_time_ms
- execution_stall_time_ms
- execution_slot_usage
- input_fetch_time_ms
- input_fetch_stall_time_ms
- input_fetch_slot_usage
- cluster_utilization
- execution_success
- completed_operations
- missing_blobs

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -39,6 +39,7 @@ import build.buildfarm.common.redis.RedisMap;
 import build.buildfarm.common.redis.RedisNodeHashes;
 import build.buildfarm.instance.Instance;
 import build.buildfarm.instance.shard.RedisShardSubscriber.TimedWatchFuture;
+import build.buildfarm.metrics.prometheus.PrometheusPublisher;
 import build.buildfarm.operations.FindOperationsResults;
 import build.buildfarm.operations.FindOperationsSettings;
 import build.buildfarm.operations.OperationsFinder;
@@ -51,6 +52,7 @@ import build.buildfarm.v1test.OperationChange;
 import build.buildfarm.v1test.OperationsStatus;
 import build.buildfarm.v1test.ProvisionedQueue;
 import build.buildfarm.v1test.QueueEntry;
+import build.buildfarm.v1test.QueueStatus;
 import build.buildfarm.v1test.QueuedOperationMetadata;
 import build.buildfarm.v1test.RedisShardBackplaneConfig;
 import build.buildfarm.v1test.ShardWorker;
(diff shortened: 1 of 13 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1479,4 +1490,40 @@ public GetClientStartTimeResult getClientStartTime(String clientKey) throws IOEx
...
+
+    prometheusMetricsThread.start();
+  }
+
+  private void updateQueueSizes(List<QueueStatus> queues) {
+    for (QueueStatus queueStatus : queues) {
+      if (queueStatus.getName().contains("cpu")) {
+        PrometheusPublisher.updateCpuQueueSize(queueStatus.getSize());
+      } else if (queueStatus.getName().contains("gpu")) {
+        PrometheusPublisher.updateGpuQueueSize(queueStatus.getSize());
```
luxe, line 1525: Can this be more generic? The names of the queues are customizable and I would be nice if this were derived instead.  If not, its okay for now.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1479,4 +1490,40 @@ public GetClientStartTimeResult getClientStartTime(String clientKey) throws IOEx
...
+              break;
+            } catch (Exception e) {
+              logger.log(Level.SEVERE, "Could not update RedisShardBackplane metrics", e);
+            }
+          }
+        },
+        "Prometheus Metrics Collector");
+
+    prometheusMetricsThread.start();
+  }
```
luxe, line 1518: how does prometheus work when people don't intend to use it? Does it break users anyway?  Not sure if there needs to be a config  `use_prometheus: true/false`

luxe (approved):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
jacobmou: Good point! It will send to the default port `8080` if port number is not set. I dont think this will break anything, but setting have a config explicitly might be useful to make everything more clear.

jacobmou (approved):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
80degreeswest: You don't really have to use it - the metrics will be collected and exposed on default port 9090. I'm going to change it to explicitly configure the port or not start the web server by default though.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
80degreeswest: Yeah it needs to be made generic. I just didn't think of a way to do it yet, but wanted to at least have these exposed in the initial run rather than not.

PR #674, opened 2021-02-11
Fix CasFIleCache to not return empty blobs in findMissingBlobs

sbueringer:
@werkt I think the previous fix from PR #670 was not entirely correct. This one works (at least in my environment)


werkt (approved):
Nice fix, thanks!

PR #310, opened 2019-12-19
Initial support for integration tests

werkt:
Provide a memory/operationqueue integration test example via shell,
adapting bazel's integration test suite.

PR #676, opened 2021-02-12
(CI) Buildfarm integration test

luxe:
Protect master from any major service regressions.  This ensures that the worker and server can start, that they can connect to their storage backends, and that the bazel client can be used against these instantiated services to build targets.


80degreeswest (approved):

80degreeswest (approved):

luxe:
The integration test takes about 3 minutes.  Maybe we could have the container dependencies already built for some stability, but we'll see if there are any future issues with it.  I'm not able to test the main bazel CI with PRs, so if this breaks anything over there, people from the bazel team usually notify downstream projects.

Previous work on an integration test was done here:  https://github.com/bazelbuild/bazel-buildfarm/pull/310
There are some ideas we may want to adopt from there.  In particular, we are currently parsing the stdout of the server to determine when its started which is not as robust as having the server directly say when its ready.

PR #678, opened 2021-02-14
Limit number of requeue attempts for an operation

80degreeswest:
It is possible for a bad action (such as one requiring specific hardware to be present) to get stuck in the "EXECUTING" state indefinitely. Even upon the build termination, the executing operation persists until the worker is terminated. This change caps the amount of retries to 5, which will cause the stuck action to error out after it's retried this many times. Note that the action will still be retried by Bazel (for a default of 5 times), unless --remote_retries flag is specified otherwise.


luxe (approved):
nice fix!!  I suppose the other issue is that actions should not be hanging in execution.  If the action timeout is not being obeyed that's a bug in the executor.  But this will ensure stability around those problems.

PR #682, opened 2021-02-17
Refactor prometheus metrics and fix queue size reporting

80degreeswest:


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -432,6 +451,55 @@ public void run() {
...
+              .name("queue_size")
+              .labelNames(getQueueNames(queues))
+              .help("Queue size.")
+              .register();
+    } else {
+      queueSize = null;
+    }
+
+    prometheusMetricsThread =
+        new Thread(
```
luxe, line 467: There may be an advantage to using Java's TimerTask so we publish at a consistent rate, but its not really an issue here.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -432,6 +451,55 @@ public void run() {
...
     }
+
+    if (queues.size() > 0) {
+      queueSize =
+          Gauge.build()
+              .name("queue_size")
+              .labelNames(getQueueNames(queues))
+              .help("Queue size.")
+              .register();
+    } else {
```
luxe, line 462: thx!

src/main/java/build/buildfarm/metrics/prometheus/PrometheusPublisher.java
```diff
@@ -86,79 +37,9 @@ public static void startHttpServer(int port) {
...
-    inputFetchStallTime.observe(val);
-  }
-
-  public static void updateInputFetchSlotUsage(int val) {
-    inputFetchSlotUsage.set(val);
-  }
-
-  public static void updateClusterUtilization() {
-    clusterUtilization.set(executionSlotUsage.get() * 100 / numExecuteStages);
-  }
```
luxe, line 163: Is the diff removing any metrics? I see we're pulling metric gathering out of the backplane and into the instance. But I'm not sure if we've shrunk the number of metrics because of it.

luxe (approved):
Yes, looks like an improvement.  I know there were some concerns about mixing Prometheus into the project. It's good to have it out of the backplane (assuming you can still get the metrics you need).  We could probably go a step further in isolating it by makinh it a separate module for gathering/publishing.  Basically have the ShardInstance only need to call a single line.

src/main/java/build/buildfarm/metrics/prometheus/PrometheusPublisher.java
80degreeswest: I realized that with having execute stage metric our cluster utilization calculation is very simple now and can just be calculated on the dashboard side.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -478,6 +531,9 @@ public void stop() throws InterruptedException {
     if (dispatchedMonitor != null) {
       dispatchedMonitor.stop();
     }
+    if (prometheusMetricsThread != null) {
+      prometheusMetricsThread.stop();
```
jacobmou, line 535: `Thread.stop()` is deprecated and `interrupt()` might be the method we want to use.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -432,6 +452,35 @@ public void run() {
...
+            () -> {
+              while (true) {
+                try {
+                  TimeUnit.SECONDS.sleep(30);
+                  OperationsStatus operationsStatus = operationsStatus();
+                  workerPoolSize.set(operationsStatus.getActiveWorkersCount());
+                  dispatchedOperations.set(operationsStatus.getDispatchedSize());
+                  preQueueSize.set(operationsStatus.getPrequeue().getSize());
+                  updateQueueSizes(operationsStatus.getOperationQueue().getProvisionsList());
+                } catch (InterruptedException e) {
```
jacobmou, line 467: If `Thread.stop()` is used, there won't be any `InterruptedException`, but the thread get killed directly. 

Considering `stop()` is deprecated (see my comment below), we might want to use `Thread.interrupt()`. If that's the case, the behavior of the thread might be different from our expectation: If the current thread is interrupted when it's sleeping (i.e. ` TimeUnit.SECONDS.sleep(30)`, it will throw InterruptedException, which is fine. However, if it's not sleeping, there wont be any execution, we have to do check the status of `Thread.currentThread.isInterrupted()` to know if the thread should be stopped. If ignore it, the thread wont exit.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
jacobmou: https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#stop--

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
werkt: Prefer scheduled executor service to TimerTask. Unless you believe that they mesh well.

80degreeswest: We should probably stay consistent and utilize the same service for all whatever one we would pick.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
80degreeswest: Good catch, I see we are using stop() in other places as well - this may be a good story to replace all of them in the same PR. Not really sure if it matters in this use case (same with DispatchedMonitor and other threads) as in this case the main process is already stopping so we just want to shut everything down.

jacobmou: Yeah, since we are shutting everything done, the problems of `Thread.stop()` are not really concerns. We don't have to deal with it right now.

jacobmou (approved):

PR #685, opened 2021-02-19
refactor exec_properties

luxe:
Organize existing exec_properties to prepare for the addition of "memory constraint" properties.
Improve the in-code documentation of exec_properties.

### DIFF
--- a/src/main/java/build/buildfarm/common/ExecutionProperties.java
+++ b/src/main/java/build/buildfarm/common/ExecutionProperties.java
@@ -1,0 +1,86 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


jacobmou:
The pr itself looks good and the documentation is awesome! 
Since you mentioned `memory constraint` addition, I'm more curious about the plan, like which idea in discussion in https://github.com/bazelbuild/bazel-buildfarm/issues/673 do you plan to further explore?

jacobmou (approved):

luxe:
> Since you mentioned `memory constraint` addition, I'm more curious about the plan, like which idea in discussion in #673 do you plan to further explore?

I like idea 3.  It requires the workers to know the memory constraints for each action.  
That would let workers delay executions as needed.  It seems like the closet thing to what the user in the issue is doing.   
But we can discuss that further before doing anything concrete.

At the very least, I think we can make the memory tags available, and ensure that they are parsed and stored the same way as the cpu tags-- even if we aren't acting on them right away.

Maybe an earlier proof of concept should be to extend the cgroups implementation to also set memory limits based on a config.
Then if we tag an action with `max-mem: 4GB`, and the action allocates `5GB` we should see the action fail (eventhough there is more memory available on the worker).  That would be a nice start.

PR #691, opened 2021-02-25
[bug] add explanation for resolving block list error

luxe:
A user hit this error message and wasn't sure what to do.  Add more context to the error message.


jacobmou (approved):
Thanks for the extra explanation!

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1842,6 +1835,14 @@ public void observe(Operation operation) {
...
   }
 
+  private String blockListErrorMessage() {
+    StringBuilder message = new StringBuilder();
+    message.append("This request is in block list and is forbidden.  ");
+    message.append("To resolve this error, you can tag the rule with 'no-remote'.  ");
+    message.append("You can also adjust the action behavior to attempt a different action hash.  ");
+    return message.toString();
+  }
+
```
werkt, line 1845: Please present this string in the current location for error messages: https://github.com/bazelbuild/bazel-buildfarm/blob/4f0c6f595126950905e5ea625eacddc06e63cafa/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java#L145-L200, or something more pertinent for ShardInstance if you feel it is isolated enough to it. StringBuilder is not a facility to be used just for 'long' error messages.

werkt (changes requested):

luxe:
@werkt fixed

luxe:
@werkt bump!

werkt:
This response is very bazel specific. When users come at it with confusion over no-remote, I'll prefer a tool check before advising.

werkt (approved):

luxe:
> This response is very bazel specific. When users come at it with confusion over no-remote, I'll prefer a tool check before advising.

yes, good idea. It will be cool to see buildfarm work with non-bazel clients.  For now, this helps guide our internal colleagues.

PR #693, opened 2021-02-26
[integration test] allow apt-get to retry for network stability

luxe:
Ubuntu packages were recently 404ing.  One way to make this more stable in the future is to let apt-get retry a few times.


jlisee:
I wonder if there is some base imaging caching based system we can use such that you don't re-generate the base image with the deps unless it changes.

luxe:
That would be better. It's only a few common deps on top of ubuntu and it all runs in ~3mins. But now that
s 404'd I'd rather have something with as little network as possible.

werkt (approved):
Pretty sure this is harmless, and very old.

luxe:
I haven't seen any new ubuntu package outrages.  We can land anyways.  The next improvement would be to use image caching as discussed above

PR #694, opened 2021-02-26
[feature] add cgroup memory restrictions via min/max-mem flags

luxe:
### Summary:
We extend the functionality of the cgroups implementation to additionally restrict memory usage.  By default, these memory restrictions are not applied to actions, and landing this diff should have no effect on current usage of buildfarm.  Memory restrictions are only applied when a user passes `min-mem` / `max-mem`.

### Cgroups setting:
The current implementation of this is done via cgroup's `memory.limit_in_bytes`.  There is also the concept of swap memory (`memory.memsw.limit_in_bytes`) which we are not setting right now. 

### Refactors:  
We refactored the implementation of `limitSpecifiedExecution` in order to incorporate this new group.  I had to do something kind of weird with the `IOResource`-merging in order to honor the function signature.  It seems fine, but the whole IOResource and current cgroup code made me a little uneasy.  Let me know if you have other suggestions around this.

### Observe executions: 
This is a demonstration of the feature using a locally running buildfarm server/worker.  
First, let's acknowledge the existing core limiting:  
  
  ```
bazel test --remote_default_exec_properties='debug-before-execution=true'  \
--remote_default_exec_properties='min-cores=2' \
--remote_executor=grpc://127.0.0.1:8980 \
--noremote_accept_cached  \
--nocache_test_results 
//foo
  ```
  
  ```
cat source/bazel-out/k8-fastbuild/testlogs/foo/test.log                                                                                                                                                                                                         
Buildfarm debug information before execution:
/usr/bin/cgexec -g cpu:executions/operations/1756bb6a-bf53-4fa7-be39-98e6acc530ab external/bazel_tools/tools/test/test-setup.sh main
{"cpu":{"limit":true,"min":2,"max":0,"claimed":2},"mem":{"limit":false,"min":0,"max":0,"claimed":0},"extraEnvironmentVariables":{},"debugBeforeExecution":true,"debugAfterExecution":false}
```
This shows us that core limiting is set to true, and we claimed the appropriate core amount.
Notice that the actual CLI that will be run is using the cpu cgroup (`cpu:executions/`).  

Next we will show this with adding memory restrictions
  ```
bazel test --remote_default_exec_properties='debug-before-execution=true'  \
--remote_default_exec_properties='min-cores=2' \
--remote_default_exec_properties='min-mem=5242880' \
--remote_executor=grpc://127.0.0.1:8980 \
--noremote_accept_cached  \
--nocache_test_results 
//foo
  ```
  
This shows that both resources are being limited and both cgroups are being chosen:  
```
cat source/bazel-out/k8-fastbuild/testlogs/foo/test.log
Buildfarm debug information before execution:
/usr/bin/cgexec -g cpu,memory:executions/operations/a60d8cfb-4501-4f51-a676-4adf736b53a6 external/bazel_tools/tools/test/test-setup.sh code/tools/example_tests/bash_1mb_allocate/main
{"cpu":{"limit":true,"min":2,"max":0,"claimed":2},"mem":{"limit":true,"min":5242880,"max":0,"claimed":5242880},"extraEnvironmentVariables":{},"debugBeforeExecution":true,"debugAfterExecution":false}
```

Note the change in CLI (`cpu,memory:executions/`).

### Does it work?  
Toying with the allowable min-mem I can get sample tests to always pass or always fail:  
 ```
bazel test --remote_default_exec_properties='debug-before-execution=true'  \
--remote_default_exec_properties='min-cores=2' \
--remote_default_exec_properties='min-mem=5242880' \
--remote_executor=grpc://127.0.0.1:8980 \
--noremote_accept_cached  \
--nocache_test_results 
--runs_per_test=5
//foo
//foo                        PASSED in 15.1s
  Stats over 5 runs: max = 15.1s, min = 5.1s, avg = 9.1s, dev = 3.8s
  ```

 ```
bazel test --remote_default_exec_properties='debug-before-execution=true'  \
--remote_default_exec_properties='min-cores=2' \
--remote_default_exec_properties='min-mem=4242880' \
--remote_executor=grpc://127.0.0.1:8980 \
--noremote_accept_cached  \
--nocache_test_results 
--runs_per_test=5
//foo
//foo                        FAILED in 5 out of 5 in 10.4s
  ```
  
  ### What did your tests look like?  
  variations of this:  
  ```
#!/bin/bash
# use certain amount of memory for n seconds.
# not perfect, since bash and other tools are also going to use memory.
total_bytes=$((1024**2*1))
sleep_s=5;
cat <( </dev/zero head -c $total_bytes) <(sleep $sleep_s) | tail
```
  
### Followup:
I rather pass things like `min-mem=250MB` /  `min-mem=4GB`'.  Possibly improve later.
Also, if it stays in bytes, we need to change int to long so it can address over 2/4GB.  I'll do that after.


jlisee:
Can you use the integration testing setup you have created to turn the manual test you have done here into an automated regression test for this feature?

luxe:
> Can you use the integration testing setup you have created to turn the manual test you have done here into an automated regression test for this feature?

yes, we'll need to add cgroup-tools to the container, and also use different config files instead of the default. I want to ensure the default configs work out of the box, so maybe it becomes another test plan that's like: "execution features"

luxe:
this PR is needed to further address https://github.com/bazelbuild/bazel-buildfarm/issues/673

jacobmou (approved):

PR #696, opened 2021-02-26
Document CAS hex_bucket_levels

### LINKED ISSUE #695: Please document hex_bucket_levels
Introduced here: #589

It would be great if `hex_bucket_levels` could be documented in the example config files with an exhaustive comment.

walles:
And provide a sensible default in the example config file.

Fixes #695.

### DIFF
--- a/examples/shard-worker.config.example
+++ b/examples/shard-worker.config.example
@@ -69,6 +69,10 @@ cas: {
     # whether the file directories bidirectional mapping should be stored in memory (HashMap) or
     # in sqlite
     file_directories_index_in_memory: false
+
+    # setup a folder hierarchy to improve filesystem performance
+    # see https://github.com/bazelbuild/bazel-buildfarm/issues/568#issuecomment-728223668
+    hex_bucket_levels: 2
   }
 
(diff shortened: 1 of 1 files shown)


examples/shard-worker.config.example
```diff
@@ -69,6 +69,10 @@ cas: {
     # whether the file directories bidirectional mapping should be stored in memory (HashMap) or
     # in sqlite
     file_directories_index_in_memory: false
+
+    # setup a folder hierarchy to improve filesystem performance
+    # see https://github.com/bazelbuild/bazel-buildfarm/issues/568#issuecomment-728223668
```
werkt, line 74: I prefer not to make users editing config files jump off to web documentation to discover what a field really does. 'Improve filesystem performance' is also a misnomer here.

If you do want to incorporate this, please explain its function, rather than a possible outcome. Feel free to mention that on some filesystems with some worker loads, this will improve performance by reducing the overall size of directories. Otherwise I can put up something suitable independently.

werkt:
Closed in favor of #697

PR #697, opened 2021-02-28
Document and normalize worker CFC proto fields

werkt:
Provide an explanation of the functionality and possible benefits of
hex_bucket_levels.
Normalize controls and support for hex_bucket_levels and
file_directories_index_in_memory in operationqueue worker.


examples/worker.config.example
```diff
@@ -46,6 +46,30 @@ root: "/tmp/worker"
...
+# the default, unspecified value is 0, to create all digests in the cas path itself
+# for each level > 0, a directory is created for all byte representations, 00-ff, with
+# as many sublevels as requested.
+#
+# This can provide a performance benefit for file access and manipulation for filesystems
+# which have problems with directories with many (e.g. millions) of files in them.
+# The CFC creates these directories upfront, resulting in a delay while each is verified or
+# created. Be cautious with large values, as the subdivision is 256^n
+#
+cas_cache_hex_bucket_levels: 0
```
walles, line 71: Shouldn't this be just `hex_bucket_levels`? Maybe I'm missing something obvious though, that has happened before...

examples/shard-worker.config.example
```diff
@@ -69,6 +69,17 @@ cas: {
...
+    # create entry files in a directory for each leveled leading byte of the hash
+    # the default, unspecified value is 0, to create all digests in the cas path itself
+    # for each level > 0, a directory is created for all byte representations, 00-ff, with
+    # as many sublevels as requested.
+    #
+    # This can provide a performance benefit for file access and manipulation for filesystems
+    # which have problems with directories with many (e.g. millions) of files in them.
+    # The CFC creates these directories upfront, resulting in a delay while each is verified or
+    # created. Be cautious with large values, as the subdivision is 256^n
+    # hex_bucket_levels: 0
```
walles, line 82: Shouldn't this match `worker.config.example`? Right now they use different variable names.

examples/worker.config.example
werkt: The operationqueue worker does not use the cas configurations that are available in the shard worker - no support for delegation, and no containerization of the configs. There will likely be efforts to singularize this stream, and create one worker that can satisfy both behaviors with just singular config differences. This is a stopgap until that time, when a cas specification is sufficient and minimal.

examples/shard-worker.config.example
werkt: Situation as with above.

PR #699, opened 2021-03-02
[backplane] Make "prequeue max depth" a part of the prequeue

luxe:
Move logic out of the backplane into the queue implementation.


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1431,8 +1432,7 @@ private boolean isBlacklisted(JedisCluster jedis, RequestMetadata requestMetadat
 
   @Override
   public boolean canQueue() throws IOException {
-    int maxQueueDepth = config.getMaxQueueDepth();
-    return maxQueueDepth < 0 || client.call(jedis -> operationQueue.size(jedis)) < maxQueueDepth;
+    return client.call(jedis -> prequeue.canQueue(jedis));
```
werkt, line 1435: The point of the logic here was to avoid tying up a jedis client while a possible check didn't need to take place. I suppose with the pool and the cpu-only comparison, it shouldn't matter.

werkt (approved):

PR #702, opened 2021-03-02
[backplane] Make "queue max depth" a part of the queue

luxe:
refactor and moving logic out of the backplane.


werkt (approved):
This and the prequeue limiting could use some tests.

PR #708, opened 2021-03-05
Add code quality gate to CI

luxe:
Let's enable static analysis in the CI using [pmd](https://pmd.github.io/).
To begin with, most of these rules are disabled.  
As problems are fixed in the codebase, these rules can be re-enabled.


jacobmou (approved):
This is a great start point! 
Looks like buildfarm code actually works with many of the excluded rules, but we can work on trimming the list later when we want.

PR #712, opened 2021-03-08
fix & enable "ArrayIsStoredDirectly"

luxe:
fix static analysis warning and enable it for future PRs.


werkt (changes requested):
This will break the behavior of the status counts being updated by the observer, the array must be assigned by reference to be usable.

PR #713, opened 2021-03-08
[configuration] fix misuse of getProcessingTimeoutMillis

luxe:
`scanProcessing` and `scanDispatching` are both using the same function `getExpiresAt`.  

This results in a problem where the dispatching code is using the wrong configuration.
`processingOperations` should use `config.getProcessingTimeoutMillis`.
`dispatchedOperations` should use `config.getDispatchingTimeoutMillis`.

Also, the configuration value is being multiplied by 2.
This is not necessary.  If needed, it can be configured to be twice as much.  

We resolve these above issues and represent processingOperatiosn and dispatchedOperations as containers.


jacobmou (approved):

werkt:
As it turns out, the * 2 here was only used for the expiration, which, when removed, permits the Dispatching key to be expired *before* the next possible check that could possibly observe it as being out of date. Will be correcting this to be a minimum of 1hr, or 10x the applicable timeout

PR #715, opened 2021-03-09
[metrics] add prometheus metrics for instance action cache

luxe:
Buildfarm's action cache is multi-layered.  There is the distributed redis implementation found in the backplane.  We are recording the size for that.  However, there is also a local readthrough cache found on each scheduler.  This local cache consults the backplane's action cache as needed.  To understand this interaction better, we extract and publish these cache stats to prometheus.  

After converting to caffeine we use Prometheus's support for cache metrics.


80degreeswest (commented):
Have you done any testing on he performance here? I'm a bit concerned we may introduce some degradation here with either speed or memory usage.

luxe:
> Have you done any testing on he performance here? I'm a bit concerned we may introduce some degradation here with either speed or memory usage.

yeah good point.  I'll need to find a way to profile this somehow.  Since its not running in its own Prometheus thread, I could see it affecting performance.

PR #716, opened 2021-03-09
[static analysis burndown] fix modifers and enable static check

luxe:
- interfaces don't need static modifiers
- final classes don't need visibility specifiers.
 - Fix issues and enable the static analysis gate.


werkt (commented):
Hold this for the caffeine transition, It will look better then.

luxe:
> Hold this for the caffeine transition, It will look better then.

updated

jacobmou (approved):

PR #719, opened 2021-03-10
Replace guava fork with caffeine

luxe:
Use upstream library caffeine to replace `ShardActionCache` and `ShardInstance` guava usage.
See discussion: https://github.com/bazelbuild/bazel-buildfarm/issues/718


defs.bzl
```diff
@@ -77,6 +77,8 @@ def buildfarm_init(name = "buildfarm"):
                         "com.googlecode.json-simple:json-simple:1.1.1",
                         "com.google.code.gson:gson:2.8.6",
                         "com.jayway.jsonpath:json-path:2.4.0",
+                        "com.github.ben-manes.caffeine:caffeine:2.9.0",
+                        "net.javacrumbs.future-converter:future-converter-java8-guava:1.2.0",
```
werkt, line 81: alpha order this below junit

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
```diff
@@ -17,47 +17,53 @@
...
+
+          @Override
+          public CompletableFuture<ActionResult> asyncLoad(ActionKey actionKey, Executor executor) {
+            return toCompletableFuture(
+                catching(
+                    service.submit(() -> backplane.getActionResult(actionKey)),
+                    IOException.class,
+                    e -> {
+                      throw Status.fromThrowable(e).asRuntimeException();
+                    },
```
werkt, line 55: We should investigate if this grpc-awareness is actually necessary - it really belongs at the service level to transform if used in a response. Also, this is a little cleaner as a non-throw catchingAsync.

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
```diff
@@ -76,11 +82,11 @@ public void put(ActionKey actionKey, ActionResult actionResult) {
 
   @Override
   public void invalidate(ActionKey actionKey) {
-    actionResultCache.invalidate(actionKey);
+    actionResultCache.synchronous().invalidate(actionKey);
   }
 
   @Override
   public void readThrough(ActionKey actionKey, ActionResult actionResult) {
```
werkt, line 89: I still question this name, maybe I'm off base here.

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
```diff
@@ -17,47 +17,53 @@
...
+        };
+
+    actionResultCache = Caffeine.newBuilder().maximumSize(maxLocalCacheSize).buildAsync(loader);
   }
 
   @Override
   public ListenableFuture<ActionResult> get(ActionKey actionKey) {
     return catching(
-        checkNotNull(actionResultCache.get(actionKey)),
+        checkNotNull(toListenableFuture(actionResultCache.get(actionKey))),
```
werkt, line 66: I believe the behavior of caffeine is to throw an NPE on anything that would return a null, we might not need the check.

src/main/java/build/buildfarm/instance/shard/BUILD
```diff
@@ -34,5 +34,6 @@ java_library(
         "@remote_apis//:build_bazel_remote_execution_v2_remote_execution_java_grpc",
         "@remote_apis//:build_bazel_remote_execution_v2_remote_execution_java_proto",
         "@maven//:org_redisson_redisson",
+        "@maven//:net_javacrumbs_future_converter_future_converter_java8_guava",
```
werkt, line 37: Alpha order deps here and above.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1195,21 +1197,22 @@ boolean next() {
...
-              public ListenableFuture<Directory> call() {
-                logger.log(
-                    Level.FINE,
-                    format(
-                        "transformQueuedOperation(%s): fetching directory %s",
-                        reason, DigestUtil.toString(directoryBlobDigest)));
-                return fetcher.get();
-              }
-            }),
+        toListenableFuture(directoryCache.get(directoryBlobDigest, getCallback)),
```
werkt, line 1215: I don't think this catching is necessary at this point. As indicated by @ben-manes, I believe that a missing entry from caffeine will return a null. Please verify with tests and ensure we've got coverage.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1239,15 +1242,17 @@ boolean next() {
       Digest commandBlobDigest, Executor executor, RequestMetadata requestMetadata) {
     Supplier<ListenableFuture<Command>> fetcher =
         () -> notFoundNull(expect(commandBlobDigest, Command.parser(), executor, requestMetadata));
+
+    BiFunction<Digest, Executor, CompletableFuture<Command>> getCallback =
+        new BiFunction<Digest, Executor, CompletableFuture<Command>>() {
+          @Override
+          public CompletableFuture<Command> apply(Digest digest, Executor executor) {
+            return toCompletableFuture(fetcher.get());
+          }
```
werkt, line 1251: Should we not be using the supplied executor in the fetcher's call to expect at this point?

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -193,15 +196,14 @@
...
-      recentCacheServedExecutions =
-          com.google.common.cache.CacheBuilder.newBuilder().maximumSize(64 * 1024).build();
+  private final AsyncCache<Digest, Directory> directoryCache =
+      Caffeine.newBuilder().newBuilder().maximumSize(64 * 1024).buildAsync();
+  private final AsyncCache<Digest, Command> commandCache =
+      Caffeine.newBuilder().newBuilder().maximumSize(64 * 1024).buildAsync();
+  private final AsyncCache<Digest, Action> actionCache =
+      Caffeine.newBuilder().newBuilder().maximumSize(64 * 1024).buildAsync();
+  private final AsyncCache<RequestMetadata, Boolean> recentCacheServedExecutions =
+      Caffeine.newBuilder().newBuilder().maximumSize(64 * 1024).buildAsync();
```
werkt, line 206: This should not be an async cache - note the use of the explicit guava upstream synchronous cache type. If you want to use caffeine for everything (I suppose I'm fine with that) then at least use the sync versions of their caches.

I also don't know how you got away with this without changing any of the usages, unless they default to synchronous, and we're only using the sync api.

src/main/java/build/buildfarm/operations/finder/BUILD
```diff
@@ -8,7 +8,7 @@ java_library(
     deps = [
         "//src/main/java/build/buildfarm/ac",
         "//src/main/java/build/buildfarm/common",
-        "//src/main/java/build/buildfarm/common/cache",
+        "@maven//:com_github_ben_manes_caffeine_caffeine",
```
werkt, line 11: alpha order

src/main/java/build/buildfarm/worker/shard/BUILD
```diff
@@ -6,7 +6,7 @@ java_library(
         "//src/main/java/build/buildfarm/backplane",
         "//src/main/java/build/buildfarm/cas",
         "//src/main/java/build/buildfarm/common",
-        "//src/main/java/build/buildfarm/common/cache",
+        "@maven//:com_github_ben_manes_caffeine_caffeine",
```
werkt, line 9: alpha order

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1904,7 +1911,7 @@ private void deliverCachedActionResult(
       Operation operation,
       RequestMetadata requestMetadata)
       throws Exception {
-    recentCacheServedExecutions.put(requestMetadata, true);
+    recentCacheServedExecutions.put(requestMetadata, CompletableFuture.completedFuture(true));
```
werkt, line 1914: unnecessary once you swap recentCacheServedExecutions to be synchronous.

werkt (changes requested):

defs.bzl
luxe: fixed

src/main/java/build/buildfarm/worker/shard/BUILD
luxe: fixed

src/main/java/build/buildfarm/operations/finder/BUILD
luxe: fixed

src/main/java/build/buildfarm/instance/shard/BUILD
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
luxe: It seemed okay to me.  we can simplify things more when framework is swapped.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
luxe: fixed. back to regular Cache

src/main/java/build/buildfarm/instance/shard/ShardInstance.java

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
luxe: moved fetcher inside of callback to use executor

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
luxe: which part makes you say it's grpc-aware?  I'm not sure how to adjust this, as it matches to previous implementation.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java

src/main/java/build/buildfarm/instance/shard/ShardActionCache.java
werkt: The grpc awareness is endemic to the Status translation. It's my gripe that anything below the service implementation, and above the stub, knows anything about grpc.

werkt (approved):
gogogo

PR #721, opened 2021-03-11
[static analysis burndown] remove identical catch branch

luxe:
fix issue and enable static analysis for future violations


luxe:
> I mean, I guess. Except that I've got these comments in there... Adapt those to only pertain to the misconfiguration and you've got a deal.

fixed

jacobmou (approved):

PR #722, opened 2021-03-11
[static analysis burndown] remove existing dead code

luxe:
This removes all of the private data members and private methods that are not being used.  We enable a static analysis to enforce this going forward.  I kept a worker profile method by making it public.  I did not shrink any constructor argument lists despite some arguments now being unused.  That will be fixed later in another check.


src/main/java/build/buildfarm/WorkerProfile.java
```diff
@@ -230,8 +230,7 @@ private static void workerProfile(String[] args) throws IOException {
     }
   }
 
-  private static void getWorkerProfile(Instance instance) {
-    // List<String> worker = instance.
+  public static void getWorkerProfile(Instance instance) {
```
werkt, line 233: Why public now?

src/main/java/build/buildfarm/cas/DigestMismatchException.java
```diff
@@ -19,15 +19,11 @@
 import java.io.IOException;
 
 public class DigestMismatchException extends IOException {
```
werkt, line 21: I'd rather make accessors for these than not retain them.

src/main/java/build/buildfarm/cas/MemoryCAS.java
```diff
@@ -73,7 +72,6 @@ public MemoryCAS(long maxSizeInBytes) {
   public MemoryCAS(
       long maxSizeInBytes, Consumer<Digest> onPut, ContentAddressableStorage delegate) {
     this.maxSizeInBytes = maxSizeInBytes;
```
werkt, line 74: We should definitely be calling this.

src/main/java/build/buildfarm/instance/shard/JedisCasWorkerMap.java
```diff
@@ -49,13 +48,6 @@
    */
   private final int keyExpiration_s;
 
-  /**
```
werkt, line 52: How did this even get in here without use?

src/main/java/build/buildfarm/instance/shard/RedissonCasWorkerMap.java
```diff
@@ -34,14 +34,6 @@
  */
 public class RedissonCasWorkerMap implements CasWorkerMap {
 
-  /**
```
werkt, line 37: Doesn't this being unused now mean the state the comment describes is invariant?

src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java
```diff
@@ -77,7 +76,6 @@ public ShardWorkerInstance(
       ShardWorkerInstanceConfig config)
```
werkt, line 76: Looks like it doesn't catch unused params...

werkt (commented):
The logger removals here are awkward and not my favorite thing. Some of the other stuff is really odd, and I don't even know what's going on the Redisson map.

src/main/java/build/buildfarm/cas/DigestMismatchException.java
luxe: fixed

src/main/java/build/buildfarm/instance/shard/JedisCasWorkerMap.java
```diff
@@ -22,7 +22,6 @@
 import java.io.IOException;
 import java.util.Map;
 import java.util.Set;
-import org.redisson.api.RSetMultimapCache;
```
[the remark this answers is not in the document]
luxe: There are two implementations of CasWorkerMap.  One uses jedis.  The other uses redisson and caching.

src/main/java/build/buildfarm/instance/shard/RedissonCasWorkerMap.java
luxe: The name is forwarded to the Redisson container which is used as part of the key names.  There is no need to additionally store the name outside that container.

src/main/java/build/buildfarm/cas/MemoryCAS.java
luxe: added usage of onPut

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
```diff
@@ -179,11 +177,6 @@
 
   public static final String INVALID_PLATFORM = "The `Platform` of the `Command` was invalid.";
 
-  private static final String INVALID_FILE_NAME =
```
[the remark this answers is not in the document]
luxe: Do you know what filename I should be validating with this error message?
`"One of the input PathNode's has an invalid name, such as a name containing a '/' character"`
I'm not sure what a PathNode is.  InputFiles are allowed to have a '/' character. What Action/Command component should I validate?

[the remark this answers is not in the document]
luxe: Oh I see.  There are some FIXMEs
```
      /* FIXME serverside validity check? regex?
      Preconditions.checkState(
          isValidFilename(fileName),
          INVALID_FILE_NAME);
      */
 ```
 
 Any recommendations on that?

luxe:
> The logger removals here are awkward and not my favorite thing. Some of the other stuff is really odd, and I don't even know what's going on the Redisson map.

The static analysis doesn't like the loggers because they are private and unused.  I'm not sure if there is a tag to let them stay despite that.  The redisson map also had an unused data member from a refactor I did.  It's actually pointing out a mistake I made which would have been nice to catch at the time.

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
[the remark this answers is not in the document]
luxe: I've preserved the `INVALID_FILE_NAME` by enabling one of the places where it was commented out.  This will prevent the code from being removed now.

werkt (approved):
My only hold was for the logging message for invalid filename. If you're sure, go for it.

Unrelated, we still need to remove the map count for O(N) scans to provide fast queue metrics that scale.

PR #723, opened 2021-03-15
Update jnr-fuse to 0.5.5. 0.5.1 is not available.

greggdonovan:


luxe:
I saw the dependency timeout as well.  I think it was temporarily down.  http 503.  But it working for me now. I'll test out the upgrade anyhow, and see if we can move it forward.

greggdonovan:
@luxe Ahh, you're right. [Bintray was down](https://status.bintray.com/). Given that JCenter appears to be shutting down on May 1st and 0.5.1 does not exist on [Maven Central](https://repo.maven.apache.org/maven2/com/github/serceman/jnr-fuse/), upgrading soon may be prudent. Thanks for the help!

luxe (approved):

PR #725, opened 2021-03-21
add java_image to utilize tini + docker patch for entry_point

luxe:
These java images were added in response to https://github.com/bazelbuild/bazel-buildfarm/issues/552.  
The linux sandbox is provided to the worker image so it may be used as an execution wrapper.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -68,6 +68,7 @@ java_image(
     main_class = "build.buildfarm.worker.shard.Worker",
     runtime_deps = [
         ":process-wrapper.binary",
+        ":linux-sandbox.binary",
         ":tini.binary",
         "//src/main/java/build/buildfarm/worker/shard",
     ],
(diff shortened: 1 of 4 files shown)


BUILD
```diff
@@ -38,3 +39,37 @@ cc_binary(
         "//conditions:default": ["as-nobody.c"],
     }),
 )
+
+
+# Docker images for buildfarm components
+java_image(
+    name = "buildfarm-server",
+    base = "@amazon_corretto_java_image_base//image",
```
werkt, line 47: Why the corretto image base? ubuntu isn't satisfactory?

luxe: @80degreeswest, do you know?

80degreeswest: We've been using it for about a year now. Here are the benefits: https://aws.amazon.com/corretto

80degreeswest (approved):

PR #726, opened 2021-03-24
initial support for ingesting Build Events [BES]

luxe:
To support the Build Event Protocol, we add the Build Event Service to the server.  
For more details about this protocol see: https://docs.bazel.build/versions/master/build-event-protocol.html

`googleapis` needed updated to match the API.
`BUILD.bazel` files were removed because of subpackage errors.

We will use BES information to improve buildfarm and eventually provide a frontend UI for users.


src/main/java/build/buildfarm/server/PublishBuildEventService.java
```diff
@@ -0,0 +1,70 @@
...
+
+      @Override
+      public void onCompleted() {
+        responseObserver.onCompleted();
+      }
+    };
+  }
+
+  private void recordEvent(PublishBuildToolEventStreamRequest in) {
+    // TODO(luxe)
```
80degreeswest, line 68: May be a good idea to add an option in config to turn this on/off. If it's turned on, then just output it to the log for now just so we can at least test it in dev. But can also be done in another diff as well if the intent here is just to provide the plumbing for it.

80degreeswest (approved):

PR #744, opened 2021-04-13
Suppress errors for unused return value of proto `.build()`

cushon:
Fixes

```
src/test/java/build/buildfarm/examples/ExampleConfigsTest.java:51: error: [ProtoBuilderReturnValueIgnored] Unnecessary call to proto's #build() method.  If you don't consume the return value of #build(), the result is discarded and the only effect is to verify that all required fields are set, which can be expressed more directly with #isInitialized().
      builder.build();
                   ^
    (see https://errorprone.info/bugpattern/ProtoBuilderReturnValueIgnored)
  Did you mean to remove this line or 'checkState(builder.isInitialized());'?
src/test/java/build/buildfarm/examples/ExampleConfigsTest.java:51: error: [ReturnValueIgnored] Return value of this method must be used
      builder.build();
                   ^
    (see https://errorprone.info/bugpattern/ReturnValueIgnored)
  Did you mean to remove this line?
```

### DIFF
--- a/src/test/java/build/buildfarm/examples/ExampleConfigsTest.java
+++ b/src/test/java/build/buildfarm/examples/ExampleConfigsTest.java
@@ -31,6 +31,7 @@ import org.junit.runners.JUnit4;
 
 // Test that example config files can load properly
 @RunWith(JUnit4.class)
+@SuppressWarnings({"ProtoBuilderReturnValueIgnored", "ReturnValueIgnored"})
 public class ExampleConfigsTest {
 
   @Before


luxe:
Is there a way we can catch these problems in CI?  I'm not sure what your build command was.

luxe (approved):

cushon:
> Is there a way we can catch these problems in CI? I'm not sure what your build command was.

It was caught in CI by an unreleased version of Bazel: https://github.com/bazelbuild/bazel/pull/13016#issuecomment-818847646

PR #746, opened 2021-04-14
[bug] log exceptions and avoid NPE in `EnrichedOperationBuilder`

luxe:
An `EnrichedOperation` is built up from the operation, action, and command information.  If fetching some of this information fails, it cannot be used to derive more information about the enriched operation.  This avoids the NPE reported in https://github.com/bazelbuild/bazel-buildfarm/issues/743.  
Additional logging is added to all caught exceptions.


src/main/java/build/buildfarm/operations/finder/EnrichedOperationBuilder.java
```diff
@@ -98,13 +114,15 @@ private static Operation jsonToOperation(String json) {
             .ignoringUnknownFields();
 
     if (json == null) {
+      logger.log(Level.SEVERE, "Operation Json is empty");
```
80degreeswest, line 117: Is this a critical error that will error out an operation and/or the build? Is this something that may be better logged as a warning?

src/main/java/build/buildfarm/operations/finder/EnrichedOperationBuilder.java
```diff
@@ -98,13 +114,15 @@ private static Operation jsonToOperation(String json) {
...
       return null;
     }
     try {
       Operation.Builder operationBuilder = Operation.newBuilder();
       operationParser.merge(json, operationBuilder);
       return operationBuilder.build();
     } catch (InvalidProtocolBufferException e) {
+      logger.log(Level.SEVERE, e.getMessage());
```
80degreeswest, line 125: We should just log "some error message", e here vs e.getMessage(). Same with below ones as well.
Ex: logger.log(Level.SEVERE, "InvalidProtocolBufferException while building an operation.", e);

80degreeswest (changes requested):

src/main/java/build/buildfarm/operations/finder/EnrichedOperationBuilder.java
luxe: For context,  this code is used when an individual is looking up operations with `//src/main/java/build/buildfarm:bf-find-operations`.  It won't run normally in production builds.  Switched this to warning.  Ideally anything that will also print to the user's console.

src/main/java/build/buildfarm/operations/finder/EnrichedOperationBuilder.java
luxe: yeah, that's better.  switched.

80degreeswest (approved):

PR #749, opened 2021-04-15
Fix problem with relative paths when running worker on windows

tobbe76:
ProcessBuilder on windows requires absolute paths for the executable or
relative CWD. ProcessBuilder.directory() only affects cwd of the proccess
to be created.

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -45,6 +45,7 @@ import com.google.rpc.Code;
 import io.grpc.Deadline;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


tobbe76:
@googlebot I signed it!

tobbe76:
@googlebot I fixed it

luxe:
thanks for the fix!  It makes sense to me, but I should probably consult bazel/remote api documentation to see what they have to say about it. I'll follow up on this.

@werkt what do you think?

werkt:
So this will break relative path discovery due to `argv[0]` interpretation for symlink based input paths. I'll insist that this is switched over to be a windows-specific behavior, and is not effective on other platforms - relative paths should stay relative there.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -200,7 +201,12 @@ private long executePolled(
           arguments.addAll(transformWrapper(policy.getWrapper()));
         }
       }
-      arguments.addAll(command.getArgumentsList());
+
+      // Make sure that the executable path is absolute, otherwise processbuilder fails on windows
+      Iterator<String> argumentItr = command.getArgumentsList().iterator();
+      String exe = argumentItr.next(); // Get first element, this is the executable
+      arguments.add(workingDirectory.resolve(exe).toAbsolutePath().normalize().toString());
+      argumentItr.forEachRemaining(arg -> arguments.add(arg));
```
werkt, line 209: Please change this to only be effective for absolute path on windows.

Also, it is a violation of iterator usage to call next() without calling hasNext() (and reacting to it)

luxe: windows only makes sense

tobbe76:
It makes sense to only do this on windows, I have update my patch for this.
I use "Win" as identifier for windows OS, I think that should be a fine solution.

luxe (approved):

PR #753, opened 2021-04-19
Fix access denied problem on windows during destroyExecDir

tobbe76:
Close the open filedescriptor after upload finished due to zero size, or it's already present in cache.
This is a must on windows, otherwise destroyExecdir will fail(access denied).

### DIFF
--- a/src/main/java/build/buildfarm/instance/stub/ByteStreamUploader.java
+++ b/src/main/java/build/buildfarm/instance/stub/ByteStreamUploader.java
@@ -393,6 +393,17 @@ public class ByteStreamUploader {
 
             @Override
             public void onClose(Status status, Metadata trailers) {
+              try {
+                // If upload was completed by someone else or already present, then chunker does
+                // does not close the Filehandle. Do it here to be sure it's always closed.
+                chunker.reset();
+              } catch (IOException e1) {
+                // This exception indicates that closing the underlying input stream failed.
+                // We don't expect this to ever happen, but don't want to swallow the exception
+                // completely.
+                logger.log(Level.WARNING, format("Chunker failed closing data source: %s", e1));
+              }
+
               if (status.isOk() || Code.ALREADY_EXISTS.equals(status.getCode())) {
                 uploadResult.set(null);
               } else {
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/instance/stub/ByteStreamUploader.java
```diff
@@ -393,6 +393,17 @@ public void onMessage(WriteResponse response) {
...
             public void onClose(Status status, Metadata trailers) {
+              try {
+                // If upload was completed by someone else or already present, then chunker does
+                // does not close the Filehandle. Do it here to be sure it's always closed.
+                chunker.reset();
+              } catch (IOException e1) {
+                // This exception indicates that closing the underlying input stream failed.
+                // We don't expect this to ever happen, but don't want to swallow the exception
+                // completely.
+                logger.log(Level.WARNING, format("Chunker failed closing data source: %s", e1));
```
werkt, line 404: Change `e1` to `e` here, and take it out of the format - we supply exceptions as the final parameter to `log` calls so that the handlers can deal with the traces as they see fit - this will not print a stack trace.

src/main/java/build/buildfarm/instance/stub/Chunker.java
```diff
@@ -175,6 +175,7 @@ public Chunk next() throws IOException {
     maybeInitialize();
 
     if (size == 0) {
+      data.close();
```
werkt, line 178: This should just call `reset()` and avoid the cleanup behaviors. It should be the only location that a close occurs, and it includes (and can function properly) regardless of the state of `data`

src/main/java/build/buildfarm/instance/stub/Chunker.java
```diff
@@ -183,6 +184,7 @@ public Chunk next() throws IOException {
     int bytesToRead = (int) Math.min(bytesLeft(), chunkSize);
     if (bytesToRead == 0) {
       chunkCache = null;
+      data.close();
       data = null;
```
werkt, line 188: replace with `reset()`

werkt (changes requested):
Logic is sound, minor nits

src/main/java/build/buildfarm/instance/stub/Chunker.java
tobbe76: Fixed

src/main/java/build/buildfarm/instance/stub/Chunker.java
tobbe76: Fixed

src/main/java/build/buildfarm/instance/stub/ByteStreamUploader.java
tobbe76: Updated

werkt (approved):

PR #755, opened 2021-04-20
[ci] add integration test for memory instances

luxe:
To avoid future regressions on the memory instance types we add an integration test similar to the existing one for shard instances.
A server and worker are started in a container and many cc files are built against them to ensure a working setup.  
This integration test would have prevented the recent regression: https://github.com/bazelbuild/bazel-buildfarm/issues/751


.bazelci/integration_test.sh
```diff
@@ -11,4 +11,4 @@ docker run -d --name buildfarm-redis -p 6379:6379 redis:5.0.9
 docker build -t buildfarm .
 
 # Start the servies and do a test build
-docker run --network host buildfarm buildfarm/.bazelci/test_buildfarm_container.sh
\ No newline at end of file
+docker run --network host --env TEST_SHARD=$TEST_SHARD buildfarm buildfarm/.bazelci/test_buildfarm_container.sh
```
jacobmou, line 14: I'm a little confused here. From the title for the pr, we want to add integration tests for memory instances right? But it seems only the shard instances are tested?

luxe: Yeah, the diff doesn't make it obvious what happened.
The existing shard integration test was automatically turned into the memory integration test.  

Then I re-added the shard specifically.  
So the `--env` is essentially `--env TEST_SHARD=true` and `--env TEST_SHARD=false` for the two instance types.  

I added `(Memory)` now to better distinguish that in  `.bazelci/presubmit.yml`

Here is Memory Test running:  
https://buildkite.com/bazel/buildfarm-male-farmer/builds/2476#03484212-5b0f-4a3d-840e-918e74e14789/95-1613

Here is Shard Test running:
https://buildkite.com/bazel/buildfarm-male-farmer/builds/2476#5ee02265-7d36-4bfa-a463-3774a1d2e185/95-1614

jacobmou: Ah, i see, didn't catch that on the first pass. Thanks for adding the test!

jacobmou (approved):

PR #757, opened 2021-04-21
ensure that cgroup cleanup does not hang indefinitely

luxe:
There appears to be two potential infinite loops during cgroups cleanup:  
In the controller: `while (exists) {...}`
In the group: `while (!killAllProcs(controllerName)) `

We leave the implementation as-is, but ensure that the closing of the controller does not exceed 10 seconds.
This means we will provide a best effort cleanup for cgroup procs, but not hang indefinitely in the above mentioned cases become stuck. 

Additional logging is added to see if these timeouts occur.

### DIFF
--- a/src/main/java/build/buildfarm/worker/cgroup/Controller.java
+++ b/src/main/java/build/buildfarm/worker/cgroup/Controller.java
@@ -22,9 +22,20 @@ import java.io.Reader;
 import java.io.Writer;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/cgroup/Controller.java
```diff
@@ -22,10 +22,21 @@
...
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 abstract class Controller implements IOResource {
   protected final Group group;
 
+  private static final Logger logger = Logger.getLogger(Controller.class.getName());
```
werkt, line 38: We seem to be using a lombok annotator for logging classes these days that might fit better here

werkt (approved):
Is there any way to get some tests on this?

werkt:
@luxe is this still relevant? Mostly I'm concerned that the future not terminating means that there will be random threads sticking around trying to kill off resistant processes, and that they will pile up.

PR #766, opened 2021-04-28
[configuration] organize prequeue and operation queue proto data

luxe:
This organizes the config files by composing some of the data.
The prequeue and operation queue are configured by name and max_depth.
The existing configuration values are moved into a prequeue and operationQueue config inside the backplane config.


80degreeswest:
I started the work to converting the configs to yaml. I'm hesitant to do any config changes at this point since most of this will become irrelevant in the not so distant future, especially since it's a breaking change.

PR #769, opened 2021-04-30
[metrics] Add metrics that explain currently dispatched work

luxe:
### Summary:
We don't have enough visibility into buildfarm's current workload.
We can spot queue spikes and see the number of dispatched operations, but we don't know how many users are actively using buildfarm and what kinds of tools and operations they are performing.
To solve this, we add additional metrics that provide a breakdown of buildfarm's current workload.

### Workload Breakdown:
These metrics help answer the following questions:
1. How much of the current workload is `build` vs. `test`?
2. How much of the current workload is `cpu` vs `gpu`?

We see queue spikes but this does not correlate to the amount of dispatched operations.  
Also the rate at which we instrument queue size, means we usually don't see the amount of operations hitting each queue.
Gpu operations need to run serially and will likely take longer than average cpu operations. 

3. Is anybody using a different version of `bazel` (or any build tool) with buildfarm?
4. What action `mnemonics` are running? (ex. CppCompile or GoLink)
5. How many unique `targets` are being fulfilled? (this will correlate to the total number of operations)
6. How many unique `configs` are being used? (debug, opt, etc.  I'm not sure these are populated)
7. How many unique clients are being served.  This lets us know how many people are using `--config=remote` at any one time.

This will let us know if the amount of work we are doing in relation to the amount of users we have.  Or if there is a single user hitting buildfarm with a lot of work at once.

8. Are any high core actions currently running?
9. What specific toolchains are being invoked? (mnemonics are often missing)

10. How many current operations have been `requeued` by the dispatch monitor?

This generally shouldn't happen, so it will be helpful to see how often it does.

### Testing:
I've played with this in the dev environment.  I don't see any performance concerns.  The metrics I've seen have only been for a small pool of workers.  I did full builds of the internal repo though.  So the content of the actions are illustrative of the kind of work we can expect.  It will be more interesting to see the metrics in prod, and then decide if we want to turn some off. 
Screenshot from 2021-05-29 21-49-08


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1419,19 +1432,245 @@ public boolean canPrequeue() throws IOException {
...
+    ActionAmounts actionAmounts = new ActionAmounts();
+    Integer requeuedOperationsAmount = 0;
+    Set<String> uniqueToolInvocationIds = Sets.newHashSet();
+    Map<String, Integer> fromQueueAmounts = new HashMap();
+    Map<String, Integer> toolAmounts = new HashMap();
+    Map<String, Integer> commandTools = new HashMap();
+    Map<String, Integer> actionMnemonics = new HashMap();
+    Map<String, Integer> targetIds = new HashMap();
+    Map<String, Integer> configIds = new HashMap();
+    Map<String, Integer> platformProperties = new HashMap();
```
80degreeswest, line 1468: Do we need to keep track of these if we only care about the size for metrics? Why don't we just set the counters directly using labels that the prometheus library provides, just like we do for queue sizes (cpu vs gpu).

80degreeswest (changes requested):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: You're suggesting we put Prometheus types into the backplane and populate them directly here (avoiding the maps)?  We could do that.  

Building up these maps is similar to what we already do for the cpu/gpu queue.  the backplane returns a  proto message, and the instance publishes it.  For queues (cpu/gpu), the backplane returns a `QueueStatus`.  For these metrics, the backplane returns a `DispatchedOperationsStatus`.

80degreeswest: Ok, I think this should work. Let's just keep it on the instance level.

80degreeswest (approved):

PR #773, opened 2021-05-05
80degreeswest/additional prometheus metrics

80degreeswest:
Add the following Prometheus metrics:

- expired_key
- actions
- operations_stage_load
- operation_status
- operation_worker
- remote_invocations
- operation_poller


luxe (approved):
These look good.  Maybe some of the counters could be labeled with the worker/server instance, so we know individual amounts and can `sum` for the whole cluster.  But let's just see what these look like first.

PR #779, opened 2021-05-10
prevent 0 cores from being chosen during 'global limit execution'

luxe:
prevent 0 cores from being chosen during 'global limit execution'

### DIFF
--- a/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
+++ b/src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
@@ -63,6 +63,11 @@ public class ResourceDecider {
       limits.cpu.max = 1;
     }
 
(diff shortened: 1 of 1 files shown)


80degreeswest (approved):

src/main/java/build/buildfarm/worker/resources/ResourceDecider.java
```diff
@@ -63,6 +63,11 @@ public static ResourceLimits decideResourceLimitations(
       limits.cpu.max = 1;
     }
 
+    if (limitGlobalExecution) {
+      limits.cpu.min = Math.max(limits.cpu.min, 1);
```
werkt, line 67: I think you want Math.min here and below. Also, I recall 0 being specified as the 'all available' option, where individual action limits weren't specified.

PR #784, opened 2021-05-11
skip execution policies when sandbox is chosen

luxe:
### Problem:
The `linux-sandbox` does not work with our internal configuration of worker "execution policies".
Our internal deployment of buildfarm configures 2 execution policies on the worker:  
They are `as-nobody` and `process-wrapper` (both bundled in the public java image).  

`as-nobody` does not have proper permissions to run under the sandbox.  We can not flip the order since execution policies are applied in the executor.  We can't remove the execution policy because choosing the sandbox is done dynamically and we don't want to juggle different buildfarm deployments.

### Solution:
If you are using the sandbox, don't use the execution policies.

### DIFF
--- a/src/main/java/build/buildfarm/common/ExecutionProperties.java
+++ b/src/main/java/build/buildfarm/common/ExecutionProperties.java
@@ -131,4 +131,26 @@ public class ExecutionProperties {
    *     with say different execution policies.
    */
   public static final String LINUX_SANDBOX = "linux-sandbox";
+
+  /**
+   * @field AS_NOBODY
+   * @brief The exec_property to inform the executor to run the action as a 'nobody' user.
+   * @details The "as nobody" functionality is supported by the bazel sandbox. This execution
+   *     property may be fulfilled through the sandbox or a standalone program. This execution
+   *     wrapper was previously used as a configured execution policy, but due to its involvement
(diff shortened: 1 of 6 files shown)


werkt:
Why wasn't sandboxing implemented as an execution policy? And why not just have some sandbox detection within the wrappers for as-nobody/process-wrapper and just not do them?

luxe:
> Why wasn't sandboxing implemented as an execution policy? And why not just have some sandbox detection within the wrappers for as-nobody/process-wrapper and just not do them?

Good point.  For testing the sandbox, we need something dynamic that we can switch on/off in the same buildfarm deployment (which is why the sandbox became an exec_property instead of being directly configured as an execution policy).  Turning the sandbox into an execution policy would be best thing long term- we just need the policies to be more flexible than what's directly written in a worker config.

I'd prefer something like this: if the sandbox is requested, the sandbox execution policy is added, and the other two policies are removed.  Since `cgexec` was already decided without being an execution policy, that's where it made sense to dynamically pick the sandbox, and that's how the implementation got bent this way.  

But it would be good to refactor things into this:
`(exec_properties + buildfarm config) => dynamic execution policies => executor`

80degreeswest (approved):

luxe:
We are going to test this out first.  I will follow up with George's advice on moving to execution policies.

PR #787, opened 2021-05-17
[bwob] Add S3 container for cloud-tier CAS

luxe:
We add an S3 container for uploading / downloading CAS files.  It obtains secrets through the AWS secrets manager (similar to the metrics publisher).  Configuration for this container is expressed in proto so it can later become part of the worker/server config.  
This does not change any current behavior- it simply adds the container for future use.


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -227,6 +227,24 @@ message MetricsConfig {
...
+  string access_key_id = 1;
+  string secret_key = 2;
+}
+
+message S3BucketConfig {
+
+  string region = 1;
+  
+  string name = 2;
+
```
werkt, line 241: Some funky spacing here...

src/main/java/build/buildfarm/common/s3/BUILD
```diff
@@ -0,0 +1,12 @@
...
+    srcs = ["S3Bucket.java"],
+    visibility = ["//visibility:public"],
+    deps = [
+        "//src/main/protobuf:build_buildfarm_v1test_buildfarm_java_proto",
+        "@maven//:com_amazonaws_aws_java_sdk_core",
+        "@maven//:com_amazonaws_aws_java_sdk_s3",
+        "@maven//:com_amazonaws_aws_java_sdk_secretsmanager",
+        "@maven//:com_fasterxml_jackson_core_jackson_databind",
+    ],
+)
```
werkt, line 12: Leaving this dangling without any usage, even a test, makes me a little worried for its health.

werkt (approved):
Hopefully that S3CAS implementation will be coming soon...

PR #795, opened 2021-05-28
add formatting rule (no whitespace after brackets)

luxe:
The current formatter is lenient on this, so we add an extra rule based on previous reviews.


jacobmou (approved):
Nice!

PR #796, opened 2021-05-31
add time travel to detect artificially slow actions and future time bombs

luxe:
### Summary
This addresses two problems in regards to an action's dependence on time.  The 1st problem is when an action takes longer than it should because it's sleeping unnecessarily.  The 2nd problem is when an action relies on time which causes it to eventually be broken on master despite the code not changing.  Both problems are expressed below as unit tests.  We demonstrate a time-spoofing mechanism (the re-writing of syscalls) which allows us to detect these problems generically over any action.  The objective is to analyze builds for performance inefficiency and discover future instabilities before they occur.

### Issue 1 (slow test)
```
#!/bin/bash
set -euo pipefail

echo -n "testing... "
sleep 10;
echo "done"
```
The test takes 10 seconds to run on average.
```
bazel test --runs_per_test=10 --config=remote //cloud/buildfarm:sleep_test
//cloud/buildfarm:sleep_test                                             PASSED in 10.2s
  Stats over 10 runs: max = 10.2s, min = 10.1s, avg = 10.2s, dev = 0.0s
```

We can check for performance improvements by using the `skip-sleep` option.  
```
bazel test --runs_per_test=10 --config=remote --remote_default_exec_properties='skip-sleep=true' //cloud/buildfarm:sleep_test
//cloud/buildfarm:sleep_test                                             PASSED in 1.0s
  Stats over 10 runs: max = 1.0s, min = 0.9s, avg = 1.0s, dev = 0.0s
```

Now the test is 10x faster.  If skipping sleep makes an action perform significantly faster without affecting its success rate, that would warrant further investigation into the action's implementation.

### Issue 2 (future failing test)
```
#!/bin/bash
set -euo pipefail

CURRENT_YEAR=$(date +"%Y")
if [[ "$CURRENT_YEAR" -eq "2021" ]]; then
        echo "The year matches."
        date
        exit 0;
fi;
echo "Times change."
date
exit -1;
```
The test passes today, but will it pass tomorrow?  Will it pass a year from now?  We can find out by using the `time-shift` option.
```
bazel test --test_output=streamed --remote_default_exec_properties='time-shift=31556952' --config=remote //cloud/buildfarm:future_fail
INFO: Found 1 test target...
Times change.
Mon Sep 25 18:31:09 UTC 2023
//cloud/buildfarm:future_fail                                            FAILED in 18.0s
```
Time is shifted to the year 2023 and the test now fails.  We can fix the problem before others see it.

### Notes:
Since these are syscalls, it is language agnostic and should work on any action.  syscalls captured are nanosleep, clock_nanosleep, select, poll, gettimeofday, clock_gettime, time.  Shifting backwards in time is not supported.  This only works on linux (dummy code is added for mac and windows in order to appease the CI)


werkt (approved):
So a lot of this is great, I think. So delay is meant to work in tandem with the skip_sleep wrapper to jump forward? Note that if they're not used together, delay will cause dramatic impact on the schedulability of an action (in the best case its expiration). Can you please write up documentation on all of these options (and the dangers of using just delay in isolation)?

luxe:
> So delay is meant to work in tandem with the skip_sleep wrapper to jump forward? Note that if they're not used together, delay will cause dramatic impact on the schedulability of an action (in the best case its expiration). 

Exactly.  Hopefully delay.sh can be removed in the future.  I didn't want two wrappers, but it was the easiest way to get it to work. Added docs and warnings about this.

PR #800, opened 2021-06-03
[executor] [sandbox] capture resource usage per action under the bazel sandbox.

luxe:
Bazel's sandbox has a [CLI option](https://github.com/bazelbuild/bazel/blob/master/src/main/tools/linux-sandbox-options.h#L46) for recording execution statistics.  These are gathered through [getrusage()](https://man7.org/linux/man-pages/man2/getrusage.2.html).  If the sandbox is chosen we will capture this additional data.

When debugging actions with buildfarm, we will now gather these statistics and report them back to the client.  Below is an example of what that looks like:

```
bazel test --test_output=streamed --remote_default_exec_properties='linux-sandbox=true' \
--remote_default_exec_properties='debug-after-execution=true' --config=remote  //:test
```

Results:  

```
    ...
    "resourceUsage_": {
      "utimeSec_": 0,
      "utimeUsec_": 28699,
      "stimeSec_": 0,
      "stimeUsec_": 9474,
      "maxrss_": 3940,
      "ixrss_": 0,
      "idrss_": 0,
      "isrss_": 0,
      "minflt_": 3695,
      "majflt_": 0,
      "nswap_": 0,
      "inblock_": 0,
      "oublock_": 0,
      "msgsnd_": 0,
      "msgrcv_": 0,
      "nsignals_": 0,
      "nvcsw_": 57,
      "nivcsw_": 0,
      ...
```

The test target I used was simple, so I'm expecting more interesting numbers depending on test analyzed.
[Buildbarn](https://github.com/buildbarn/bb-remote-execution/blob/2fa7b43a4f8f6b59ee60956466f270d013550877/pkg/runner/local_runner_unix.go#L193) also gathers rusage data (just not with the sandbox).


werkt (approved):
This would be *SICK* to add another execution control such that we get this data back in the ExecuteResult's Status debug information (or some container along those lines) now that I fixed bazels formerly jank interpretation of Status. Everybody else can be expected to make sense with their interpretation. Time to really improve the nature of what comes back from an an exec.

luxe:
> This would be _SICK_ to add another execution control such that we get this data back in the ExecuteResult's Status debug information (or some container along those lines) now that I fixed bazels formerly jank interpretation of Status. Everybody else can be expected to make sense with their interpretation. Time to really improve the nature of what comes back from an an exec.

That sounds awesome. We will eventually pair this data with BES and send it somewhere else as well.

PR #802, opened 2021-06-04
[configuration] add configuration to skip blacklist check

luxe:
In response to #801
Example configs keep the current behavior via `use_deny_list: True`


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -2563,4 +2567,11 @@ public void deregisterWorker(String workerName) {
       throw Status.fromThrowable(e).asRuntimeException();
     }
   }
+
+  private boolean isInDenyList(RequestMetadata requestMetadata) throws IOException {
```
werkt, line 2571: Prefer 'inDenyList' here

examples/shard-server.config.example
```diff
@@ -49,6 +49,10 @@ instances {
     # a fallback to non-remote recovery, rather than a catastrophic
     # failure.
     max_blob_size: 4294967296
+    
+    # Whether the instance should consult a denylist when looking up actions & invocations.
+    # For cache-only systems, users may want to disable this check.
```
werkt, line 54: Drop the comma

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -298,6 +298,7 @@
   private boolean stopping = false;
   private boolean stopped = true;
   private Thread prometheusMetricsThread;
+  private boolean useDenyList = true;
```
werkt, line 301: This can be final, no? Move it up if so.

werkt (approved):

src/main/java/build/buildfarm/instance/shard/ShardInstance.java

examples/shard-server.config.example
luxe: fixed

src/main/java/build/buildfarm/instance/shard/ShardInstance.java

PR #804, opened 2021-06-07
Gate code coverage at 40%

luxe:
We have been generating lcov files as part of the CI, however the overall code coverage process had some issues.
Empty dat files were causing `genhtml` to fail.  These are now ignored.  Hosting the static site failed because I don't have `http.server` package installed.  We switch this to the more common `SimpleHTTPServer`.  We also apply a gate to lcov results.  code coverage is gated at 40% for both line and function coverage.  They are currently 44% and 49%.  We will bump the gate over time.


werkt (approved):
Yes! Do it now! Get to the chopper!

PR #811, opened 2021-06-13
Update external repository instructions

gabware:
Add a call to compat_repositories() when configuring bazel-buildfarm as an external repository.
Without this bazel fails building the buildfarm-server with the following error:
```
ERROR: [...]/bzl-ws/external/io_grpc_grpc_java/compiler/BUILD.bazel:18:13: no such package '@com_google_guava_guava//jar': The repository '@com_google_guava_guava' could not be resolved and referenced by '@io_grpc_grpc_java//compiler:java_grpc_library_deps__do_not_reference'
```

TESTED=build and ran buildfarm-server and buildfarm-operationqueue-worker on windows host.


gabware:
This might be better handled in defs.bzl . WDYT ?
Happy to modify and test if that's a better way to do it.

werkt:
> 
> 
> This might be better handled in defs.bzl . WDYT ?
> Happy to modify and test if that's a better way to do it.

Yes, definitely - if it applies to all (and doesn't add yet another load -> initialize() layer) external consumers, then it deserves to be in defs.bzl.

luxe:
Since this is just documentation,  I'm going to land it as it may help others.  We can add it to defs.bzl and update docs after

luxe (approved):

PR #812, opened 2021-06-14
add "worker capabilities" for future support of storage workers

luxe:
We are implementing "storage workers"- which are workers that contribute to the sharded CAS, but do not participate in execution.  
To achieve this, we add a worker configuration for opting out of execution.  We've discussed simplifying these changes into a configuration called "capabilities".  Capabilities will be used to help us distinguish the type of workers we have.  
Here is how we would determine our current worker types:
| CAS      | Execution | Type |
| ----------- | ----------- | ----------- |
| true      | true       | Cpu      |
| false   | true        | Gpu       |
| true   | false        | Storage      |


examples/shard-worker.config.example
```diff
@@ -5,9 +5,16 @@ port: 8981
...
 public_name: "localhost:8981"
 
-# This determines whether the worker is a shard of the cas.
-# When not a shard, it will upload into the Sharded CAS
-omit_from_cas: false
+
+capabilities {
+  # This determines whether the worker is a shard of the cas.
+  # When not a shard, it will upload into the Sharded CAS
+  omit_from_cas: false
```
werkt, line 12: "it will upload execution outputs"

examples/shard-worker.config.example
```diff
@@ -5,9 +5,16 @@ port: 8981
 # by all shard cluster servers
 public_name: "localhost:8981"
 
-# This determines whether the worker is a shard of the cas.
-# When not a shard, it will upload into the Sharded CAS
-omit_from_cas: false
+
+capabilities {
```
werkt, line 9: I get the sensibility of omission flags, but given that we're rewriting this, let's make these positive assertions: 'cas', 'execution'

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -339,7 +339,7 @@ public Worker(String session, ServerBuilder<?> serverBuilder, ShardWorkerConfig
       throws ConfigurationException {
     super("BuildFarmShardWorker");
     this.config = config;
-    isCasShard = !config.getOmitFromCas();
+    isCasShard = !config.getCapabilities().getOmitFromCas();
```
werkt, line 342: That will help clean up these logics to be a little less awkward.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -576,6 +576,18 @@ message DequeueMatchSettings {
   build.bazel.remote.execution.v2.Platform platform = 3;
 }
 
+message WorkerCapabilities {
+  
+  // This determines whether the shard instance will participate in the backplane
+  // and cache it's operation outputs in its own cache file CAS.
+  // When omitted from CAS, the worker will send its output to another worker's CAS.
```
werkt, line 583: "send its execution outputs"

werkt (changes requested):
Did you drop the previous omit_from_cas definition? I wouldn't want both to be effective at once.

examples/shard-worker.config.example
luxe: fixed

examples/shard-worker.config.example
luxe: fixed

src/main/java/build/buildfarm/worker/shard/Worker.java
luxe: fixed

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: fixed

luxe:
> Did you drop the previous omit_from_cas definition? I wouldn't want both to be effective at once.

yes,  `omit_from_cas` will now be `capabilities::cas` (boolean meaning is flipped)

werkt (approved):

PR #817, opened 2021-06-18
Remove action cache counter and use Prometheus metric

80degreeswest:
Previously, the counter was needed as the log message was used for metrics and it was producing huge amounts of data. Since then, we have added Prometheus metric for this data point, rendering the counter unnecessary. This change removes the counter.


luxe (approved):
great!

PR #818, opened 2021-06-23
[bugfix] dispatched operations show incorrect requeue attempt

luxe:
### Summary
Each operation holds metadata about how many times it has been "re-queued".  Generally speaking, we don't expect any operations to be re-queued.  This is a robustness feature for ensuring that the cluster doesn't lose operations.  It is therefore a useful metric for us in identifying issues.  Currently the dispatched operations show an incorrect requeue attempt.

### Fix
All the dispatched operations say their requeueAttempt is 1.  This is misleading because we increment the value as part of the dispatching.  I would expect it to be 0 for most dispatched operations.  We adjust the increment logic to be performed after the dispatch.


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1255,7 +1254,11 @@ private QueueEntry dispatchOperation(JedisCluster jedis, List<Platform.Property>
                 operationName, operationQueue.getDequeueName()));
       }
       dispatchedOperations.remove(jedis, operationName);
-      return queueEntry;
+
+      // Return an entry so that if it needs re-queued, it will have the correct "requeue attempts".
+      QueueEntry entryForRequeue =
+          queueEntryBuilder.setRequeueAttempts(queueEntryBuilder.getRequeueAttempts() + 1).build();
```
werkt, line 1260: Generally considered bad practice to get this value out of the builder. I would also ask, given that this logic was wrong, that you write a test to ensure that for established transitions (i.e. queued -> dispatched -> expiration -> requeue) that you see the value go from 0, to 1, to 2.

werkt (changes requested):

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
luxe: good idea.  I tested this from a backplane API

80degreeswest (approved):

luxe:
@werkt ready for re-review

luxe:
@werkt bump

src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
```diff
@@ -178,6 +183,137 @@ public void requeueDispatchedOperationQueuesAndPublishes() throws IOException {
...
+            });
+
+    // ACT
+    // dispatch the operation and test properties of the QueueEntry and internal jedis calls.
+    QueueEntry readyForRequeue = null;
+    try {
+      List<Platform.Property> properties = new ArrayList<>();
+      readyForRequeue = backplane.dispatchOperation(properties);
+    } catch (InterruptedException e) {
+    }
```
werkt, line 245: This and the interrupted exception block below seem unnecessary. Let the throw happen, since the signatures already allow for it.

werkt (approved):
One minor nit.

werkt:
> @werkt bump

protip, github wouldn't let me update a review for this until I clicked 're-request review' for me. Click that after addressing comments in the future.

PR #820, opened 2021-06-25
fix linux aarch64

kekxv:


werkt (commented):
Why did you include these docker definitions?

kekxv:
@werkt docker definitions were added by mistake,Only `bazelw` needs to be updated

image

kekxv:
@werkt update https://github.com/bazelbuild/bazel-buildfarm/pull/882

PR #847, opened 2021-07-08
Initial Commit of Buildfarm Admin

80degreeswest:
Initial commit of buildfarm admin open sourced by Aurora. Follow up commits will be needed to provide for more documentation, testing, and general improvements to make the app more generic for everyone's use. Initial commit is not built with bazel, but utilizes maven.


luxe:
Is maven able to pull in the contents for `admin/main/src/main/resources/static/assets/vendor/`?  
Ideally we don't want to land jquery, fontawesome, bootstrap if it can be avoided.

80degreeswest:
> Is maven able to pull in the contents for `admin/main/src/main/resources/static/assets/vendor/`?
> Ideally we don't want to land jquery, fontawesome, bootstrap, etc if it can be avoided.

Good point. The templated included those in source but yeah should be able to just pull them from the public repo.

luxe (approved):
I'm not sure how the licenses should be, but maybe it will need to match the bazel license used elsewhere.  
Everything else seems good.  I'm less concerned about any of the code since we just want to get it into the repo first.

admin/main/src/main/resources/templates/fragments/template.html
```diff
@@ -0,0 +1,221 @@
...
+							<span class="mr-2 d-none d-lg-inline text-gray-600 small"></span>
+							<img class="img-profile rounded-circle" th:src="@{/assets/img/thisisfine.jpg}">
+						</a>
+						<!-- Dropdown - User Information -->
+						<div class="dropdown-menu dropdown-menu-right shadow animated--grow-in" aria-labelledby="userDropdown">
+							<a class="dropdown-item" href="http://buildfarmadmin-dev.uberinternal.com">
+								<i class="fas fa-tag fa-sm fa-fw mr-2 text-gray-400"></i>
+								BF Admin Dev
+							</a>
+							<a class="dropdown-item" href="http://buildfarmadmin.uberinternal.com">
```
luxe, line 102: just checking for any traces of "uber"

80degreeswest: Good catch! Removed.

PR #869, opened 2021-07-23
[docs] use github pages for hosting documentation

luxe:
There are a few issues with our existing wiki documentation:  

**1. Changes cannot be made via PR**
This prohibits public contributors from suggesting changes.  They usually just make issues saying that the docs are wrong (instead of a PR to fix it). Also, documentation is either updated before the change is landed, or there is a period of time where docs are incorrect after the PR lands. Better to update docs alongside code.

**2. Documentation is not released with buildfarm versions**
Depending on what buildfarm release someone is using the documentation may be incorrect.  It would be better for documentation to be cut on the same branch as our releases.  It is not easy for users to determine what has changed in the wiki, but if it's in the repo, they can more easily look at commit history. 

**3. Information is difficult to nest**
The wiki doesn't let you organize content by nesting directories.  That is why our sidebar is > 20 pages that get alphabetized.  We could build our own navigation in the wiki and link to the pages correctly, but that is unnecessary maintenance since github pages can generate correct navigation based on markdown layout.

**Solution:**
Documentation remains in markdown but it is moved to `_site/docs/`.  
We add a github hook to automatically publish it when main changes.  

Here is a sample of what that will look like from using a "just the docs" jekyll template:  
https://luxe.github.io/buildfarm/

**Follow-up:**
A github token needs added for the workflow hook.  I tested this on my fork and it was easy to setup.  For this repository, I think an admin will need to add it.  Also, THESE DOCS ARE OUTDATED.  Once a publishing hook works, a follow-up PR can update the content here before deprecating the wiki.


werkt (approved):
Not being able to see the content here, I have to wonder a) how out of date it is at this point, and where we're going to make our cutoff, and b) what it looks like in terms of presentation. That said, should be all good

_site/docs/tools/bf-cat.md
```diff
@@ -0,0 +1,30 @@
...
+
+`bf-cat <host[:port]> <instance-name> <hash-function> <command> [params...]`
+
+**instance-name** is the name of the specific instance to inquire about, typically configured on schedulers. A literal empty string parameter (i.e. bash: `""`) will use the default instance for a server.
+
+**hash-function** is one of MD5, SHA1, SHA256, etc selected to match the supported digest functions of an instance, and used to compute digests for content retrieved.
+
+**command** is typically one of the following, with digest parameters as <hash>/<size>, as typically represented in log entries:
+
+* **Action <digest...>**: Retrieves Action definitions from the CAS and renders them with field identifiers.
```
bshashank, line 20: Looking at the generated doc: https://luxe.github.io/buildfarm/docs/tools/bf-cat/

> ****Action** **: Retrieves Action definitions

The <digest ...> part is completely swallowed up somewhere.

_site/_config.yml
```diff
@@ -0,0 +1,124 @@
...
+gh_edit_branch: "gh-pages" # the branch that your docs is served from
+# gh_edit_source: docs # the source that your files originate from
+gh_edit_view_mode: "edit" # "tree" or "edit" if you want the user to jump into the editor immediately
+
+# Color scheme currently only supports "dark", "light"/nil (default), or a custom scheme that you define
+color_scheme: light
+
+# Google Analytics Tracking (optional)
+# e.g, UA-1234567-89
+ga_tracking: UA-2709176-10
```
bshashank, line 105: Who is tracking the analytics? (or did this come with the template?) If so we can probably remove to avoid tracking (if that's unnecessary).

If that's necessary, then perhaps we should list who the owner is here.

_site/_sass/color_schemes/light.scss
```diff
@@ -0,0 +1,22 @@
+$sidebar-color: #43a047;
+$border-color: #43a047;
```
bshashank, line 2: nit: feel free to ignore

I think this border color causes all the code-blocks to also get the green border. IMHO this interrupts with fast reading. 

Compare [generated action-cache](https://luxe.github.io/buildfarm/docs/architecture/action_cache/) with [wiki action-cache](https://github.com/bazelbuild/bazel-buildfarm/wiki/ActionCache)

bshashank (commented):
+1 to the idea. A couple of questions/comments:

luxe:
> a) how out of date it is at this point, and where we're going to make our cutoff

its very out-of-date.  I'd like to first see the website publish on each commit to main. (make sure that the github hook is working first).  We won't advertise these docs until the content is correct and we decide a cuttoff period for the wiki.  There would be a follow-up diff updating all the docs to be in-sync with wiki.

> b) what it looks like in terms of presentation. That said, should be all good

Presentation is okay. Mobile friendly.  There are some css styles worth changing. Should be easy to improve going forward

_site/docs/tools/bf-cat.md
luxe: it seems that markdown will interpret them as tags which become invalid html: `<digest></digest>`.  We will need to either annotate them as code, or escape them.

_site/_config.yml
luxe: good catch, its from the template.
removed.

_site/_sass/color_schemes/light.scss
luxe: haha yeah that's too much green. we can give get rid of the code border or make it black

luxe: fixed

_site/docs/tools/bf-cat.md
luxe: fixed

PR #870, opened 2021-07-23
[executor] tag action resources with optional container information

luxe:
In order to support running individual actions under client specified containers,  
we first attach container information to the action's resources.

The properties are derived from https://github.com/bazelbuild/bazel-toolchains and is how platforms are created via `create_rbe_exec_properties_dict`.

This is for issue https://github.com/bazelbuild/bazel-buildfarm/issues/867

### DIFF
--- a/src/main/java/build/buildfarm/worker/resources/ContainerSettings.java
+++ b/src/main/java/build/buildfarm/worker/resources/ContainerSettings.java
@@ -44,7 +44,7 @@ public class ContainerSettings {
   /**
    * @field network
    * @brief Docker network.
-   * @details Whether or not the contianer should allow network (standard, off).
+   * @details Whether or not the container should allow network (standard, off).
    */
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/worker/resources/ContainerSettings.java
```diff
@@ -0,0 +1,108 @@
...
+   * @field addCapabilities
+   * @brief Docker add capabilities.
+   * @details Allows adding capabilities through --cap-add.
+   */
+  public ArrayList<String> addCapabilities = new ArrayList<>();
+
+  /**
+   * @field network
+   * @brief Docker network.
+   * @details Whether or not the contianer should allow network (standard, off).
```
werkt, line 47: typo

werkt (approved):
urandom all the things. But seriously, don't :D

PR #872, opened 2021-07-26
[ci] extend formatter to support protobuf

luxe:
protobuf files can be formatted with `clang-format`.  We extend `format.sh` to support this.  The gate is currently turned off.  We can turn it on later; I didn't wanted to format all of the protobuf files as part of the integration.

### DIFF
--- a/.bazelci/format.sh
+++ b/.bazelci/format.sh
@@ -7,7 +7,9 @@ FORMAT_JAVA=true
 JAVA_FORMATTER_URL=https://github.com/google/google-java-format/releases/download/google-java-format-1.7/google-java-format-1.7-all-deps.jar
 LOCAL_FORMATTER="java_formatter.jar"
 
-FORMAT_PROTO=false
+FORMAT_PROTO=true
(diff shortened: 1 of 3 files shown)


werkt (commented):
So the only way we can get a proto formatter is with clang-format, and we have to incorporate the entirety of an llvm_toolchain to do it? Seems pretty heavy handed...

luxe:
> So the only way we can get a proto formatter is with clang-format, and we have to incorporate the entirety of an llvm_toolchain to do it? Seems pretty heavy handed...

agreed.  I wish llvm hosted clang-format individually.  this seems to be what buildbarn is doing as well:
https://github.com/buildbarn/bb-remote-execution/runs/3243994208#step:8:1

werkt (approved):
This is fine for now, but really, I'd rather we host/point to clang-format as an architecture dependent binary somewhere and use that. We don't want to depend upon anything but released versions, and I'd prefer we not be in the business of building this.

luxe:
sounds good.  For now, the formatter skips using this.  bazel won't fetch it since no target depends on it.  We'll look for something more lightweight.

PR #873, opened 2021-07-28
[cfc] create CAS fallback handler

luxe:
We want to make it easier to configure how the CAS fallback is used.  
See issue: https://github.com/bazelbuild/bazel-buildfarm/issues/859

By isolating the current delegate behavior, this will make it easier for us to configure.
Not all delegate behavior was pulled out of CasFileCache due to its dependence on other code.


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -459,16 +459,16 @@ boolean containsLocal(
...
 
   @Override
   public boolean contains(Digest digest, Digest.Builder result) {
-    return containsLocal(digest, result, (key) -> accessed(ImmutableList.of(key)))
-        || (delegate != null && delegate.contains(digest, result));
+    if (containsLocal(digest, result, (key) -> accessed(ImmutableList.of(key)))) {
+      return true;
+    }
+
+    return CasFallbackHandler.contains(delegate, digest, result);
```
werkt, line 471: feel free to do this, but leave the original logic. single return and all.

src/main/java/build/buildfarm/cas/cfc/CasFallbackHandler.java
```diff
@@ -0,0 +1,105 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
```
werkt, line 1: Prefer 'Delegate' instead of 'Fallback'

src/main/java/build/buildfarm/cas/cfc/CasFallbackHandler.java
```diff
@@ -0,0 +1,105 @@
...
+   * @param result The digest results to be populated.
+   * @return Whether the digest was found.
+   * @note Suggested return identifier: found.
+   */
+  public static boolean contains(
+      ContentAddressableStorage delegate, Digest digest, Digest.Builder result) {
+    if (delegate == null) {
+      return false;
+    }
+    return delegate.contains(digest, result);
```
werkt, line 103: Still prefer `return delegate != null && delegate.contains(digest, result);` to this.

werkt (changes requested):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java

src/main/java/build/buildfarm/cas/cfc/CasFallbackHandler.java
luxe: fixed

src/main/java/build/buildfarm/cas/cfc/CasFallbackHandler.java
luxe: fixed

luxe:
thx for the review.  I wanted to isolate the delegate functionality so it was easier for some to add configuration for "readonly" behavior.

PR #875, opened 2021-07-28
[executor] imply sandbox automatically for block-network & tmpfs

luxe:
If you want to enable block-network or tmpfs on an action, you need to also enable the sandbox.  
This is not obvious.  If you only enable tmpfs without enabling the sandbox its basically a no op.  
Let's honor the user's execution constraint by enabling the sandbox as needed.

This boils down to the following.  
Instead of doing something like this:
```
--remote_default_exec_properties='linux-sandbox=true'
--remote_default_exec_properties='block-network=true'
```

or 
```
exec_properties = {"linux-sandbox": "True", "block-network": "True"},
```

You will only need to do this:  
```
--remote_default_exec_properties='block-network=true'
```

or 
```
exec_properties = {"block-network": "True"},
```

sandbox will be implied for you.  
sandbox is still disabled by default because block-network and tmpfs are also disabled by default.


werkt (approved):
Looks good. Curious where the description actually gets surfaced - do we inform the client with extra debug info on the Operation com.google.rpc.Status response?

luxe:
> Looks good. Curious where the description actually gets surfaced - do we inform the client with extra debug info on the Operation com.google.rpc.Status response?

If a debugging exec_property is given, the executor will fail the operation before or after execution.  It sends the debugging info back on the operation's stderr.  This is not an ideal way to surface, but it has been used as a way to debug individual actions.  In the future, it would be nice to see in some kind of UI.

PR #877, opened 2021-08-02
[config] apply useful defaults for queue selection in shard configs

luxe:
I've seen various users have issues with their queue configuration.  In particular, they want to use additional exec_properties but this causes their actions to longer be eligible to the operation queue during matching.  I think these config changes would provide a better default for everyone.  The idea of exactly matching a queue no longer seems like a good default, and setting `allow_unmatched` will use a superset of properties that match.

Additionally, `allow_everything` during dequeue is also a safe default since the worker has already matched to a queue.


luxe:
This already landed in https://github.com/bazelbuild/bazel-buildfarm/pull/987

PR #886, opened 2021-08-11
[data structures] add a redisson map

luxe:
We currently have `RedisMap` which uses Jedis.  This map has a few downsides in that it doesn't do caching outside of redis and some calculations (like size)  are O(n) instead of O(1).  Let's add a redission version so we can more easily experiment swapping the two containers and seeing how they perform at a larger scale.


werkt (approved):
Go for this, obviously benign as is.

luxe:
thx, this will make the usage diff smaller

PR #887, opened 2021-08-12
Add TLS support for buildfarm server

krisstakos:
It was added an optional TLS support for BuildFarm server. If a ssl_certificate_path is populated with path to a PEM file, the code will use a gRPC**s** otherwise it will default to gRPC (plain-text). It was added a new field in the .proto file to accommodate the new changes.

### DIFF
--- a/src/main/java/build/buildfarm/server/BuildFarmServer.java
+++ b/src/main/java/build/buildfarm/server/BuildFarmServer.java
@@ -1,5 +1,5 @@
-//“Copyright 2021 VMware, Inc.”
-//SPDX-License-Identifier: Apache-2.0
+// “Copyright 2021 VMware, Inc.”
+// SPDX-License-Identifier: Apache-2.0
 // Copyright 2017 The Bazel Authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


luxe (approved):

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -1,3 +1,5 @@
+// “Copyright 2021 VMware, Inc.”
+// SPDX-License-Identifier: Apache-2.0
```
80degreeswest, line 2: Remove additional copyright header.

werkt: Yeah, strong negative here.

bshashank: Our employer requires us to retain the copyright for code added (apparently even small diffs to existing files).

Would it be acceptable to add the employer (VMware) to the AUTHORS file? Similar to https://github.com/bazelbuild/bazel-buildfarm/pull/844?

80degreeswest: I think that would acceptable, especially since VMware will be a regular contributor to the project.

krisstakos: Removed the copyright stuff and added `vmware` in authors file.

80degreeswest (approved):

PR #889, opened 2021-08-17
[breaking] [config] Remove support for multiple instance types in config

80degreeswest:
Simplify configuration to only allow a single instance.


luxe (approved):
this seems like a nice simplification.  I don't really understand how multiplexing instance types would work anyway.  And didn't even know it was a thing until your diff.

PR #890, opened 2021-08-17
80degreeswest/code analyzer cleanup

80degreeswest:
Run Intellij code analyzer to fix and cleanup various code issues built up over the years. Some of the fixes could address some performance issues, such as high memory utilization as the analyzer did flag those code blocks, which frequenly show up as problematic in our application profiler.


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -1458,7 +1442,7 @@ private void processRootFile(
           });
     }
 
-    joinThreads(pool, "Populating Directories...", 1, MINUTES);
```
luxe, line 1461: why hard-code this further down the callstack?

src/main/java/build/buildfarm/common/io/LibC.java
```diff
@@ -25,7 +25,7 @@
 
   Pointer opendir(String s);
 
-  int closedir(Pointer dir);
+  void closedir(Pointer dir);
```
luxe, line 28: not sure about making this void because JNI was used to match with the underlying [libc api](https://man7.org/linux/man-pages/man3/closedir.3.html).  Depending on how that binding works, this could be an issue.

src/main/java/build/buildfarm/instance/queues/WorkerQueues.java
```diff
@@ -90,24 +90,21 @@ public boolean enqueueOperation(Operation operation, SetMultimap<String, String>
...
 
-  public boolean AddWorker(SetMultimap<String, String> provisions, MatchListener listener) {
+  public void AddWorker(SetMultimap<String, String> provisions, MatchListener listener) {
     try {
       WorkerQueue queue = MatchEligibleQueue(provisions);
       queue.workers.add(new Worker(provisions, listener));
     } catch (Exception e) {
-      return false;
     }
-    return true;
```
luxe, line 100: I guess its okay to get rid of the bool if the callers aren't using it.. Might be better to make the callers check the return instead and log failures

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -24,37 +24,15 @@
 import static com.google.common.util.concurrent.MoreExecutors.shutdownAndAwaitTermination;
 import static java.lang.String.format;
 
-import build.bazel.remote.execution.v2.ActionCacheGrpc;
+import build.bazel.remote.execution.v2.*;
```
luxe: i guess it does clean up a lot, okay

src/main/java/build/buildfarm/backplane/Backplane.java
```diff
@@ -79,7 +79,7 @@ public ActionCacheScanResult(
    * <p>Returns true if the worker was newly added, and false if it was already a member of the set.
    */
   @ThreadSafe
-  boolean addWorker(ShardWorker shardWorker) throws IOException;
+  void addWorker(ShardWorker shardWorker) throws IOException;
```
luxe, line 82: thinking clients should just start checking this for failures to add

luxe (approved):
AWESOME! especially getting rid of exceptions

looked at each file :sleepy:,  It all looks harmless.  Summary of observations:
- remove unused params
- remove throw clauses (not sure about InterruptedException? Can't that happen anywhere?)
- remove `@link` (came from other codebases most likely)
- remove unnecessary blocks `-> {}`
- Long -> long
- `Collections.sort` instead of `.sort`
- `::` instead of `.`
-  marking things final
-  string / print changes
- using algos (remove_if / merge / etc)

Is the intellij scan paid for?  Wondering if they have a free CLI tool that you could add to CI at some point.

src/main/java/build/buildfarm/cas/GrpcCAS.java
```diff
@@ -19,14 +19,10 @@
 import static com.google.common.util.concurrent.Futures.transform;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 
-import build.bazel.remote.execution.v2.BatchReadBlobsRequest;
+import build.bazel.remote.execution.v2.*;
```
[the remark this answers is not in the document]
80degreeswest: I'll rolle some of these back. I didn't want it to auto replace with * and thought I reverted all of them but I guess missed a few.

80degreeswest:
> AWESOME! especially getting rid of exceptions
> 
> looked at each file 😪, It all looks harmless. Summary of observations:
> 
> * remove unused params
> * remove throw clauses (not sure about InterruptedException? Can't that happen anywhere?)
> * remove `@link` (came from other codebases most likely)
> * remove unnecessary blocks `-> {}`
> * Long -> long
> * `Collections.sort` instead of `.sort`
> * `::` instead of `.`
> * marking things final
> * string / print changes
> * using algos (remove_if / merge / etc)
> 
> Is the intellij scan paid for? Wondering if they have a free CLI tool that you could add to CI at some point.

It's part of the IDE. It suggests a lot of stuff and some of it we probably don't want to do so I did selective fixes here. But there are other tools we could integrate into CI. We should discuss at our next meeting.

src/main/java/build/buildfarm/cas/GrpcCAS.java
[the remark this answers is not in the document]
80degreeswest: Oh yeah you must have reviewed it prior to the latest commit that fixed it.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
80degreeswest: Just to remove duplication and put common code (time and unit in this case) into the common method.

src/main/java/build/buildfarm/common/io/LibC.java
80degreeswest: I'll roll this back just to be safe.

src/main/java/build/buildfarm/backplane/Backplane.java
bshashank: (sorry for drive-by comment)
Please also update the comment in line 79.

PR #891, opened 2021-08-18
Additional code clean up as a result of Intellij code analyzer

80degreeswest:
Clean up various minor code issues built up over the years using results from Intellij Idea code analyzer. This is a round 2 cleanups that required a more thorough review.
Add suppress annotations to issues found which will not be addressed at this time.


src/main/java/build/buildfarm/server/ExecutionService.java
```diff
@@ -69,12 +69,12 @@ private void withCancellation(
         future,
         new FutureCallback<Void>() {
           boolean isCancelled() {
-            return serverCallStreamObserver.isCancelled() || Context.current().isCancelled();
+            return !serverCallStreamObserver.isCancelled() && !Context.current().isCancelled();
```
luxe, line 72: isn't expression now testing is `isNotCancelled`? maybe change method:  You could change to `isActive`? I'm guessing that static analysis just wanted to avoid `if (!expressions)`?

src/main/java/build/buildfarm/proxy/http/HttpBlobStore.java
```diff
@@ -374,10 +374,10 @@ private void releaseDownloadChannel(Channel ch) {
...
 
   private boolean isChannelPipelineEmpty(ChannelPipeline pipeline) {
-    return (pipeline.first() == null)
-        || (useTls
-            && "ssl-handler".equals(pipeline.firstContext().name())
-            && pipeline.first() == pipeline.last());
+    return (pipeline.first() != null)
+        && (!useTls
+            || !"ssl-handler".equals(pipeline.firstContext().name())
+            || pipeline.first() != pipeline.last());
```
luxe, line 380: function name needs changed.  maybe `isChannelFull`..  something like that

src/main/java/build/buildfarm/worker/memory/OperationQueueClient.java
```diff
@@ -55,7 +55,7 @@ void match(MatchListener listener) throws InterruptedException {
 
           @Override
           public boolean getMatched() {
-            return matched;
+            return !matched;
```
luxe, line 58: dont think these are a good idea.  we would need to change the function name to is getUnmatched so it makes sense,  
but someone could along and then use `if (!x.getUnmatched())`.  So I'm not really sure what the analyzer is trying to achieve.

luxe (commented):
the best part of this diff is getting rid of unused parameters.  We have a PMD check for this in the CI that's disabled.  So we can probably start enforcing it after your diff lands.

All the suppressions are kind of annoying, but I don't mind sprinkling them everywhere if it makes intellij happy.  And they hint at things that we should probably just fix so nothing needs suppressed in the first place.  So I don't mind any of them but since its not enforced I'll likely cause more issues like that.

See my comments on the expression changes.  If we decide what to do about those, I'll be good to land. 

or CI, Those static analysis issues should be easy to fix.  There is `./bazelw run //:buildifier` available to fix that other issue.

src/main/java/build/buildfarm/proxy/http/HttpBlobStore.java
80degreeswest: Just going to revert here.

src/main/java/build/buildfarm/server/ExecutionService.java
80degreeswest: Will revert this one.

luxe (approved):

PR #892, opened 2021-08-19
[metrics] provide metric for CAS size and entry count

luxe:
We need to understand how the CAS is growing in both physical storage and entry amount.  This adds a metric thread to the `CASFileCache` to record those metrics.

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -128,7 +128,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
   private static final Gauge casSizeMetric =
       Gauge.build().name("cas_size").help("CAS size.").register();
   private static final Gauge casEntryCountMetric =
-      Gauge.build().name("cas_entry_count").help("number of entries in the CAS.").register();
+      Gauge.build().name("cas_entry_count").help("Number of entries in the CAS.").register();
 
   protected static final String DEFAULT_DIRECTORIES_INDEX_NAME = "directories.sqlite";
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -1213,6 +1218,26 @@ public StartupCacheResults start(
...
 
+    // Start metrics collection thread
+    prometheusMetricsThread =
+        new Thread(
+            () -> {
+              while (!Thread.currentThread().isInterrupted()) {
+                try {
+                  casSizeMetric.set(size());
+                  casEntryCountMetric.set(entryCount());
+                  TimeUnit.SECONDS.sleep(30);
```
80degreeswest, line 1229: I think since the change here will be fairly gradual we probably don't need to have this push every 30 seconds. I would say 1 or maybe even 5 minutes would be more than sufficient.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -124,6 +125,10 @@
   // Prometheus metrics
   private static final Counter expiredKeyCounter =
       Counter.build().name("expired_key").help("Number of key expirations.").register();
+  private static final Gauge casSizeMetric =
+      Gauge.build().name("cas_size").help("CAS size.").register();
+  private static final Gauge casEntryCountMetric =
+      Gauge.build().name("cas_entry_count").help("number of entries in the CAS.").register();
```
80degreeswest, line 131: Capitalize 'n' in the help string.

80degreeswest (approved):

PR #893, opened 2021-08-19
[static analysis burndown] enable gates from previous intellij fixes

luxe:
Here are some static analysis gates we can now enable from the recent intellij fixes (#890, #891)


80degreeswest (approved):
Nice!

PR #896, opened 2021-08-24
Make maxRequeueAttempts configurable

80degreeswest:
Current default is hardcoded at 5. This could present a problem with very long timeout times, especially since bazel's default remote retry count is 5. Make this configurable to allow for more flexibility with addressing hung builds.


luxe:
can you change the example config files to be 5

80degreeswest:
> can you change the example config files to be 5

done.

luxe (approved):

PR #916, opened 2021-09-23
Rework get uptime service in Buildfarm and BF Admin

jiaquan1:


80degreeswest (changes requested):
This does not address the issue as we are still making a separate call for each host. We should be making a single call to buildfarm, which then returns all uptimes for all hosts.

PR #920, opened 2021-09-29
[bfadmin] Add basic API security

80degreeswest:
Add basic API security utilizing a security token in the http header.


admin/main/src/main/java/tech/aurora/bfadmin/controller/AdminController.java
```diff
@@ -43,22 +48,39 @@ public String getMainApp() {
...
-    return "dashboard";
+    if (ui) {
+      generateSecurityKey();
+      model.addAttribute("clusterInfo", clusterInfo);
+      model.addAttribute("clusterDetails", adminService.getClusterDetails());
+      model.addAttribute("awsRegion", region);
+      model.addAttribute("securityKey", SecurityConfig.securityKey);
+      return "dashboard";
+    } else {
+      model.addAttribute("status", "999");
```
luxe, line 59: is that http status?  you could use a common one

luxe (approved):

admin/main/src/main/java/tech/aurora/bfadmin/controller/AdminController.java
80degreeswest: Nope just a custom error code/message set to be displayed on the UI. It's arbitrary.

PR #942, opened 2021-10-30
[refactor] delete instantiations of BuildFarmInstances

luxe:
### Summary:
`BuildFarmInstances` is a wrapper around `Instance`.  We are going to start passing around the `Instance` directly.  This will avoid the indirection (see removal of get() requests and try/catch blocks).  We also delete `InstanceNotFoundException` as it can't happen now. 

### For Review:
To best understand the change, first look at `src/main/java/build/buildfarm/server/BuildFarmServer.java`.  That's the initial replacement that caused the cascade of other changes.

### Followup:
In regards to what's left of `BuildFarmInstances`, it might be better renamed to `InstanceCreator` since it only has a static method to do that.  Or maybe we'll move that method out and delete it.


jacobmou (approved):
Cool! I remember Yuriy wanted to remove `BuildFarmInstances` and place it under `common` instead `server` cuz it's used in both `server` and `worker`. Looks it's not anymore!

luxe:
> Cool! I remember Yuriy wanted to remove `BuildFarmInstances`or place it under `common` instead `server` cuz it's used in both `server` and `worker`. Looks it's not anymore!

I'm also surprised its never used by `worker`.  I think its because memory/shard workers are separate executables so they just instantiate the right one in code.   However the server is the same executable so it needs to dynamically pick.

PR #944, opened 2021-11-04
Upgrade grpc java version to support latest Bazel.

comius:


deps.bzl
```diff
@@ -1,4 +1,5 @@
 """
+miu
```
luxe, line 2: typo?

deps.bzl
```diff
@@ -34,9 +35,9 @@ def archive_dependencies(third_party):
         # Needed for @grpc_java//compiler:grpc_java_plugin.
         {
             "name": "io_grpc_grpc_java",
-            "sha256": "2c6d5606abfd221e156ae9f6b52719e015751b98c642b78ab65d0accdf6e7efe",
-            "strip_prefix": "grpc-java-1.38.0",
-            "urls": ["https://github.com/grpc/grpc-java/archive/v1.38.0.zip"],
+            "sha256": "d25fdde33fe21ab659cdf5c577acc06e71b51195e0cdcf73fac188b818b00212",
+            "strip_prefix": "grpc-java-93a77a78a9a1ee889bef4c1fcfb02be343c671d1",
+            "urls": ["https://github.com/grpc/grpc-java/archive/93a77a78a9a1ee889bef4c1fcfb02be343c671d1.zip"],
```
luxe, line 40: can we move to `grpc-java-1.41.0`?  Or was there something unreleased that bazel needs?

luxe (commented):
If we can move to a versioned release of grpc, I'll be good to upgrade.  Or an explanation for the non-versioned.  thanks!

deps.bzl
comius: Yes, thanks!

comius:
Friendly ping.

luxe (approved):

PR #945, opened 2021-11-09
[logging] remove "GetActionResult 1 Requests" messages

luxe:
This log message isn't helpful.  It's always going to report "1 Requests".  
And our log viewer is showing the message 114k times.  Let's remove it and if anyone is interested in ac hits, they can check the prometheus counter `action_results` which increments at the same rate.


80degreeswest (approved):
Yep, we should be using the prom metric anyways.

PR #948, opened 2021-11-13
Increase jedis client timeout to 10 seconds

80degreeswest:
Current timeout of 2s is not sufficient enough for some calls, for example, the new get client start time one. Increasing to 10s to make sure some of the longer running calls (which are expected to be due to the number of keys they need to load) will complete in time.


80degreeswest:
I'm most likely going to end up closing this one as I found the connection timeout config in the config file, which I didn't see prior to putting up this PR.

PR #951, opened 2021-11-18
[refactor] rename variable to digestToActionCache

luxe:
This change is part of https://github.com/bazelbuild/bazel-buildfarm/pull/943, but separating it here to make smaller diffs. 

We'd like to rename this variable to `digestToActionCache` because its not the REAPI actionCache most would think of.  This is to avoid confusion.


werkt (approved):
This name seems awkward.

I would support a wholesale change for the AbstractServerInstance to retain an actionResultCache, the only possible key for which would be the ActionKey, and leave this as actionCache.

Approving since it's generally positive as is, and negates the shadowed member.

luxe:
I'll keep that in mind. ^
I was tracing the paths of using the AC which lead to the rename.

PR #952, opened 2021-11-18
[refactor] move functionality out of getActionResult

luxe:
This change is part of https://github.com/bazelbuild/bazel-buildfarm/pull/943, but separating it here to make smaller diffs. 

I'd like `ShardInstance` to implement its own version of `getActionResult`, and that can be done more cleanly if it still leverages the existing EnsureOutputs code from the abstract class.  This change splits a function into two functions.

### DIFF
--- a/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
@@ -344,6 +344,11 @@ public abstract class AbstractServerInstance implements Instance {
   public ListenableFuture<ActionResult> getActionResult(
       ActionKey actionKey, RequestMetadata requestMetadata) {
     ListenableFuture<ActionResult> result = checkNotNull(actionCache.get(actionKey));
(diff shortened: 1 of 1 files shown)


werkt:
So your intent is to call the ensure check after doing special things within the ShardInstance.

I'd prefer if we decorated the actionCache object, and left this pristine, since it represents the correct logic for an asynchronous actionCache fetch in a single point that won't require maintenance for implementation specifics.

luxe:
> So your intent is to call the ensure check after doing special things within the ShardInstance.
> 
> I'd prefer if we decorated the actionCache object, and left this pristine, since it represents the correct logic for an asynchronous actionCache fetch in a single point that won't require maintenance for implementation specifics.

We want to move the AC functionality into the backplane.  The current AC is split between the backplane and the instance hierarchy.  I see what you mean by having only 1 implementation.  Here were the issues I was hoping to address: https://github.com/bazelbuild/bazel-buildfarm/pull/943.

PR #963, opened 2021-12-07
presubmit.yml: specify a bazel version for `rpm_builds`

meteorcloudy:
The rpm build test fails in Bazel downstream because it requires a release version of Bazel.
Specifying a Bazel version number for `rpm_builds` will skip this test in Bazel downstream.

Fixes https://github.com/bazelbuild/bazel-buildfarm/issues/915

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -60,6 +60,7 @@ tasks:
     - "..."
   rpm_builds:
     name: "Rpm Builds"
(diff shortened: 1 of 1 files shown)


luxe:
thank you!
I guess our bazelisk wrapper(bazelw) is ignoring our `.bazelversion` file because of an environment var?

luxe (approved):

PR #965, opened 2021-12-13
create worker root if it doesn't already exist

luxe:
This is to simplify deployments so that the worker can start regardless of whether its root has been created yet


80degreeswest (approved):
Niiice!

PR #974, opened 2021-12-17
[deployment] [testing] fix java_image for buildfarm-shard-worker

luxe:
similar to https://github.com/bazelbuild/bazel-buildfarm/pull/968, the shard-worker container cannot be directly run from bazel.  This fixes it, so that it can be spawned that way.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -100,13 +100,17 @@ java_image(
 
 java_image(
     name = "buildfarm-shard-worker",
+    args = ["/app/build_buildfarm/examples/shard-worker.config.example"],
     base = "@ubuntu-bionic//image",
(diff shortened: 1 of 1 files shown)


werkt:
Are we losing our tiny init wrapper with this?

luxe:
> Are we losing our tiny init wrapper with this?

**tini** is a part of docker now.  We can pass `--init` to `docker run`.   

This is done either through  `./bazelw run :buildfarm-shard-worker -- --init`,
or if you `bazel build` the image first: `docker run --init <IMAGE ID>`

werkt (approved):

PR #976, opened 2021-12-22
[discussion] issue 975 example fix

luxe:


werkt:
Definitely need some discussion on this. Can you frame the situation that you discovered to determine that the entry is actually missing on the filesystem?

The longer term presentation here is that we need to a) invalidate the entry if it no longer exists, b) invalidate the downstream directories that contain it, and c) switch to a future to ensure that we can block all concurrent requests for this truth on it, assuming we move ahead with this.

luxe:
This is for the discussion on https://github.com/bazelbuild/bazel-buildfarm/issues/975.  I don't think its a good solution, but worth @shirchen trying

shirchen:
thanks! yeah, this worked and fixed #975.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -393,7 +397,7 @@ private boolean contains(
         result.mergeFrom(digest).setSizeBytes(entry.size);
       }
       onContains.accept(key);
-      return true;
+      return blobExistsOnLocalStorage(key);
```
80degreeswest, line 400: To prevent the extra overhead, can we only perform the check if ensure outputs present is set?

luxe: that's a good idea.  I'll adjust for that.

PR #983, opened 2021-12-31
[logging] remove logging noise from lognet_grpc_spring_boot_starter

luxe:
https://github.com/bazelbuild/bazel-buildfarm/issues/978


werkt (approved):
Tyvm

PR #984, opened 2022-01-02
Improve local development with Tilt

luxe:
### Problem
Deploying a complete dev environment for buildfarm requires many services to be active.  This includes server(s), worker(s), redis nodes.  Possibly a load balancer, tracing tools, buildfarm admin.  In the future, I'd like a frontend UI (backed by a database), the list goes on.  Given all these services, we need a better solution for deploying & testing locally.

### Existing strategies
Spawning individual programs with bazel is fine for getting started.  However, not everything can be tested this way.  The workers need extra programs bundled with them (the execution wrappers). There are also issues with our services being non-root.  We shouldn't rely on a dev's machine beyond simple tests.

The alternative is to spawn everything as docker containers.  There is a script in `examples/` called `bf-run` that will start and stop all our services this way.  The problem here is that the containers are already registered in dockerhub.  So it's not easy to iterate on code, and redeploy the containers.  It functions more as a pinned version of our system for someone to try out.

We could use a docker-compose file.  But I don't want to lose the ability to see all the logs (as I would with multiple terminal panes). And I don't want to sh into containers to figure out if services have crashed, etc.  Even doing this with tmux, there is no live-reload functionality based on the service that's changed.

### What I want
 - 1 command to start / stop everything
 - access to all the logs
 - reload services based on code changes

### Solution
Use Tilt with kubernetes.

Tilt explains the problems I'm alluding to here:
https://docs.tilt.dev/product_faq.html#what-kinds-of-teams-should-use-tilt

The `Tilt` file defines a complete dev environment for buildfarm.  
Run `tilt up` and see system:
Screenshot from 2022-01-01 12-11-42

The `bazel.Tiltfile` contains functionality for live-reload (discovered through bazel query).


80degreeswest (approved):
This is nice - looking forward to try it!

PR #990, opened 2022-01-08
[config] set ensure_outputs_present globally

luxe:
This was discussed for easier testing.  We also have a scons build system that was retrofitted to use REAPI and it can't handle missing blobs.  So we may also use it for that scenario.

no breaking change.  forwards compatible.

### DIFF
--- a/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
@@ -332,8 +332,15 @@ public abstract class AbstractServerInstance implements Instance {
   private static boolean shouldEnsureOutputsPresent(
       boolean ensureOutputsPresent, RequestMetadata requestMetadata) {
 
-    // We will ensure outputs are present if the system is globally configured to check.
-    // Otherwise we will determine if dynamically from optional URI parameters.
+    // The 'ensure outputs present' setting means that the AC will only return results to the client
+    // when all of the action output blobs are present in the CAS.  If any one blob is missing, the
+    // system will return a cache miss.  Although this is a more expensive check to perform, some
+    // users may want to enable this feature. It may be useful if you cannot rely on requestMetadata
+    // of incoming messages (perhaps due to a proxy). Or other build systems may not be reliable
+    // without this extra check.
+
+    // We perform the outputs present check if the system is globally configured to check for it.
+    // Otherwise the behavior is determined dynamically from optional URI parameters.
     if (ensureOutputsPresent) {
       return true;
     }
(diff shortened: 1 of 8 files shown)


src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
```diff
@@ -326,7 +329,15 @@ public DigestUtil getDigestUtil() {
             directExecutor()));
   }
 
-  private static boolean shouldEnsureOutputsPresent(RequestMetadata requestMetadata) {
+  private static boolean shouldEnsureOutputsPresent(
+      boolean ensureOutputsPresent, RequestMetadata requestMetadata) {
+
+    // We will ensure outputs are present if the system is globally configured to check.
+    // Otherwise we will determine if dynamically from optional URI parameters.
+    if (ensureOutputsPresent) {
```
80degreeswest, line 337: I don't think we should ignore the ENSURE_OUTPUTS_PRESENT but rather check for the validity of both. So if ensureOutputsPresent AND ENSURE_OUTPUTS_PRESENT exists in the url then we should perform this check. This will ensure that we can set the global config but only perform the check if the build specifically requests to use bwob.

80degreeswest (changes requested):

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
80degreeswest: Unless the intent here is to do global INSTEAD of passing ENSURE_OUTPUTS_PRESENT?

shirchen: Yeah, use case is if you can't ass in `requestMetadata` or don't want to set it for every request. I think it's also nice to save ourselves time in parsing URI.

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
```diff
@@ -326,7 +329,15 @@ public DigestUtil getDigestUtil() {
             directExecutor()));
   }
 
-  private static boolean shouldEnsureOutputsPresent(RequestMetadata requestMetadata) {
+  private static boolean shouldEnsureOutputsPresent(
```
shirchen, line 332: Can you make this a non-static method to avoid having to plumb through a boolean? Or are you doing this to make testing easier?

80degreeswest (approved):

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
luxe: improved comment

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
luxe: that would probably look cleaner.  I've been preserving `static` so that the code can be easily moved out of the class in a future refactor (and tested as you say).  Currently not unit tested.

PR #994, opened 2022-01-18
Provide real_input_directories configuration

werkt:
Migrate the hardcoded 'external' path resolution to a configurable list
of paths that will, if provided as directories, be created as regular
directories, capable of being written to, and avoiding the directory
cache.

### DIFF
--- a/examples/shard-worker.config.example
+++ b/examples/shard-worker.config.example
@@ -125,6 +125,20 @@ input_fetch_deadline: 60
 # multiple actions' executions.
 link_input_directories: true
 
+# A list of paths that will not be subject to the effects of the above
+# link_input_directories setting.
+# Relative to the input root, any directory in the input tree which matches
+# this path completely will have a real directory created for its inputs
+# for every action that executes on the worker. The example 'external' is
+# provided based on the characteristics of bazel's population of a top-level
+# external compendium which will exhibit poor directory caching performance.
+# This may also be used to provide writable directories as input roots for
+# actions which expect to be able to write to an input location and will fail
+# if they cannot.
+real_input_directories: "external"
+
+# real_input_directories: "path/to/uncached/directory"
+
 # an imposed action-key-invariant timeout used in the unspecified timeout case
 default_action_timeout: {
   seconds: 600
--- a/src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
+++ b/src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
@@ -19,6 +19,7 @@ import static build.buildfarm.common.io.Utils.readdir;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.Iterables.concat;
+import static com.google.common.collect.Iterables.filter;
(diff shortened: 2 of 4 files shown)


examples/shard-worker.config.example
```diff
@@ -124,6 +124,20 @@ input_fetch_deadline: 60
...
+# link_input_directories setting.
+# Relative to the input root, any directory in the input tree which matches
+# this path completely will have a real directory created for its inputs
+# for every action that executes on the worker. The example 'external' is
+# provided based on the characteristics of bazel's population of a top-level
+# external compendium which will exhibit poor directory caching performance.
+# This may also be used to provide writable directories as input roots for
+# actions which expect to be able to write to an input location and will fail
+# if they cannot.
+link_input_directories: "external"
```
djmarcin, line 137: Should this say "real_input_directories"?

werkt: Yes! Good catch.

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
```diff
@@ -308,14 +315,63 @@ private static void checkExecErrors(Path path, List<Throwable> errors) throws Ex
     }
   }
 
+  // binary search of our sorted list of dirs
```
luxe, line 318: we could use `Collections.binarySearch()` or something equivalent

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
```diff
@@ -308,14 +315,63 @@ private static void checkExecErrors(Path path, List<Throwable> errors) throws Ex
...
+      }
+      if (compare > 0) {
+        // search right
+        cur += len;
+      }
+    }
+    return null;
+  }
+
+  private static boolean directoryInInputs(
```
luxe, line 338: is "InInputs" a typo or is In referring to Index?

luxe (approved):
okay cool, I'll have to experiment with the setting a bit.  
Doesn't `directoryPath.split("/")` imply that it won't be top-level?  It will be any input path/that/has/external/somewhere?

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
werkt: It's just a crappy function name. 'is directory in input tree'. Changing it to treeContainsPath

werkt:
> okay cool, I'll have to experiment with the setting a bit. Doesn't `directoryPath.split("/")` imply that it won't be top-level? It will be any input path/that/has/external/somewhere?

No, the nature of the search there is such that it only matches fully formed paths: It will look for an external on the top level directory. If found, it will proceed, and if that is the end of the split list (true if just "external"), we will return true. This mimics the previous behavior that was very casually inferred by recreating the 'first incidence of a non-output external directory' behavior.

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
werkt: Yup. Working with that method is always awkward, but Lists.transform from guava might actually make it palatable (confers RandomAccess).

PR #998, opened 2022-01-19
derive execute_stage_width based on host resources

luxe:
### Summary:  
Users generally prefer their `execute_stage_width` to match the amount of CPU cores available.  When combined with `limit_execution` it properly saturates their CPU and maximizes parallelism.

Following this intuition, the diff provides 2 conveniences:  
 - Tells the user on startup if their configured stage width does not match their available CPUs.
 - Sets the stage width automatically if none is provided.
 
This helps remove any duplicate configuration that would normally occur across container deployments & worker configs.  It's also helpful for debugging a log as you know what the worker thinks its host resources are.

### Testing:  
tested this with k8s and saw the expected results.

### DIFF
--- a/src/main/java/build/buildfarm/common/config/ConfigAdjuster.java
+++ b/src/main/java/build/buildfarm/common/config/ConfigAdjuster.java
@@ -74,7 +74,7 @@ public class ConfigAdjuster {
           "grpc timeout not configured.  Setting to: " + defaultDuration.getSeconds() + "s");
     }
 
-    adjustExecuteStageWidth(builder);
+    builder.setExecuteStageWidth(adjustExecuteStageWidth(builder.getExecuteStageWidth()));
   }
 
   /**
@@ -96,7 +96,7 @@ public class ConfigAdjuster {
       builder.setCasCacheDirectory(options.casCacheDirectory);
     }
 
-    adjustExecuteStageWidth(builder);
+    builder.setExecuteStageWidth(adjustExecuteStageWidth(builder.getExecuteStageWidth()));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jacobmou:
I remember we discussed the idea a while ago, and yeah, I like it!
One thing I'm wondering is, maybe it's even better if we split the number of cores to executeStageWidth and inputeFetchStageWidth. I think that's what we have been doing in worker config.

src/main/java/build/buildfarm/common/config/ConfigAdjuster.java
```diff
@@ -136,4 +140,42 @@ public static void adjust(BuildFarmServerConfig.Builder builder, ServerOptions o
           "Bytestream timeout not configured.  Setting to: " + defaultDuration.getSeconds() + "s");
     }
   }
+
+  private static void adjustExecuteStageWidth(WorkerConfig.Builder builder) {
```
80degreeswest, line 144: Looks like there is duplication here. It may be better to have a single method that accepts the execute stage width and just set it in the calling method. Something like builder.setExecuteStageWidth(adjustExecuteStageWidth(builder.getExecuteStageWidth());

80degreeswest (commented):
I love this!

src/main/java/build/buildfarm/common/config/ConfigAdjuster.java
luxe: ty, that is way better

80degreeswest (approved):
:shipit:

jacobmou:
> > I remember we discussed the idea a while ago, and yeah, I like it! One thing I'm wondering is, maybe it's even better if we split the number of cores to executeStageWidth and inputeFetchStageWidth. I think that's what we have been doing in worker config.
> 
> yeah, and when you think about it, the worker is probably running other threads outside the fetch->execute->report pipeline. But I hope that the fetch input stage is already configured to be low and is mostly I/O bound. It does make wonder how many threads our workers actually run in parallel, and whether the whole application needs some kind of global thread pool or something to be most optimal.

Yeah, I'm also curious if we can benefit from a global thread pool. I just want to point out there are some behavior change with this PR in terms of worker execution pipe width. 

Another thing I remember is, the number of cores reported by Java `Runtime.getRuntime().availableProcessors();` will be **2X** if the cores support Hyper Threading. I remember i have tested it in some course project and noticed that if I use thread pool with the size of `Runtime.getRuntime().availableProcessors();` on machine with Hyper Thread, the performance is actually lower, which might be caused by frequent thread context switching. However, i don't know if that's associated with specific generation/architecture of CPU.

luxe:
> > > I remember we discussed the idea a while ago, and yeah, I like it! One thing I'm wondering is, maybe it's even better if we split the number of cores to executeStageWidth and inputeFetchStageWidth. I think that's what we have been doing in worker config.
> > 
> > 
> > yeah, and when you think about it, the worker is probably running other threads outside the fetch->execute->report pipeline. But I hope that the fetch input stage is already configured to be low and is mostly I/O bound. It does make wonder how many threads our workers actually run in parallel, and whether the whole application needs some kind of global thread pool or something to be most optimal.
> 
> Yeah, I'm also curious if we can benefit from a global thread pool. I just want to point out there are some behavior change with this PR in terms of worker execution pipe width.
> 
> Another thing I remember is, the number of cores reported by Java `Runtime.getRuntime().availableProcessors();` will be **2X** if the cores support Hyper Threading. I remember i have tested it in some course project and noticed that if I use thread pool with the size of `Runtime.getRuntime().availableProcessors();` on machine with Hyper Thread, the performance is actually lower, which might be caused by frequent thread context switching. However, i don't know if that's associated with specific generation/architecture of CPU.

Okay sounds like it may have issues on host machines then.  If a container approach is taken the CPU resources are usually specified as CPU units which I hope make an accurate correspondence

PR #1001, opened 2022-01-25
Compare min-mem and max-mem exec_properties to max-mem platform property

axel7083:
When defining a queue, the `max-mem` property can be used to define the maximum amount of memories any action can have per worker. If it is not defined, then we consider unlimited amount of memory for the workers and actions.

This is link to #999 .

**use case**: The workers are deployed in kubernetes and have resources limits, and having some actions which requires a lot of memories comparing to other. Instead of having for example 6 workers with 6GB, you can have 1 worker with 6 and 5 with 3, and the target in question can set the `min-mem: 5368709120` (5 * 1024 * 1024).


luxe:
ty!  did `./.bazelci/format.sh` not work for you?

src/main/java/build/buildfarm/worker/DequeueMatchEvaluator.java
```diff
@@ -161,6 +161,21 @@ private static boolean satisfiesProperty(
       return true;
     }
 
+    // validate min or max amount of memories
+    if (property.getName().equals(ExecutionProperties.MIN_MEM)
+            || property.getName().equals(ExecutionProperties.MAX_MEM)) {
+      // Considere unlimited memories if the worker have not set the MAX_CORES property.
```
luxe, line 167: consider*

luxe (approved):
great!  Just for context, I don't personally use the `DequeueMatchEvaluator` because we rely on multiple queues and allow the workers to accept everything they match with.  However, your usage in the ticket makes sense so I'm glad the mechanism has utility.

I don't think the matching logic is correct in regards to properties that are below a specified MIN_MEM, but we can add tests to verify / adjust after.  It might not be right for cores either...

I'm guessing you tested this with your own configuration & use-case.

PR #1003, opened 2022-01-27
Compare min-mem and max-mem exec_properties to max-mem platform property

luxe:
fix formatting for #1001


src/main/java/build/buildfarm/worker/DequeueMatchEvaluator.java
```diff
@@ -161,6 +161,21 @@ private static boolean satisfiesProperty(
...
 
+    // validate min or max amount of memories
+    if (property.getName().equals(ExecutionProperties.MIN_MEM)
+        || property.getName().equals(ExecutionProperties.MAX_MEM)) {
+      // Considere unlimited memories if the worker have not set the MAX_CORES property.
+      if (!workerProvisions.containsKey(ExecutionProperties.MAX_MEM)) {
+        return true;
+      }
+
+      long memoriesRequested = Long.parseLong(property.getValue());
```
werkt, line 172: Would prefer `memBytes` to all uses of `memories` here.

werkt (approved):

PR #1007, opened 2022-01-31
provide a configurable offset width when deriving execute_stage_width

luxe:
https://github.com/bazelbuild/bazel-buildfarm/issues/1006

### DIFF
--- a/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
+++ b/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
@@ -643,6 +643,11 @@ message ShardWorkerConfig {
 
   // execute width
   int32 execute_stage_width = 16;
+  
+  // An offset width used when deriving the execute stage width.
+  // Some cores of a machine are reserved for usage other than buildfarm's execution.
+  // Provide that offset here when deriving the execute stage width to avoid over saturation.
+  int32 execute_stage_width_offset = 40;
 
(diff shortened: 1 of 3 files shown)


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -1062,6 +1067,11 @@ message WorkerConfig {
 
   // execute width
   int32 execute_stage_width = 18;
+  
+  // An offset width used when deriving the execute stage width.
+  // Some cores of a machine are reserved for usage other than buildfarm's execution.
+  // Provide that offset here when deriving the execute stage width to avoid over saturation.
+  int32 execute_stage_width_offset = 31;
```
80degreeswest, line 1074: I think we should do oneof here as it doesn't make sense to use both configs at the same time. Either we explicitly specify the stage width OR we provide an offset and omit the the width.

80degreeswest (approved):

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: that sounds good.  I think we have to break backwards compat with configs because the oneof feature creates a parent type?
```
oneof execute_stage_width {
    int32 width;
    int32 offset;
} 
```

PR #1008, opened 2022-01-31
Add ability to pause worker from taking on new work.

80degreeswest:
Use case and example:

Worker is running in a cloud environment. A notification is sent to the host that the host will be shutting down. Instead of losing any work in flight, we can create an empty .paused in the CAS root, which will be read during the next worker re-registration and have worker stop taking on new work, reducing (or eliminating) the work in flight which will be lost due to a scheduled worker shutdown.

As a byproduct, this can also be used to troubleshoot any worker, by taking it out from the pool (even temporarily).

Usage - Create a new ".paused" file in the worker CAS root:

`sudo touch /var/buildfarm/worker/.paused`

If you wish to unpause the worker and have it start taking on new work, simply delete the .paused file.

A log entry will be created on a worker every 10 seconds to info rm of the worker's paused state.
A worker_paused counter metric will be emitted to allow tracking of paused workers.

### DIFF
--- a/src/main/java/build/buildfarm/worker/shard/Worker.java
+++ b/src/main/java/build/buildfarm/worker/shard/Worker.java
@@ -88,6 +88,7 @@ import io.grpc.netty.NegotiationType;
 import io.grpc.netty.NettyChannelBuilder;
 import io.grpc.services.HealthStatusManager;
 import io.prometheus.client.Counter;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -122,6 +123,8 @@ public class Worker extends LoggingMain {
           .labelNames("lifecycle")
           .help("Service health check.")
           .register();
+  private static final Counter workerPausedMetric =
+      Counter.build().name("worker_paused").help("Worker paused.").register();
 
   private static final int shutdownWaitTimeInSeconds = 10;
   private final boolean isCasShard;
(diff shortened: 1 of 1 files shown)


luxe (approved):
I might prefer that the lock go in redis instead, but that would require additional plumbing and I assume whatever is creating the `.paused` file for you, is less capable of doing that.  So this seems best.

Having the file would be helpful when sshing around (as you say), but perhaps harder if wanting to query all machines that are paused, etc.  

Anyhow, looks like progress for 'graceful shutdown', so approving. 

Maybe you will combine these two state variables going forward:
`inGracefulShutdown` and `isWorkerPausedFromNewWork`.  
Seems like `isWorkerPausedFromNewWork` would imply `inGracefulShutdown`??

80degreeswest:
> I might prefer that the lock go in redis instead, but that would require additional plumbing and I assume whatever is creating the `.paused` file for you, is less capable of doing that. So this seems best.
> 
> Having the file would be helpful when sshing around (as you say), but perhaps harder if wanting to query all machines that are paused, etc.
> 
> Anyhow, looks like progress for 'graceful shutdown', so approving.
> 
> Maybe you will combine these two state variables going forward: `inGracefulShutdown` and `isWorkerPausedFromNewWork`. Seems like `isWorkerPausedFromNewWork` would imply `inGracefulShutdown`??

Yeah I actually gave it a very hard thought on doing the lock in Redis (similar how bfadmin used to handle this) but in the end the primary use case I see for this is some external trigger deciding the worker needs to shut down (for example aws metric for which I listen to on the ec2 instance). I did add the prometheus metric to track hosts which are paused. Eventually if noone is using the old implementation of graceful shutdown we can deprecate it but I didn't want to do it just yet as it does it via a different route (using instance termination protection) and is very aws specific. This implementation is generic.

PR #1010, opened 2022-02-01
[feature] capture GRPC header data

luxe:
bazel supports the following remote header flags:  

`--remote_header=<a 'name=value' assignment>`
`--remote_exec_header=<a 'name=value' assignment>`
`--remote_downloader_header=<a 'name=value' assignment>`
`--remote_cache_header=<a 'name=value' assignment>`

We adjust our grpc intercept to extract these headers and make them available for use.  We use any headers that are given in place of platform properties.

This is also relevant to a public comment:  

Anders Brujordet:
> I did a build bazel build //... and then a bazel test //.... My tests are configured with --nosandbox_default_allow_network and --remote_default_exec_properties=linux-sandbox=True --remote_default_exec_properties=block-network=True to block network access during tests both locally and when executed remotely on our BuildFarm. Since exec_properties are part of the action digest. To prevent rebuilds I would  need to set --remote_default_exec_properties= on build instead of only test. But I only want to block network access on tests. How could I solve this?
> Some RBE implementations (e.g. BuildBuddy) allow setting the platform properties via --remote_header, to prevent the action cache issue. Is there something similar for BuildFarm?


src/main/java/build/buildfarm/common/grpc/TracingMetadataUtils.java
```diff
@@ -71,8 +80,25 @@ public <ReqT, RespT> Listener<ReqT> interceptCall(
...
       }
+
+      capturedHeaders = extractHeaders(headers);
+
       Context ctx = Context.current().withValue(CONTEXT_KEY, meta);
       return Contexts.interceptCall(ctx, call, headers, next);
     }
+
+    private static ArrayList<Property> extractHeaders(Metadata headers) {
+      ArrayList<Property> capturedHeaders = new ArrayList<>();
```
werkt, line 91: Prefer ImmutableList for this construction, with an Iterable return.

src/main/java/build/buildfarm/common/grpc/TracingMetadataUtils.java
```diff
@@ -35,6 +37,8 @@ private TracingMetadataUtils() {}
   private static final Context.Key<RequestMetadata> CONTEXT_KEY =
       Context.key("remote-grpc-metadata");
 
+  private static ArrayList<Property> capturedHeaders = new ArrayList<>();
```
werkt, line 40: Specify variable type as bare semantics required - Iterable likely.

Also, what on earth? Why is this static? You will completely destroy any concurrent use, and never be guaranteed that 'headersFromCurrentContext' is valid with this workflow.

werkt (changes requested):
I see this having some major trouble, please reevaluate...

src/main/java/build/buildfarm/common/grpc/TracingMetadataUtils.java
```diff
@@ -45,9 +49,14 @@ public static RequestMetadata fromCurrentContext() {
     if (metadata == null) {
       metadata = RequestMetadata.getDefaultInstance();
     }
+
```
werkt, line 52: nit: whitespace

src/main/java/build/buildfarm/common/grpc/TracingMetadataUtils.java
```diff
@@ -35,6 +37,8 @@ private TracingMetadataUtils() {}
   private static final Context.Key<RequestMetadata> CONTEXT_KEY =
       Context.key("remote-grpc-metadata");
 
+  private static ImmutableList<Property> capturedHeaders = ImmutableList.of();
+
```
werkt, line 41: This should not be static to TracingMetadataUtils - it does not follow by guarantee the current context, nor is it threadsafe or aware.

src/main/java/build/buildfarm/common/grpc/TracingMetadataUtils.java
```diff
@@ -45,9 +49,14 @@ public static RequestMetadata fromCurrentContext() {
...
       metadata = RequestMetadata.getDefaultInstance();
     }
+
     return metadata;
   }
 
+  public static Iterable<Property> headersFromCurrentContext() {
+    return capturedHeaders;
+  }
+
```
werkt, line 59: You must change this to perform a headers extraction from the current context - you can relatively easily accomplish this by setting a separate key on the context with an additional `withValue` for a separate key.

werkt (changes requested):
This uses a static that does not follow a 'current' context scheme, it must be changed to be consistent and thread-friendly.

PR #1013, opened 2022-02-04
[startup] warn about missing execution wrappers

luxe:
Improper dockerizing of a worker image will lead to missing dependencies needed during action execution.  For example, `cgexec` is needed to limit execution- but did you install it?  To combat these mistakes, we correlate all features with execution wrappers and warn the user if they are missing anything on startup.  

For example,  if I run the worker directly from bazel, I may not have the things I need. 
```
./bazelw run --jvmopt="-Djava.util.logging.config.file=/home/luxe/Desktop/repos/bazel-buildfarm/src/main/java/build/buildfarm/logging.properties" //src/main/java/build/buildfarm:buildfarm-shard-worker $PWD/examples/shard-worker.config.example
```

And I'll see the following warnings:  
```
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/skip_sleep is missing and therefore the following features will not be available: skip-sleep, time-shift  
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/skip_sleep_preload.so is missing and therefore the following features will not be available: skip-sleep, time-shift  
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/delay.sh is missing and therefore the following features will not be available: skip-sleep, time-shift  
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/linux-sandbox is missing and therefore the following features will not be available: linux-sandbox, block-network, tmpfs  
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/as-nobody is missing and therefore the following features will not be available: as-nobody  
[WARNING] build.buildfarm.common.config.ConfigAdjuster lambda$checkExecutionWrapperAvailability$0 - the execution wrapper /app/build_buildfarm/process-wrapper is missing and therefore the following features will not be available: process-wrapper
```

If you run a properly built container (like the one on the root of the repo) you will not see these warnings.  

Actually you'll see 1 warning... the root image definition is missing `cgexec`.  But now I know, and am fixing it https://github.com/bazelbuild/bazel-buildfarm/pull/1024.


werkt (approved):
Seems alright. Wish we had a non-.sh version of the delay script.

PR #1015, opened 2022-02-04
Add workaround for SIGBUS crash

philwo:
I noticed that your tests are flakily crashing with a SIGBUS error.

Adding --sandbox_tmpfs_path=/tmp will prevent this error. Bazel CI usually injects it, but as you're using calling Bazel manually via shell scripts as part of your tests, this mechanism doesn't work, but we can just add it to the project-level bazelrc file.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,6 @@
+common --enable_platform_specific_config
+
 build:fuse --define=fuse=true
+
+# Workaround for https://github.com/bazelbuild/bazel/issues/3236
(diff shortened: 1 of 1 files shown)


philwo:
Example for a SIGBUS crash that I mentioned: https://buildkite.com/bazel/bazel-bazel/builds/18580#8fcfbcd2-bef4-4094-bb52-831df15b044f

luxe (approved):

PR #1017, opened 2022-02-04
[metrics] Allow java-grpc-prometheus metrics (needed for SLA enforcement)

luxe:
In order to enforce SLAs, there is a requirement to provide latency metrics for all endpoints of our service.  This could be done at the proxy level, but server intercepts are equally useful.  They make it easy to track individual server performance.  It also let's us observe latency differences between `client<->proxy` vs `proxy<->server`- or what we consider, frontend/backend performance.

I realize some of these metrics overlap with the individual counters and timers put in the grpc implementations.  However, this is a more consistent way for us to track everything.  Additionally, the registration of these metrics are completely optional, so nobody needs to use this if they don't want to.

This was tested locally in Tilt.  I'll find out in future deployments how much load it causes our agents.  Some example metrics:  
```
curl -q localhost:9090 2>&1 | grep ActionCache
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.001",} 11.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.005",} 17.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.01",} 18.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.05",} 18.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.075",} 18.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.1",} 22.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.25",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="0.5",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="1.0",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="2.0",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="5.0",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="10.0",} 27.0
grpc_server_handled_latency_seconds_bucket{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",le="+Inf",} 27.0
grpc_server_handled_latency_seconds_count{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",} 27.0
grpc_server_handled_latency_seconds_sum{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",} 0.9319999999999999
grpc_server_handled_total{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",code="NOT_FOUND",grpc_code="NOT_FOUND",} 18.0
grpc_server_handled_total{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",code="OK",grpc_code="OK",} 9.0
grpc_server_started_total{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",} 27.0
grpc_server_handled_created{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",code="NOT_FOUND",grpc_code="NOT_FOUND",} 1.644006403761E9
grpc_server_handled_created{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",code="OK",grpc_code="OK",} 1.644006401604E9
grpc_server_handled_latency_seconds_created{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",} 1.644006401604E9
grpc_server_started_created{grpc_type="UNARY",grpc_service="build.bazel.remote.execution.v2.ActionCache",grpc_method="GetActionResult",} 1.644006401503E9
```

**PLEASE NOTE:**
- [x] This is forwards compatible with config files.
- [x] It's entirely opt-in.  Landing this will have no effect on existing deployments.
- [x] Prometheus metrics are not even registered in default config.


jlisee:
An aside:

Really great to see these kind of automatic system metrics like this.  @luxe do you have any distributed tracing setup in any of your lab or production environments?  At Aurora we use Honeycomb and are starting to use  [OpenTelemetry](https://opentelemetry.io/) for similar kind of work.  There is [a package](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-grpc-1.6) ([source](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6)) for the same automatic gRPC instrumentation.

The use case I am thinking about is being able to see problem or failed traces and all the others bits of the infrastructure they interacted with and how long each part took.  On a large cluster we would probably need [Refinery](https://github.com/honeycombio/refinery/) (sampling proxy) and maybe even client side sample too.  For example you could save all traces that involved errors or took a certain amount of time, than aggressively sample the rest.  Then in an instant if you had or slow build you could look for all traces from that build and dig into particularly bad operations and see what is up.

luxe:
> An aside:
> 
> Really great to see these kind of automatic system metrics like this. @luxe do you have any distributed tracing setup in any of your lab or production environments? At Aurora we use [Honeycomb](https://honeycomb.io/) and are starting to use [OpenTelemetry](https://opentelemetry.io/) for similar kind of work. There is [a package](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-grpc-1.6) ([source](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/grpc-1.6)) for the same automatic gRPC instrumentation as you have here for Java.
> 
> The use case I am thinking about is being able to see problem or failed traces and all the others bits of the infrastructure they interacted with and how long each part took. On a large cluster we would probably need [Refinery](https://github.com/honeycombio/refinery/) (sampling proxy) and maybe even client side sampling as well. For example you could save all traces that involved errors or took above a certain amount of time, than aggressively sample the rest. Then in an instant if you had or slow build you could look for all traces from that build and dig into particularly bad operations and see what is up.
> 
> They complement this data - the metrics tells you "oh no this end point is pretty slow and/or has lots of errors". Then the tracing would let you look at example slow or broken queries to see what is going on and why. If you add in the right tracing you can also monitoring requests to services like redis or AWS to see if they are causing issues.

Thank you for posting this!  The integration looks somewhat easy considering we don't have to modify code. I don't have a distributed tracing UI such as honeycomb, but I've used honeycomb in the past, and know what you're referring to.  Perhaps the best path forward would be to attach opentelemetry and see what that extracted data looks like.  I also wonder if there is some free visualizer out there similar to honeycomb that I could use during local development.

jlisee:
> Thank you for posting this! The integration looks somewhat easy considering we don't have to modify code. I don't have a distributed tracing UI such as honeycomb, but I've used honeycomb in the past, and know what you're referring to. Perhaps the best path forward would be to attach opentelemetry and see what that extracted data looks like. I also wonder if there is some free visualizer out there similar to honeycomb that I could use during local development.

You should be able to get a free local only setup by using Jeager as your backend and UI and the OpenTelemetry collector to collect from Buildfarm and export to Jeager in Zipkin format.

You would essentially do the following:
 - Start up Jeager (open source distributed tracing UI and storage) locally with docker ([docs here](https://www.jaegertracing.io/docs/1.31/getting-started/))
 - Start the [OpenTelemetry collector](https://github.com/open-telemetry/opentelemetry-collector) server configured to convert receive `otlp` and export in `zipkin` format to Jeager  ([configuration docs](https://opentelemetry.io/docs/collector/configuration/))
 - Start Buildfarm configured to send OpenTelemetry in OLTP format to the local collector
 - View traces in the local Jeager UI

80degreeswest:
One of the libraries that's been on my radar to use is Micrometer (https://github.com/micrometer-metrics/micrometer). It seems to be very flexible and supports a ton of different metrics out of the box.

80degreeswest:
Also, Honeycomb has a free tier plan if you wanted to use it.

luxe:
Okay, thank you for the suggestions.  [Here](https://github.com/bazelbuild/bazel-buildfarm/pull/1032) is my first pass at it.  There is a lot of potential around these ideas, and it's likely to be the future way we do things.  I'd still like to land this PR due to its simplicity around how we currently deploy.  Going forward, I will look into [opentelemetry-grpc-1.6](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-grpc-1.6) as a replacement.

examples/shard-server.config.example
```diff
@@ -265,4 +293,4 @@ cas_write_timeout: {
 bytestream_timeout: {
   seconds: 3600
   nanos: 0
-}
+}
```
werkt, line 296: Don't kill my newlines, bro.

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -107,40 +114,67 @@ public BuildFarmServer(
...
+            config.getExecuteKeepaliveAfterSeconds(),
+            TimeUnit.SECONDS,
+            keepaliveScheduler,
+            getMetricsPublisher(config.getMetricsConfig())));
+    serverBuilder.addService(new OperationQueueService(instance));
+    serverBuilder.addService(new OperationsService(instance));
+    serverBuilder.addService(new AdminService(config.getAdminConfig(), instance));
+    serverBuilder.addService(new FetchService(instance));
+    serverBuilder.addService(ProtoReflectionService.newInstance());
+    serverBuilder.addService(new PublishBuildEventService(config.getBuildEventConfig()));
```
werkt, line 142: Not a fan of this style change. It hides the actual modification here.

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -107,40 +114,67 @@ public BuildFarmServer(
...
+      if (config.getGrpcMetrics().getProvideLatencyHistograms()) {
+        grpcConfig = grpcConfig.allMetrics();
+      }
+
+      // Enable latency buckets with status codes.
+      if (config.getGrpcMetrics().getTagLatenciesByCode()) {
+        // TODO(luxe):
+        // The feature is not yet included in latest release.
+        // Once the feature is available, we can enable this code:
+        // grpcConfig = grpcConfig.withCodeLabelInLatencyHistogram();
```
werkt, line 170: Safe to leave this in if the default is disabled and this feature is also disabled. I have an aversion to commented-out code.

werkt (changes requested):
Some nits

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -90,10 +92,15 @@ public BuildFarmServer(
...
-
     healthStatusManager = new HealthStatusManager();
+    instance = BuildFarmInstances.createInstance(session, config.getInstance(), this::stop);
+    server = createServer(serverBuilder, instance, config);
+    logger.log(Level.INFO, String.format("%s initialized", session));
+  }
 
+  private Server createServer(
+      ServerBuilder<?> serverBuilder, Instance instance, BuildFarmServerConfig config)
+      throws InterruptedException, ConfigurationException {
```
shirchen, line 103: i would do these (and other refactoring below) in a separate diff.

src/main/java/build/buildfarm/server/BuildFarmServer.java
shirchen: +1, you can refactor later or stack diffs.

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -107,40 +114,67 @@ public BuildFarmServer(
...
+  }
+
+  private static void handleGrpcMetricIntercepts(
+      ServerBuilder<?> serverBuilder, BuildFarmServerConfig config) {
+    // Decide how to capture GRPC Prometheus metrics.
+    // By default, we don't capture any.
+    if (config.getGrpcMetrics().getEnabled()) {
+      // Assume cheap metrics.
+      // Cheap metrics include send/receive totals tagged with return codes.  No latencies.
+      Configuration grpcConfig = Configuration.cheapMetricsOnly();
```
shirchen, line 158: nit: "cheap" should probably be replaced by something like 'core'. otherwise, you may want to flip the config, and add an option like `include_latencies`. If we add more things, it'll be a clear pattern on how to add more flags in the future.

src/main/java/build/buildfarm/server/BuildFarmServer.java
shirchen: +1. Just put this in a separate branch and reuse it later.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -209,6 +209,28 @@ message PollOperationRequest {
   build.bazel.remote.execution.v2.ExecutionStage.Value stage = 3;
 }
 
+// Publish GRPC metrics to promethesus for each server endpoint. This is used
+// for SLAs at API boundaries. You may want something similar at the proxy level
+// to see frontend/backend performance.
+message GrpcPrometheusMetrics {
```
shirchen, line 215: Does buildfarm server serve anything other than GRPC? I think there is a proxy, but is prefixing with GRPC a bit redundant if its the only protocol buidlfarm serves.

src/main/java/build/buildfarm/server/BuildFarmServer.java
shirchen: Also, I dont see a cheap_metrics_only option. Looks like this function is missing unit tests.

shirchen (commented):
It would be great to write some tests for this and I think we can improve on naming, and reconfigure to have an "include_latencies" option.

examples/shard-server.config.example
luxe: fixed

src/main/java/build/buildfarm/server/BuildFarmServer.java
luxe: reverted.  The diff is a little clearer now

src/main/java/build/buildfarm/server/BuildFarmServer.java
luxe: removed

src/main/java/build/buildfarm/server/BuildFarmServer.java
luxe: reverted createServer function.  A little more clear now.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: There is a similar grpc server built in `buildfarm/proxy/http/HttpProxy.java`.  It could get the same configuration in the future if there is a need.  I haven't used that proxy before.  Our load balancers (haproxy) forward all the traffic. Yes, all the Prometheus metrics are prefixed with `grpc_*`.  That's from the intercept library deciding to do that.

src/main/java/build/buildfarm/server/BuildFarmServer.java
luxe: its a good point.  the "cheap" terminology came from the library itself.  
For our purposes, we could say that our core metrics are what this library is considering as default (which is what you get through the cheapMetricsOnly() call).  

If you enable the metrics with `grpc_metrics.enabled: true`, you'll get those core metrics.  
If you want more, you will further enable the hisograms with `grpc_metrics.provide_latency_histograms: true`.  

Trying to match how their [library config](https://github.com/grpc-ecosystem/java-grpc-prometheus/blob/8b4d92e76f0f59bf4393b3f8d6cdfdd1386154e3/src/main/java/me/dinowernli/grpc/prometheus/Configuration.java#L25) code works:  

```
  /** Returns a {@link Configuration} for recording all cheap metrics about the rpcs. */
  public static Configuration cheapMetricsOnly() {
    return new Configuration(
        false /* isIncludeLatencyHistograms */,
        CollectorRegistry.defaultRegistry,
        DEFAULT_LATENCY_BUCKETS,
        new ArrayList<>(),
        false /* isAddCodeLabelToHistograms */);
  }

  /**
   * Returns a {@link Configuration} for recording all metrics about the rpcs. This includes metrics
   * which might produce a lot of data, such as latency histograms.
   */
  public static Configuration allMetrics() {
    return new Configuration(
        true /* isIncludeLatencyHistograms */,
        CollectorRegistry.defaultRegistry,
        DEFAULT_LATENCY_BUCKETS,
        new ArrayList<>(),
        false);
  }
```

You can see their idea of "cheap" metrics... or core metrics, is having the other metric options turned off.

luxe: changing our use of "cheap" to say "core".

shirchen: Oh, my bad, I should have followed the object type to the top. Yeah, I see that cheap is something we may be stuck with, though I don't think this is a descriptive term.  

Also, how about that unit test for this static function? ;)

examples/shard-server.config.example
```diff
@@ -255,6 +255,34 @@ ac_policy: READ_AND_WRITE
...
+  # with return codes.  No latencies.
+  provide_latency_histograms: False
+  
+  # A status code label will be added to latency histogram. If latency
+  # histogram itself is disabled, this takes no effect. Warning: this will
+  # increase the number of histograms by a factor of actually happened codes
+  # (up to io.grpc.Status.Code values count), which could lead to additional
+  # local memory usage and load on prometheus (storage and memory usage,
+  # query-time complexity)
+  tag_latencies_by_code: False
```
shirchen, line 283: You can remove this, now that you put the commented out code into a branch to work on this later.

shirchen (approved):
Did you want to add some new metrics to the documentation as well? https://bazelbuild.github.io/bazel-buildfarm/docs/metrics/metrics/

src/main/java/build/buildfarm/server/BuildFarmServer.java
luxe: added unit tests

examples/shard-server.config.example
luxe: removed

luxe:
> Did you want to add some new metrics to the documentation as well? https://bazelbuild.github.io/bazel-buildfarm/docs/metrics/metrics/

added docs

luxe:
bump

luxe:
bump

80degreeswest (approved):

luxe:
@werkt please re-review

werkt (approved):
LGTM

PR #1018, opened 2022-02-06
[security] add an access control policy for the ActionCache

luxe:
### problem:
A client with write access to the action cache can write anything, including malicious code and binaries, which can then be returned to other users on cache lookups.  To prevent this security vulnerability we provide a control policy when configuring the server.  

### implementation:
The buildfarm server can be set to read-only for the action cache.   When clients attempt to populate the action cache they will be given a PERMISSION_DENIED response.  In the future, we may want this enforced through a security token (so that different clients can have different levels of permission).  Deploying a server with the endpoint disabled is a stronger guarantee that the action cache will never be written to by clients.

### related:
related issues https://github.com/bazelbuild/bazel-buildfarm/issues/386, https://github.com/bazelbuild/bazel-buildfarm/issues/502, https://github.com/bazelbuild/bazel-buildfarm/issues/391

Kind of similar to https://github.com/bazelbuild/bazel-buildfarm/issues/216.  However, we are providing a global config to disable the endpoint.  The other issue seems to imply an analysis of incoming requests.

### configuration:
 - [x] Configuration is forwards compatible.
 - [x] This has no behavior changes for existing deployments


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -211,6 +211,8 @@ message PollOperationRequest {
 
 message BuildFarmServerConfig {
   InstanceConfig instance = 1;
+  
+  ActionCacheAccessPolicy ac_policy = 12;
```
werkt, line 215: Is there a more standard (likely not protobuf, but maybe googleapis) type we could use for this?

src/main/java/build/buildfarm/server/ActionCacheService.java
```diff
@@ -100,6 +103,16 @@ public void onFailure(Throwable t) {
   @Override
   public void updateActionResult(
       UpdateActionResultRequest request, StreamObserver<ActionResult> responseObserver) {
+    // A user with write access to the cache can write anything, including malicious code and
+    // binaries, which can then be returned to other users on cache lookups.  This is a security
+    // concern.  To counteract this, we allow enforcing a policy where clients cannot upload to the
+    // action cache.  In this paradigm, it is only the remote execution engine itself that populates
+    // the action cache.
+    if (policy.equals(ActionCacheAccessPolicy.READ_ONLY)) {
```
werkt, line 111: Set a boolean in the service instance that enables writing based on this policy, rather than the equals check each time.

if (!writable) {
  <perm denied>
  return;
 }

werkt (approved):
Glad to see this finally get some love. I think the sec teams would prefer a unit test for the service though.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: hmm,  [visibility](https://github.com/googleapis/googleapis/blob/master/google/api/visibility.proto#L89) is close.  We could consider the service method having a visibility restriction (i.e. not client facing).  
This [read-only / read-write mode](https://github.com/googleapis/googleapis/blob/master/google/datastore/v1/datastore.proto#L403) may be closer.  If we say the AC is a "datastore" and clients make "transactions" to it..

src/main/java/build/buildfarm/server/ActionCacheService.java
luxe: fixed

luxe:
> Glad to see this finally get some love. I think the sec teams would prefer a unit test for the service though.

fixed

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
luxe: landing for now, but I like the idea of adopting something from googleapi to represent our storages

PR #1020, opened 2022-02-09
[executor] remove execLock

luxe:
Why is there a synchronization block on starting a process?  Can we remove it?


werkt:
This is why:

https://github.com/bazelbuild/bazel/commit/ba76255ea973d22c3cf5456eda93d301ed4b4626

luxe:
ty!  I will document that in follow-up diff

PR #1028, opened 2022-02-15
[backplane] use redisHashMap for workers

luxe:
We continue simplifying the `RedisShardBackplane` by encapsulating redis logic into generic container interfaces.  This will make it easier to configure different backplanes in the future.  It also moves us closer to combining our memory/shard implementations through container configuration.


krisstakos:
Looks good to me!

PR #1031, opened 2022-02-18
[bugfix] fix worker shutdown on storage-only configuration (#1030)

luxe:
### Problem:
When using storage-only configuration:  
```
capabilities {
  cas: true
  execution: false
}
```
Buildfarm avoids the execution pipeline and instead constructs a pipeline with a single stage of `PipelineStage.NullStage`.  This is a foolish way to construct a pipeline because it causes the pipeline to finish immediately.  A finished pipeline then causes the worker to prematurely shutdown.

### Solution:
To resolve this, we create a standalone "CasReplication" stage to prevent worker shutdown.  Now users can continue to use workers for storage-only while avoiding the construction of the execution pipeline.

We also log the case in which workers shutdown because of a completed pipeline.

### Testing:
Not that happy about the test.  Any better ideas?

### Related:
Tracking issue for storage workers: https://github.com/bazelbuild/bazel-buildfarm/issues/857
The bug report:  https://github.com/bazelbuild/bazel-buildfarm/issues/1030


vtomilov:
Hi! I'm facing the same issue. As a temporary workaround I've enabled execution capability on cache-only nodes with no `provisioned_queues` so the node won't actually do any tasks, but I'm not sure if there are any pitfalls in such approach.
I see it's still waiting for the review, but it seems like a major issue to me, don't you agree?

src/main/java/build/buildfarm/worker/CasReplicationStage.java
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.worker;
+
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.logging.Logger;
+
+public class CasReplicationStage extends PipelineStage {
```
werkt, line 21: Not a fan of this name, it does nothing with CasReplication or anything at all except keeping the pipeline alive.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -515,13 +514,17 @@ private Server createServer(
           new WorkerProfileService(
               storage, inputFetchStage, executeActionStage, context, completeStage, backplane));
     } else {
-      PipelineStage casReplicationStage = new CasReplicationStage();
-      pipeline.add(casReplicationStage, 1);
+      createStorageOnlyPipeline(pipeline);
```
werkt, line 517: Realistically it just doesn't need to have a pipeline running at all, and I'd prefer to have a dead pipeline (null or otherwise meaningless) that is interpreted by blockUntilShutdown as something that only hangs off of the server. That will obviate much of the need for the extra classes here and handling. Further reorganization should make the worker not have any knowledge of either server or pipeline, and just hang off of a critical service presentation layer to determine if it should shut down.

werkt (changes requested):

src/main/java/build/buildfarm/worker/CasReplicationStage.java
luxe: removed

src/main/java/build/buildfarm/worker/shard/Worker.java
luxe: for now, I removed the unnecessary pipeline stage.  We detect there are no pipeline stages and prevent shutdown.

werkt (approved):

PR #1032, opened 2022-02-19
[tracing] integrate distributed tracing via OpenTelementry & Jaeger

luxe:
### Feature
We enable **distributed tracing** for our k8s cluster by instrumenting java and including new pods for [open telementry](https://opentelemetry.io/) & [jaeger](https://www.jaegertracing.io/).  The buildfarm JARs get instrumented automatically via [opentelemetry-java-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation).  This causes the JVM agent to send tracing information to an **opentelementry collector pod**.  The collector pod then exports it to **jaeger**.

### Setup
[Tilt](https://tilt.dev/) is used as a development environment for testing our k8s cluster.  
Here is a grouping of buildfarm pods along with a grouping of tracing-related pods:  

Screenshot from 2022-02-18 22-19-37

### Results
Out of the box we can see that jaeger has received information for **http** and **jedis**.

Screenshot from 2022-02-18 22-48-13

Overall, the information is sparse.  On a positive note, its correlated traces across server and worker boundaries.  We'll see what 3rd party extensions are available and how much we should add our own instrumentation to improve it further.  I was hoping to get something similar to datadog's java callstack profile as part of the trace.  This is either a feature I didn't see, or a fundamental limitation to what an agent can instrument automatically.


BUILD
```diff
@@ -30,6 +45,13 @@ java_library(
     ],
 )
 
+java_library(
+    name = "telemetry_tools",
```
werkt, line 49: somewhat unrelated - we should pick a convention between hyphens and underscores for target names. Below you have opentelemetry-javaagent for comparison

Tiltfile
```diff
@@ -41,7 +41,7 @@ custom_build(
   deps=bazel_sourcefile_deps('//:buildfarm-server.tar')
 )
 
-local_resource("unit tests",'./bazelw test --javabase=@bazel_tools//tools/jdk:remote_jdk11 //src/test/java...')
+local_resource("unit tests",'./bazelw test --javabase=@bazel_tools//tools/jdk:remote_jdk11 src/test/java/...')
```
werkt, line 44: Why remove the `//`?

werkt (approved):
Some nits

BUILD
luxe: agree

Tiltfile
luxe: no good reason. added it back.

PR #1037, opened 2022-02-26
[compression] rewrite ByteStream URI parser for compression support

luxe:
## Background Information:
Interacting with the **CAS** can be done through google's [ByteStream API](https://github.com/googleapis/googleapis/blob/master/google/bytestream/bytestream.proto).  The ByteStream API contains `ReadRequest` and `WriteRequest` types which carry an attribute known as **resource name**.  The resource name is a **URI** that must be parsed in order to derive specific information defined by REAPI ([see here](https://github.com/bazelbuild/remote-apis/blob/04784f4a830cc0df1f419a492cde9fc323f728db/build/bazel/remote/execution/v2/remote_execution.proto#L206)).

In order to properly support compression we need to update the parser to understand `compressed-blobs` and `compressor` path segments.  This is what those URI formats look like:

**Uploading Blobs:**
{instance_name}/uploads/{uuid}/blobs/{hash}/{size}{/optional_metadata}
{instance_name}/uploads/{uuid}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}{/optional_metadata}

**Downloading Blobs:**
{instance_name}/blobs/{hash}/{size}
{instance_name}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}

## Compliance Issues
The existing parser is not compliant for the following reasons:
 - Does not account for **compression**
 - Does not account for multi-segment **instance names**
 - Does not account for multi-segment **resource metadata**
 
 ## Changes
 - Create a standalone URI parser that is fully compliant
 - Integrate it back  to where it is currently being used
 - Resolve the [FIXME](https://github.com/bazelbuild/bazel-buildfarm/blob/f84ecd95b2ad451c761ddcaa20b770220a37ca7d/src/main/java/build/buildfarm/common/UrlPath.java#L22) about using proto sum type
  - Resolve the [FIXME](https://github.com/bazelbuild/bazel-buildfarm/blob/f84ecd95b2ad451c761ddcaa20b770220a37ca7d/src/main/java/build/buildfarm/common/UrlPath.java#L149) about adding compliance
   - Improve test coverage

Tracking issue: https://github.com/bazelbuild/bazel-buildfarm/issues/956
Additional reference: https://github.com/bazelbuild/bazel-buildfarm/compare/main...shirchen:shirchen/compression?expand=1

 ## Testing
 New module has 100% test coverage.  Integrated back to caller such that the integration tests will also be using this.


shirchen:
this lgtm to me. thanks for taking the branch and making it production ready!

src/main/java/build/buildfarm/common/resources/BUILD
```diff
@@ -0,0 +1,54 @@
...
+
+java_proto_library(
+    name = "resource_java_proto",
+    deps = [
+        ":resource_proto",
+    ],
+)
+
+java_library(
+    name = "ResourceParser",
```
werkt, line 27: Rename this to `resources` - we have no capitalized names throughout, and our convention is to leverage bazel's basename target policy as much as possible.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -0,0 +1,201 @@
...
+    // from.
+    String[] segments = tokenize(resourceName);
+    MutableInt index = startIndex(segments, Resource.TypeCase.UPLOAD_BLOB_REQUEST);
+
+    // Extract each of the following segments:
+    // `{instance_name}/uploads/{uuid}/blobs/{hash}/{size}{/optional_metadata}` or
+    // `{instance_name}/uploads/{uuid}/compressed-blobs/{compressor}/{uncompressed_hash}/{uncompressed_size}{/optional_metadata}`
+    UploadBlobRequest.Builder builder = UploadBlobRequest.newBuilder();
+    builder.setInstanceName(
+        String.join("/", Arrays.copyOfRange(segments, 0, index.getAndIncrement())));
```
werkt, line 85: please add an appropriate named `static final` separator constant and a `resourcePath` method that does the join over components - fewer magic numbers, fewer convoluted `String.join`s

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -0,0 +1,201 @@
...
+   * @return A lookup table for keyword to resource type.
+   * @note Suggested return identifier: keywordMap.
+   */
+  private static HashMap<String, Resource.TypeCase> keywordResourceMap() {
+    HashMap<String, Resource.TypeCase> keywords = new HashMap<>();
+    keywords.put("blobs", Resource.TypeCase.DOWNLOAD_BLOB_REQUEST);
+    keywords.put("compressed-blobs", Resource.TypeCase.DOWNLOAD_BLOB_REQUEST);
+    keywords.put("uploads", Resource.TypeCase.UPLOAD_BLOB_REQUEST);
+    keywords.put("operations", Resource.TypeCase.STREAM_OPERATION_REQUEST);
+    return keywords;
```
werkt, line 126: Recommend `ImmutableMap.of` here, rather than the hash map construction, and make this into a publicly accessible static property (ImmutableMap will guarantee attr_reader) that exists as a single instance.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * @brief Tokenize the resource name URI.
+   * @details A simple split should be fine.
+   * @param resourceName The resource name URI defined by REAPI.
+   * @return A tokenized resource name.
+   * @note Suggested return identifier: uRITokens.
+   */
+  private static String[] tokenize(String resourceName) {
+    return resourceName.split("/");
+  }
```
werkt, line 137: good place to put the neighbor `resourcePath(Iterable)` method

werkt (changes requested):
Some nits, good cleanup overall.

src/main/java/build/buildfarm/common/resources/BUILD
luxe: fixed

src/main/java/build/buildfarm/common/resources/ResourceParser.java
luxe: added

src/main/java/build/buildfarm/common/resources/ResourceParser.java
luxe: fixed

src/main/java/build/buildfarm/common/resources/ResourceParser.java
luxe: I'm thinking what the implementation would be.  I wasn't able to parse the resource path sequentially.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -0,0 +1,232 @@
...
+ *     utilities are for parsing "resource name" and extracting the information into easy to work
+ *     with types.
+ */
+public class ResourceParser {
+  /**
+   * @field RESOURCE_SEPERATOR
+   * @brief Path seperator to use when parsing.
+   * @details Part of REAPI's format of resource name.
+   */
+  private static final String RESOURCE_SEPERATOR = "/";
```
werkt, line 44: sp: Separator

src/main/java/build/buildfarm/common/resources/ResourceParser.java
werkt: Just the ever present join analogue to `tokenize`

```
private static String resourcePath(Iterable components) {
  return RESOURCE_SEPARATOR.join(components);
}
```

luxe: Replaced array argument with `Iterable<String>`.  Removed original implementation since it wouldn't get test coverage.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
luxe: fixed

luxe:
bump

werkt (approved):

PR #1040, opened 2022-03-13
[stability] Allow pipeline stages to restart during exceptions

luxe:
### Problem:
We have a re-occurring issue where some of the workers temporarily lost their connection to redis.  This caused the worker's "match stage" to throw an exception and exit the thread (i.e. exit the specific pipeline stage).  The worker itself remained online.  All of the other pipeline stages kept running.  Without the "match stage" running those workers remain idle and underutilized.  

### Solutions:  
One solution to prevent this is to allow pipeline stages to continue after an exception.  It doesn't make sense for part of the pipeline to have ended due to an exception while the rest still run.  In order for the worker to self-heal, we propose logging the issue, but restarting the stage to avoid the worker remaining in an unhealthy state.  The pipeline stages can still end normally or through interrupt.  This is only to handle the unexpected exception being thrown.


src/main/java/build/buildfarm/worker/PipelineStage.java
```diff
@@ -47,24 +48,35 @@ private void runInterruptible() throws InterruptedException {
...
+
+      } catch (Exception e) {
+        runStage = handleTermination(e);
       }
     }
+
+    close();
+  }
+
+  private boolean handleTermination(Exception e) {
```
werkt, line 65: Document what the return value means here.

werkt (changes requested):
The intent of the Pipeline ownership and construction was that if any stage feeding or targetted by the current stage was shut down, it would make sense to perform an orderly shutdown of the affected stage, presumably at empty.

I would have expected this circumstance to be addressed by such monitoring, but I may have only done it for downstream, not upstream stages. However, I'd also expect logic like this to belong in the Pipeline class, and I'm very curious to see if this works for the orderly shutdown request.

This needs unit tests before approval as well. You've cited an excellent example in the description, now let's see it proven with a test.

werkt (commented):
Ping on this. Still no tests here

src/test/java/build/buildfarm/worker/PipelineTest.java
```diff
@@ -71,4 +76,94 @@ public void run() {
     pipeline.start();
     pipeline.join();
   }
+
+  // Create a test stage that exists because of an interrupt.
```
werkt, line 80: that exits?

luxe: ty!

src/main/java/build/buildfarm/worker/PipelineStage.java
luxe: improved

luxe:
I know this change may not be exactly what you want for pipeline behavior.  For us, its sort of a stop gap to improve cluster stability.  Since redis going away can cause the match stage to stop, and we've had the report results stage stop because of an exception.  The core problems are indeed  outside of the pipeline, but its helpful to have the exceptions logged and to keep workers processing.

werkt:
My problem is that any unchecked exception thrown during pipeline stage execution could (i.e. IMHO is likely) cause serious disruption to the state of the worker, and render it untrustworthy. This change means that no amount of misbehavior by a stage can result in shutdown, which is a very [systemically] dangerous pattern for a trusted context executing processes on behalf of users producing blindly used cached outputs into shared storage that they may not have the ability to recover from. My problem, in essence, is not the exception, which I want to make sure I hear about because it means that we cannot cope with some new normal input, environment, or condition, but is instead the fact that we're trying to continue on after living with whatever condition created the unchecked exception in the first place. They are bugs, and you're hiding them as a consequence of this.

luxe:
Then we will need to ensure that the pipeline properly shuts down when a stage fails and that the worker also shuts down when a pipeline fails.

Note to self, revisit this issue from the perspective of graceful shutdown.

PR #1042, opened 2022-03-17
Add priority queue 

krisstakos:
Added priority queue implemented with the help of [ordered sets](https://redis.io/commands/zadd) in the lua script
Not entirely sure how blocking operations work, so I'm open for suggestions (There are some left overs in the lua script)
Fixed all non-mock tests, also added new ones
Made a QueueInterface(with RedisQueueFactory) which will pick one of the  two types of queues (regular or priority). At the moment it's hard codded to use 'regular' (old behaviour). 
Later will add an option to set it through 'server.config/worker.config'

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
+++ b/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
@@ -96,11 +96,18 @@ public class BalancedRedisQueue {
    * @note Overloaded.
    */
   public BalancedRedisQueue(String name, List<String> hashtags) {
-    this.originalHashtag = RedisHashtags.existingHash(name);
-    this.name = RedisHashtags.unhashedName(name);
-    this.type = "regular";
-    this.maxQueueSize = -1; // infinite size
-    createHashedQueues(this.name, hashtags, this.type);
+    this(name, hashtags, -1, "regular");
+  }
+
+  /**
+   * @brief Constructor.
+   * @details Construct a named redis queue with an established redis cluster.
+   * @param name The global name of the queue.
+   * @param hashtags Hashtags to distribute queue data.
+   * @note Overloaded.
+   */
+  public BalancedRedisQueue(String name, List<String> hashtags, String queueType) {
+    this(name, hashtags, -1, queueType);
   }
 
   /**
@@ -112,9 +119,23 @@ public class BalancedRedisQueue {
    * @note Overloaded.
    */
   public BalancedRedisQueue(String name, List<String> hashtags, int maxQueueSize) {
+    this(name, hashtags, maxQueueSize, "regular");
+  }
+
+  /**
+   * @brief Constructor.
+   * @details Construct a named redis queue with an established redis cluster.
+   * @param name The global name of the queue.
+   * @param hashtags Hashtags to distribute queue data.
+   * @param maxQueueSize The maximum amount of elements that should be added to the queue.
+   * @param queueTYpe Type of the queue
+   * @note Overloaded.
+   */
+  public BalancedRedisQueue(
+      String name, List<String> hashtags, int maxQueueSize, String queueType) {
     this.originalHashtag = RedisHashtags.existingHash(name);
     this.name = RedisHashtags.unhashedName(name);
-    this.type = "regular";
+    this.type = queueType;
     this.maxQueueSize = maxQueueSize;
     createHashedQueues(this.name, hashtags, this.type);
   }
--- a/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
+++ b/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
@@ -43,7 +43,12 @@ public class BalancedRedisQueue {
    */
   private final String name;
 
-  private final String type;
+  /**
+   * @field queueType
+   * @brief Type of the queue.
+   * @details It's used for selecting between regular and priority queues
+   */
+  private final String queueType;
 
   /**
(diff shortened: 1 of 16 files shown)


src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
```diff
@@ -68,7 +70,9 @@
    * @details Although these are multiple queues, the balanced redis queue treats them as one in its
    *     interface.
    */
-  private final List<RedisQueue> queues = new ArrayList<>();
+  private final List<QueueInterface> queues = new ArrayList<>();
+
+  // private final List<QueueInterface> queuesTest;
```
luxe, line 75: remove

src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
```diff
@@ -96,8 +100,9 @@
   public BalancedRedisQueue(String name, List<String> hashtags) {
     this.originalHashtag = RedisHashtags.existingHash(name);
     this.name = RedisHashtags.unhashedName(name);
+    this.type = "regular";
```
luxe, line 103: this is good. we'll turn it into a protobuf enum later or something.  And this will have no performance impact since the dispatching is only on object construction.  existing users should not notice.

src/main/java/build/buildfarm/common/redis/QueueInterface.java
```diff
@@ -0,0 +1,125 @@
...
+ *     persists before and after the queue data structure is created (since it exists in redis).
+ *     Therefore, two redis queues with the same name, would in fact be the same underlying redis
+ *     queue.
+ */
+public abstract class QueueInterface {
+
+  // void push(JedisCluster jedis, String val, double priority);
+
+  /**
+   * @brief Push a value onto the queue with default priority of 1.
```
luxe, line 34: yes, that's what I was thinking too.  Its weird combining a non-priority queue with a priority queue under the same interface, but it makes the most sense we do this for migration.

src/main/java/build/buildfarm/common/redis/QueueInterface.java
```diff
@@ -0,0 +1,125 @@
...
+ * @brief A redis queue interface.
+ * @details A redis queue is an implementation of a queue data structure which internally uses redis
+ *     to store and distribute the data. Its important to know that the lifetime of the queue
+ *     persists before and after the queue data structure is created (since it exists in redis).
+ *     Therefore, two redis queues with the same name, would in fact be the same underlying redis
+ *     queue.
+ */
+public abstract class QueueInterface {
+
+  // void push(JedisCluster jedis, String val, double priority);
```
luxe, line 31: remove

src/main/java/build/buildfarm/common/redis/RedisQueue.java
```diff
@@ -54,6 +54,15 @@ public void push(JedisCluster jedis, String val) {
     jedis.lpush(name, val);
   }
 
+  /**
+   * @brief Push a value onto the queue.
+   * @details Adds the value into the backend redis queue.
+   * @param val The value to push onto the queue.
+   */
+  public void push(JedisCluster jedis, String val, double priority) {
+    throw new IllegalArgumentException("That's not a priority queue");
```
luxe, line 63: actually, can you do a regular push here instead.  Simply ignore the priority.
We'll be forwarding the priority queue data in later.  It will be easier for us to migrate data structures when the client is already calling this API.

src/test/java/build/buildfarm/common/redis/BalancedRedisQueueMockTest.java
```diff
@@ -105,39 +107,39 @@ public void removeFromDequeueTrueWhenValueExists() throws Exception {
...
-    String val = queue.dequeue(redis);
+  //   // ACT
+  //   String val = queue.dequeue(redis);
 
-    // ASSERT
-    assertThat(val).isEqualTo("foo");
-  }
+  //   // ASSERT
+  //   assertThat(val).isEqualTo("foo");
+  // }
```
luxe, line 142: I assume these were already broken because we don't run redis integration tests in the CI.  I'll fix them later.

src/test/java/build/buildfarm/common/redis/BalancedRedisQueueMockTest.java
```diff
@@ -359,7 +362,7 @@ public void canQueueIfiniteQueueAllowsQueuing() throws Exception {
   @Test
   public void canQueueFullQueueNotAllowsQueuing() throws Exception {
     // MOCK
-    when(redis.llen(any(String.class))).thenReturn(123L);
```
luxe, line 362: I think these changes are only valid when the queue is configured as ""priority",  
The default is "regular", and it can only be changed by changing the code.  

We'll run the redis tests in the CI next and that will clear up any differences.

luxe (approved):

src/test/java/build/buildfarm/common/redis/BalancedRedisQueueMockTest.java
krisstakos: Yeah, I fixed that. We now can speciffy through BalancedRedisQueue constructor witch queue to use

src/test/java/build/buildfarm/common/redis/BalancedRedisQueueMockTest.java
krisstakos: Fixed that

PR #1043, opened 2022-03-19
[ci] gate on redis integration tests

luxe:
Unit tests that rely on redis are mocked and run in the CI.  However, there are additional tests that talk directly to redis that don't run in the CI. It would be helpful to start running these integration tests now to avoid unexpected issues. 
Considering we already run a full cluster with redis as an end-to-end integration test.  Using a redis container to run these additional unit tests seems acceptable.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -67,6 +67,14 @@ tasks:
     shell_commands:
     - ./bazelw build //src/main/java/build/buildfarm/rpms/server:buildfarm-server-rpm //src/main/java/build/buildfarm/rpms/worker:buildfarm-worker-rpm
 
+  # This will run unit tests that integrate with redis.
+  # Outside of this step, similar functionality may be tested with mocks.
+  redis_unit_tests:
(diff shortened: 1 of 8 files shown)


werkt (approved):
Looks good, curious why you had to add the hash to the sizeAdjustPushDequeue

luxe:
> Looks good, curious why you had to add the hash to the sizeAdjustPushDequeue

The underlying implementation is two redis lists and the dequeue method performs `brpoplpush`.  
Without the hashtag, jedis will throw: **No way to dispatch this command to Redis Cluster because keys have different slots**.
Maybe we should provide a default hashtag when missing? Luckily all the implementations that leverage the RedisQueue provide a hashtag on construction.

meteorcloudy:
Looks like this PR is causing breakage with Bazel@HEAD:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2406#4aa14e8b-7b35-4592-b5d2-60b597c453e5

It's possible the reason is that Bazel built at HEAD doesn't have a version number, can you please fix?

PR #1044, opened 2022-03-20
default_max_cores setting

werkt:
Deprecated the legacy limit_execution option, which was a composite of
several valid specifications.
Provided a default value for max-cores without specification and used
for a range of non-tests in only_multicore_tests.
Cleaned up execution cpu limit decisions which could have resulted in 0
claims, and conditionalized some core limiting debug logging.
Removed test conditions that varied from onlyMulticoreTests default of
false where the check was immaterial

### DIFF
--- a/examples/shard-worker.config.example
+++ b/examples/shard-worker.config.example
@@ -298,6 +298,19 @@ redis_shard_backplane_config: {
   #max_attempts: 10
 }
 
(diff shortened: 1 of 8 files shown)


luxe (approved):
Having `limit_execution` and `limit_global_execution` was confusing.  Glad to see that simplified.  Also glad to see more configuration available.

PR #1050, opened 2022-03-24
[refactor] merge duplicated parsing logic for queued operation

luxe:


src/main/java/build/buildfarm/common/ProtoUtils.java
```diff
@@ -0,0 +1,61 @@
...
+ * @class ProtoUtils
+ * @brief Utilities related to parsing proto data.
+ * @details Performs validation and error reporting of proto data.
+ */
+public class ProtoUtils {
+  private static final Logger logger = Logger.getLogger(ProtoUtils.class.getName());
+
+  public static QueuedOperation getQueuedOperation(
+      ByteString queuedOperationBlob, QueueEntry queueEntry)
+      throws IOException, InterruptedException {
```
werkt, line 36: This should not be throwing either of these exceptions.

src/main/java/build/buildfarm/common/ProtoUtils.java
```diff
@@ -0,0 +1,61 @@
...
+    Digest queuedOperationDigest = queueEntry.getQueuedOperationDigest();
```
werkt, line 37: Pull `operation_name` here as well, if you're going to try to reduce interaction with queueEntry. You could also just take a string identifier for the pair and construct/pass it externally, but it's string construction that is only used in the error case.

werkt (changes requested):

src/main/java/build/buildfarm/common/ProtoUtils.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @class ProtoUtils
+ * @brief Utilities related to parsing proto data.
+ * @details Performs validation and error reporting of proto data.
+ */
+public class ProtoUtils {
+  private static final Logger logger = Logger.getLogger(ProtoUtils.class.getName());
+
+  public static QueuedOperation getQueuedOperation(
```
werkt, line 33: One last nit. This is now `parseQueuedOperation` by convention.

werkt (approved):

src/main/java/build/buildfarm/common/ProtoUtils.java
luxe: fixed

src/main/java/build/buildfarm/common/ProtoUtils.java
luxe: fixed

src/main/java/build/buildfarm/common/ProtoUtils.java
luxe: fixed

PR #1056, opened 2022-03-29
[metrics] Add action result metadata logging

80degreeswest:
Add logging for action result metadata in a similar way we currently log execution metadata. This will be helpful in analyzing cache hits and can be used by external systems to display individual build results broken down by executions and cache hits for a specific target.

Example metadata for a cache hit:

`
{
"toolDetails":{
"toolName":"bazel",
"toolVersion":"4.2.2rc1"
},
"ActionId":"0d20c119b262b200ff78216b55bdf3120de0ac59",
"toolInvocationId":"051bd515-953b-4b04-b2f2-d34084a0cd96",
"correlatedInvocationsId":"e7306c7b-6621-4999-bca0-0468e6af4d02",
"ActionMnemonic":"CppLink",
"targetId":"@com_google_googletest//:gtest_main",
"configurationId":"95857d6db24aff313b24ef0c714c9790f4c76764ffeef41ea1f2d21a8e50138c"
}`


src/main/java/build/buildfarm/server/ActionCacheService.java
```diff
@@ -98,6 +103,7 @@ public void onFailure(Throwable t) {
         },
         directExecutor());
     actionResultsMetric.inc();
+    metricsPublisher.publishRequestMetadata(requestMetadata);
```
luxe, line 106: 1.  If we don't configure a metricsPublisher in our server config it defaults to using a log publisher.  That seems okay.  I was thinking there might be an option to disable it entirely, but we can also set a FINE log level in the `LogMetricsConfig` which will do the same thing.  I'm just double checking that this can be no-op if desired.

2. `requestMetadata` is only part of the full action request.  To get the full context you could combine `GetActionResultRequest` + `RequestMetadata`.  Then you'll have additional context about getting results inlined (which could have different performances and success rates, the instance name (which probably isn't that important to you).  I also see the metadata also has the actionId so the extra information from `GetActionResultRequest` might not be that important after all.

luxe (approved):
looks good.  I hope it helps you debug.  At some point we need to collect & send all this information to a DB.  Then we can query by innovation ID.  Get some kind of UI started to show all the contextual information about a build.  I assume your doing something similar with an aws logger, but eventually we'll move in the direction of having a db/ui upstream.

src/main/java/build/buildfarm/server/ActionCacheService.java
80degreeswest: For 1. Correct - you can just set the log level to FINE and never see the log message. For 2. I'd have to check and see if there is any useful information in the GetActionResultRequest. For my purpose, the RequestMetadata contains everything I need. We may need to balance the size of the data vs useful information.

80degreeswest:
> looks good. I hope it helps you debug. At some point we need to collect & send all this information to a DB. Then we can query by innovation ID. Get some kind of UI started to show all the contextual information about a build. I assume your doing something similar with an aws logger, but eventually we'll move in the direction of having a db/ui upstream.

I would actually prefer to not have buildfarm be responsible for managing a database. For example, there is an option to send these events to an SNS topic in aws and hand the ingestion of that data to an external service (which could be database). Otherwise we could be taking valuable resources from doing builds to managing data. We could always make a generic topic implementation that does not rely on AWS's sns or implement a google solution.

werkt:
I'm leery of this only because we explicitly removed logging and other per-action result request statistics previously due to performance. How is this low enough impact to justify it?

80degreeswest:
> I'm leery of this only because we explicitly removed logging and other per-action result request statistics previously due to performance. How is this low enough impact to justify it?

You are right I forgot about that. Let me think about it for a bit. I'll either put a follow up PR to split up the config or I may just revert the whole thing.

PR #1063, opened 2022-03-31
[executor] allow the docker executor to fetch container-images that start with docker://

luxe:
resolves: https://github.com/bazelbuild/bazel-buildfarm/issues/1060


src/main/java/build/buildfarm/common/StringUtils.java
```diff
@@ -0,0 +1,34 @@
+// Copyright 2022 The Bazel Authors. All rights reserved.
```
80degreeswest, line 1: Could just use an existing library for this instead of writing our own - org.apache.commons.lang3.StringUtils has this.

src/test/java/build/buildfarm/common/StringUtilsTest.java
```diff
@@ -0,0 +1,59 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
```
80degreeswest, line 1: Won't need an extra test if using an existing library :)

src/main/java/build/buildfarm/common/StringUtils.java
luxe: thank you! way smarter!

src/test/java/build/buildfarm/common/StringUtilsTest.java
luxe: thank you! way smarter!

80degreeswest (approved):

PR #1074, opened 2022-04-06
Assign different queue names per queue type 

krisstakos:
We hit a issue where our redis cluster has a persistent storage hence it keeps the data after restart/crash and when we tried to fast switch between regular/priority queue through the server/worker configs the hole infra crash. The crash was due to stored data which messes up the queue types in the redis.


luxe:
This looks good.  I'm a little worried about users who upgrade an existing cluster.  Their queue names will change despite using the same implementation.  If they upgrade all the nodes at the same time there shouldn't be an issue.  However, if they do a gradual rollout they will see problems.  Let me bring this up to others before landing.  Worst case,  we'll keep the regular queue names the same, but the priority implementation will specifically add the prefix.

luxe:
Okay, talked about this.  We need to preserve the original queue names.  Only add the _suffix if the type is priority

krisstakos:
> Okay, talked about this. We need to preserve the original queue names. Only add the _suffix if the type is priority

Sure, will change it soon

luxe (approved):

PR #1075, opened 2022-04-07
Propagate priority value from remote API

krisstakos:
In order to use priority queue, we had to propagate the priority value received from remote execution API through the new code
How to use it: https://docs.bazel.build/versions/main/command-line-reference.html#flag--remote_execution_priority 
Possible values for the priority: min:0(default) max:1 - the smaller the value the highest the priority


luxe (approved):
In the future we may consider making a separate queue type for the prequeue.  It's possible that we keep the prequeue regular while the operation queue is priority.

PR #1079, opened 2022-04-19
set allow_unmatched: True in worker example config.  Improve error message.

luxe:
The server example has `allow_unmatched: True` by default, but the worker does not.  The inconsistency will confuse users.  
see: https://buildteamworld.slack.com/archives/C9C4H1SN7/p1650405057441019

we'll adjust the example config and provide a more helpful error message.


shirchen (approved):
Thanks!

80degreeswest (approved):

PR #1082, opened 2022-04-25
[bugfix] resolve stuck action example

luxe:
https://github.com/bazelbuild/bazel-buildfarm/issues/578 shows an example of a unit test that works locally, but hangs on buildfarm.  The issue is that although the process ends, `ByteStringWriteReader.getData()` throws an `InterruptedException` and we were only handling the `IOException`.  We resolve this issue by refactoring the way stdout/stderr is extracted and ensure the example test still completes successfully.  

I'm not sure if this is the core issue related to action hangs, but by not handling the  `InterruptedException` I think we skip logic related to handling the poller which might be causing deadlocks in the future. 

With this change:
1. https://github.com/bazelbuild/bazel-buildfarm/issues/578 unit test will pass successfully.  
2. Infinite loop programs will still timeout and fail as expected (they get re-queued but will fail on requeue amount)

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -303,7 +303,6 @@ class Executor {
         operationContext.toBuilder().setOperation(operation).build();
     boolean claimed = owner.output().claim(reportOperationContext);
     operationContext.poller.pause();
-
     if (claimed) {
       try {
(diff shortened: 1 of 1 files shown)


80degreeswest (approved):
If this is in fact the root cause of the hangs we see it's going to be a huge fix. Looking forward to getting this tested in our environment.

PR #1084, opened 2022-04-30
[redis] honor redis connection configurations

luxe:
We have hard-coded min values in our JedisCluster initialization. This is is not necessary.  Let's honor what the user configures.  For zero values we will still set the appropriate defaults.  We will warn about these defaults being set in `config/ConfigAdjuster.java`.  This gets the specific hard-coded values out of the JedisClusterFactory


werkt (approved):
So be it. I'm not 100% confident that we can't go haywire with unspecified timeout/max_attempts here, but a test would solidify that.

werkt:
Putting up a revert for this. You're not checking whether the server config is a shard type before checking the timeout for redis, and the creation of a 2000 timeout default removes an active memory config.

Also, I don't see any call to the ShardInstanceConfig variety of adjust here, which may have been where that could have been more appropriate.

PR #1086, opened 2022-05-06
[execution pipeline] Define slot usage the same way between Fetch and Execute stages

luxe:
Group the intake queue, claims, and job pool together.  Make it a responsibility of the SuperScalar stage.  This simplifies the state management for the the fetch and execute state.  The metrics `input_fetch_slot_usage` and `execution_slot_usage` are now computed the same way.

This is a refactoring effort to get a better understanding of throughput of the worker pipeline.


src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -137,8 +124,8 @@ protected void iterate() throws InterruptedException {
     Thread executorThread = new Thread(() -> executor.run(limits));
 
     synchronized (this) {
-      executors.add(executorThread);
-      size = executorClaims.addAndGet(limits.cpu.claimed);
+      slots.jobs.add(executorThread);
+      int size = getSlotUsage();
```
werkt, line 128: Don't you need an executionSlotUsage.set here?

src/main/java/build/buildfarm/worker/Slots.java
```diff
@@ -0,0 +1,38 @@
...
+
+  // The total number of claims available between jobs.  If every job took 1 claim, the width would
+  // correspond to the total number of jobs that can run concurrently in a stage.
+  public int width = 0;
+
+  // The current number of claims currently held by all the jobs.  Keep in mind, that claims amount
+  // isn't necessarily job amount.
+  public Semaphore claims = new Semaphore(1);
+
+  // The number of concurrent jobs running.  Should this be a thread pool?
```
werkt, line 36: Yes, it should - that would be a better mechanism to manage these threads as well.

Nominally this could be a thread pool specialization (subclass), which already encapsulates the job mechanics, that just makes each submission have a cost.

src/main/java/build/buildfarm/worker/Slots.java
```diff
@@ -0,0 +1,38 @@
...
+  // claims.
+  public BlockingQueue<OperationContext> intake = new ArrayBlockingQueue<>(1);
+
+  // The total number of claims available between jobs.  If every job took 1 claim, the width would
+  // correspond to the total number of jobs that can run concurrently in a stage.
+  public int width = 0;
+
+  // The current number of claims currently held by all the jobs.  Keep in mind, that claims amount
+  // isn't necessarily job amount.
+  public Semaphore claims = new Semaphore(1);
```
werkt, line 34: switch this to width initialization in a constructor per below comment about Slots usage.

src/main/java/build/buildfarm/worker/Slots.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.common.collect.Sets;
+import java.util.Set;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Semaphore;
+
+public class Slots {
+  // Each stage has an input context queue.  These contexts are evaluated to become jobs that take
+  // claims.
+  public BlockingQueue<OperationContext> intake = new ArrayBlockingQueue<>(1);
```
werkt, line 26: This still makes me queasy - I wish there was a better representation of our interchange here.

src/main/java/build/buildfarm/worker/SuperscalarPipelineStage.java
```diff
@@ -14,22 +14,16 @@
...
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
 
 abstract class SuperscalarPipelineStage extends PipelineStage {
-  protected final int width;
-
   @SuppressWarnings("rawtypes")
-  protected final BlockingQueue claims;
+  protected Slots slots;
```
werkt, line 23: Make this final, and would prefer initialization with width parameterization - don't use class just for containerization.

luxe: fixed

src/main/java/build/buildfarm/worker/Slots.java
luxe: I couldn't think of a better construct either.  I also wonder if anything > 1 would make a meaningful difference in behavior

src/main/java/build/buildfarm/worker/Slots.java
luxe: fixed

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
luxe: We could.  Currently we only capture the metric on slot release.  This was probably done to avoid computing the metric twice as much.  InputFetchStage does the same thing.

src/main/java/build/buildfarm/worker/Slots.java
luxe: I'll experiment with a threadpool after.  For now, this keeps the same performance / behavior we had before.

luxe:
@werkt re-review?

PR #1091, opened 2022-05-07
Ensure visibility for fetchInputs exception

werkt:
The finally will hide, rather than suppress, exceptions thrown during
fetchInputs and output stamping, if decrementReferences or
destroyExecDir throws.


luxe (approved):

luxe:
Landing as this is a net positive.  We can followup by combining memory/shard behavior.

PR #1095, opened 2022-05-10
[backplane] rename dispatchedOperations to dispatchingOperations. Containerize dispatchedOperations 

krisstakos:
There is a difference between "dispatched" operations and "dispatching" operations. The previous redis container was incorrectly named. We fix it by renaming instances of dispatchedOperations to dispatchingOperations. And we containerize all the jedis calls that actually are for dispatched operations.

This is to continue the refactoring effort of the RedisShardBackplane so that containers can be more easily swapped with different implementations.
- luxe


werkt (approved):
Hard for me to follow the logic here, but hopefully it checks out.

PR #1098, opened 2022-05-11
[refactor] distributed state in backplane

luxe:
Move distributed state containers and their creation to modules outside the backplane.  
This is to continue the effort on simplifying the backplane and allowing different backends


krisstakos:
That looks good to me

PR #1102, opened 2022-05-17
inline lua logic for priority queue

luxe:
Discussion with @80degreeswest on using the priority queue:  It would be easier for our deployments if we do not have to worry about bundling lua scripts alongside our deployed jars.  We solve this by inline the lua logic in java.  This will avoid any potential runtime failure to find the lua code.


80degreeswest (approved):
Per our conversation in Slack, should be ok to land now but implement a with Jedis pipeline at a later time.

PR #1104, opened 2022-05-17
allow build clients to send priority values larger than 1

luxe:
The priority queue implementation can handle values above 1.  We should adjust the client range to allow passing higher values.
Otherwise clients will see: `ERROR: --remote_execution_priority 10 is outside of server supported range 0-1.`


80degreeswest (approved):
Thanks!

PR #1109, opened 2022-05-26
Add cache regression test

krisstakos:
DO NOT MERGE till that one is not on upstream(https://github.com/bazelbuild/tools_remote/pull/27) 
Calculate latency times, number of remote executions (up/down) and CAS operations (up/down)
Sample output:

> Remote executions made: 300
> CAS uploads made: 2514
> CAS downloads made: 1100
> Action Cache requests made: 300
> Method: google.bytestream.ByteStream/Write
> 	Min: 4 ms
> 	Max: 394 ms
> 	Avg: 22 ms
> Method: build.bazel.remote.execution.v2.ContentAddressableStorage/FindMissingBlobs
> 	Min: 2 ms
> 	Max: 59 ms
> 	Avg: 5 ms
> Method: build.bazel.remote.execution.v2.Capabilities/GetCapabilities
> 	Min: 254 ms
> 	Max: 254 ms
> 	Avg: 254 ms
> Method: build.bazel.remote.execution.v2.Execution/Execute
> 	Min: 74 ms
> 	Max: 1835 ms
> 	Avg: 643 ms
> Method: build.bazel.remote.execution.v2.ActionCache/GetActionResult
> 	Min: 1 ms
> 	Max: 186 ms
> 	Avg: 7 ms
> Method: google.bytestream.ByteStream/Read
> 	Min: 2 ms
> 	Max: 485 ms
> 	Avg: 12 ms


.bazelci/ingest_grpc_logs.py
```diff
@@ -0,0 +1,75 @@
...
+
+def main():
+    n = len(sys.argv)
+    for i in range(1, n):
+        print("=====================")
+        print("Build #%d results" % i)
+        methods = defaultdict(list)
+        parsed = parse_file(sys.argv[i])
+        count_remote_actions(parsed, methods)
+        calculate_latency(methods)
```
luxe, line 23: looks good.  
For the CI can we also check that the 2nd grpc log makes 0 remote execution requests and 0 CAS uploads.  If the second log makes those requests it would mean something is wrong with our caching and we have a regression.  If the 2nd log makes those requests we can have the script fail and then hopefully the CI plan fail too.  

something like:  
```
if i == 2:
    ensure_fully_cached_build(parsed, methods)
```
Or whatever you think is cleaner

krisstakos: Hi, Trevor. It already does that. It will print out build:1 and build:2 results which include what you've described. Or you are talking for explicit check that the second run we for sure get 0 remote execution.

luxe: Yeah, I'd like to gate on the 2nd log's results it to avoid regressions. I'm going to refactor the ActionCache and CAS in the future and want to make sure they preserve 100% of cache under this circumstance in the CI.

luxe: Printing out the build:1 build:2 results is still helpful and will probably be used by clients.  We can glance at it in the CI as well.  But I want the CI step to fail if we are rebuilding when we shouldn't.

luxe (approved):
Thanks!  After the [PR](https://github.com/bazelbuild/tools_remote/pull/27)  for tools_remote lands, we should be good.

.bazelci/test_buildfarm_container.sh
```diff
@@ -98,6 +101,29 @@ start_server_and_worker(){
     fi
 }
 
+init_grpc_parser(){
+    if [ "${CACHE_TEST:-false}" = true ]; then
+        echo "Fetch tools_remote from git"
+
+        git clone https://github.com/bazelbuild/tools_remote.git
```
luxe, line 108: We can make this reference a specific SHA256 after the tools_remote change lands.

krisstakos: Yeah, we can!

PR #1128, opened 2022-07-16
containerize operations in backplane

luxe:
This continues the effort of moving individual jedis calls behind a container interface.  This will make it easier to swap out redis with alternative implementations.


src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
```diff
@@ -100,6 +100,10 @@ void verifyChangePublished(JedisCluster jedis) throws IOException {
     assertThat(opChange.getReset().getOperation().getName()).isEqualTo("op");
   }
 
+  String OperationName(String name) {
```
werkt, line 103: nit: dislike the methods-with-proper-name capitalization

src/main/java/build/buildfarm/common/redis/RedisMap.java
```diff
@@ -104,12 +108,32 @@ public void remove(JedisCluster jedis, Iterable<String> keys) {
...
+   * @param keys The name of the keys.
+   * @return The values of the keys. null if key does not exist.
+   * @note Overloaded.
+   * @note Suggested return identifier: values.
+   */
+  public List<Map.Entry<String, String>> get(JedisCluster jedis, List<String> keys) {
+    JedisClusterPipeline p = jedis.pipelined();
+    List<Map.Entry<String, String>> values = new ArrayList(keys.size());
+    for (String key : keys) {
+      values.add(new AbstractMap.SimpleEntry<>(key, p.get(createKeyName(key)).get()));
```
werkt, line 131: This has the effect of syncing for each item. I don't think this is what you want.

I recommend an Iterable response, constructed after a single sync, of a transform of a response set.

This will mostly devastate performance as is, as you're negating all of the purpose of the pipeline.

werkt (changes requested):
One big breaker in RedisMap.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -981,7 +976,7 @@ public boolean putOperation(Operation operation, ExecutionStage.Value stage) thr
     String name = operation.getName();
     client.run(
         jedis -> {
-          jedis.setex(operationKey(name), config.getOperationExpire(), json);
+          state.operations.insert(jedis, name, json, config.getOperationExpire());
```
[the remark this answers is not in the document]
luxe: fixed

src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
luxe: fixed

src/main/java/build/buildfarm/common/redis/RedisMap.java
luxe: what do you think about this change?  
I want to avoid returning the Response type to the caller since its a redis specific type.

werkt: It's more correct, but looks really clunky. This should have a much cleaner collapse into an Iterable (return an iterable, and take an iterable as keys to boot) conventionally, rather than a List (the list interface connotes more than we want to imply here). Comprehend to a Map via a stream, then transform the map values and return its items.

src/main/java/build/buildfarm/common/redis/RedisMap.java
```diff
@@ -35,13 +40,32 @@ public class RedisMap {
...
+  }
+
+  /**
+   * @brief Constructor.
+   * @details Construct a named redis map with an established redis cluster.
+   * @param name The global name of the map.
+   */
+  public RedisMap(String name, int timeout_s) {
+    this.name = name;
+    this.defaultExpiration_s = timeout_s;
```
werkt, line 68: nit: call the verbose constructor with the terse one and that 86400 default. And this isn't the 'default' expiration, it's the current one, worthy of just being called 'expiration.'

luxe: fixed

src/main/java/build/buildfarm/common/redis/RedisMap.java
luxe: I've switched the input and return type to Iterable.  However, the implementation does not return a lazily list to the caller-- I think that's something to improve.

luxe:
@werkt re-review?

werkt (approved):
Looks pretty good, no obvious red flags for me

PR #1136, opened 2022-08-10
Apple based execution wrapper

jerrymarino:
Implement an execution wrapper for macOS to set `SDKROOT` and `DEVELOPER_DIR`

In Apple based builds - this is required by clang, llvm, and related
native compilers and linkers.

Because there is no way to govern where Xcode.app is installed locally
or remotly it must be detected by Bazel. Because isn't possible to
assume the remote path is identical to the local path a buildfarm worker
must solve for this. We either query launchd where it was installed (
using Bazel's `xcode-locator` on darwin) _or_ map to the directory based
on convention. I added documentation about this problem and how it's
solved in the wrapper incase folks are new to this.

At the time of writing there is no notion of a macOS worker in this repo
 so I didn't add build rules for this yet. We don't have docker on
macOS but, I'd propose making a tarball rule longer term if people are
on-board with this idea. We can work towards a reference and canonical
example to build and release a worker end to end including, config, and
wrapper, and xcode-locator.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -38,6 +38,7 @@ java_library(
         ":as-nobody",
         ":delay",
         ":linux-sandbox.binary",
+        ":macos-wrapper",
         ":process-wrapper.binary",
         ":skip_sleep.binary",
         ":skip_sleep.preload",
@@ -106,6 +107,11 @@ genrule(
 sh_binary(
     name = "delay",
     srcs = ["delay.sh"],
+)
+
+sh_binary(
+    name = "macos-wrapper",
+    srcs = ["macos-wrapper.sh"],
 )
 
 # Docker images for buildfarm components
--- a/macos-wrapper.sh
+++ b/macos-wrapper.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# An Apple specific Apple execution wrapper - mainly to set SDKROOT and
+# DEVELOPER_DIR.
+# There is no enforcement of where Xcode.app is installed locally or remote so
+# isn't possible to assume the remote path is identical to the local path.
+
+# In "Apple" actions we will see the following ENV - but Bazel has to read the
+# Xcode config to map this to DEVELOPER_DIR and SDKROOT.
+# XCODE_VERSION_OVERRIDE=13.0.0.13A233
+# APPLE_SDK_PLATFORM=iPhoneSimulator
+# APPLE_SDK_VERSION_OVERRIDE=15.0
+# PWD=/private/tmp/worker-macos/shard/operations/6d9822e7-29b8-451b-908e-41b09250536f
+
+set -e
+
+# When given a XCODE_VERSION_OVERRIDE locate the specific Xcode and export
+if [[ -n "${XCODE_VERSION_OVERRIDE:-}" ]]; then
+    # For DEVELOPER_DIR - we need to map from how the host is configured to this
+    # path.
+    OVERRIDE_PATH=""
+
+    # Store the version file adjacent to the worker to avoid re-querying launchd
+    # for a given version.
+    CACHED_VERSION_FILE="$(dirname $(dirname "$PWD"))/$XCODE_VERSION_OVERRIDE.info"
+    if [[ -f "$CACHED_VERSION_FILE" ]]; then
+        OVERRIDE_PATH="$(cat $CACHED_VERSION_FILE)"
+    else
+        SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+        # Assuming the user has installed xcode-locator adjacent to the wrapper
+        # https://github.com/bazelbuild/bazel/blob/master/tools/osx/BUILD
+        # note: This program is highly tied to darwin
+        if [[ -f "$SCRIPT_DIR/xcode-locator" ]]; then
+            VERSION=$($SCRIPT_DIR/xcode-locator -v 2> /dev/null | grep ${XCODE_VERSION_OVERRIDE})
+            VERSION_PARTS=(${VERSION//:/ })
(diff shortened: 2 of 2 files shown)


kkpattern:
Thanks for your great job. This can help us a lot in building iOS applications with buildfarm. Is there any documentation about how to use this wrapper? We can help test this patch.

kkpattern:
We enabled this wrapper in config:

```bash
execution_policies: {
  wrapper: {
    path: "/path/to/macos_wrapper.sh"
  }
}
```

works like a charm for us. We only enabled `CppCompile` and `ObjcCompile` for remote execution. So far both work very well. Thanks!

luxe:
we can add this to the existing execution wrappers so that its made available in containers.
see: https://github.com/bazelbuild/bazel-buildfarm/blob/a8ec55703b5e94663accdcd3586c9178a951be50/BUILD#L36

luxe (approved):

kkpattern:
Currently, the action will fail if the required Xcode is not installed on the worker's platform. It would be nice if the worker could skip this action; other workers with the required Xcode installed can finish it then. It would be very helpful during an Xcode version updating across a relatively large build fleet. But I assume this requires a mac worker then? Maybe another PR.

kkpattern:
https://github.com/kkpattern/bazel-buildfarm/commit/bc331988f2c58feaf1ecb7ef4b6f5d3cb17d125b I did a very quick and dirty implementation and it works for us.

jerrymarino:
Glad you like it - I've been using this extenvily with iOS so far! Sorry for the delay on responding to this.

> we can add this to the existing execution wrappers so that its made available in containers. see:
> 
> https://github.com/bazelbuild/bazel-buildfarm/blob/a8ec55703b5e94663accdcd3586c9178a951be50/BUILD#L36

Cool - I'll push that here real quickly.

The only issue Apple related issue that popped for me on Apple Silicon based virtual machine execution. On the first launch, their `gatekeepr` subsystem will end up having a timeout. I have a pretty bad hack, but I'm  going to dig more into it to see if we can get the fix elsehwhere:
```
# On Apple Silicon - we see these processes being terminated by gatekeeper
# timeouts. Basically this code detects an immediate failure ( where gatekeeper
# terms it ) and then retries it. We _only_ retry a sigkill  / 9 - it
# subtracts 128 from the exit code to get the signal value
SIG=$(( $ST - 128 ))
if [[ $SIG -eq 9 ]]; then
    if [[ "${RBE_EXEC_LOGGING:-}" == "true" ]]; then
        >&2  echo "[WARN][Gatekeeper-retry] $(hostname) - $@" 
    fi
    "$@" 
    ST=$?
fi
```
I'm going to leave this here as a note - and we can followup with it if other people have the issue.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -251,7 +251,7 @@ public class ShardInstance extends AbstractServerInstance {
   private static BuildfarmConfigs configs = BuildfarmConfigs.getInstance();
 
   // TODO: move to config
-  private static final Duration queueTimeout = Durations.fromSeconds(60);
+  private static final Duration queueTimeout = Durations.fromSeconds(42);
```
luxe, line 254:
```suggestion
  private static final Duration queueTimeout = Durations.fromSeconds(60);
```

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -1256,7 +1256,7 @@ public CompletableFuture<Directory> apply(Digest digest, Executor executor) {
   protected <T> ListenableFuture<T> expect(
       Digest digest, Parser<T> parser, Executor executor, RequestMetadata requestMetadata) {
     Context.CancellableContext withDeadline =
-        Context.current().withDeadlineAfter(60, SECONDS, contextDeadlineScheduler);
+        Context.current().withDeadlineAfter(99, SECONDS, contextDeadlineScheduler);
```
luxe, line 1259:
```suggestion
        Context.current().withDeadlineAfter(60, SECONDS, contextDeadlineScheduler);
```

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -2019,7 +2019,7 @@ private ListenableFuture<Boolean> checkCacheFuture(
     }
 
     Context.CancellableContext withDeadline =
-        Context.current().withDeadlineAfter(60, SECONDS, contextDeadlineScheduler);
+        Context.current().withDeadlineAfter(99, SECONDS, contextDeadlineScheduler);
```
luxe, line 2022:
```suggestion
        Context.current().withDeadlineAfter(60, SECONDS, contextDeadlineScheduler);
```

luxe (approved):

luxe:
thanks for adding to `execution_wrappers`.  We should land this.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
jerrymarino: It looks like this got pushed in a frenzy - sorry for the confusion 🏃

jerrymarino:
@luxe you got it! Thanks again for the review ( and fixing the misc changes ) 👍 . Yeah it should be ready to land on my end, probably better to do the others as followups

luxe (approved):

luxe:
@jerrymarino I'm not sure why,  but github won't let me land this with merge button.  It just says "This branch is out-of-date with the base branch".  Maybe we should squash commits?

luxe (approved):

jerrymarino:
Yeah the github UI looked like it was in a broken state on my side too, I just rebased / squashed everything again, please LMK if you have other issues with it.

PR #1139, opened 2022-08-11
Report slot usage metrics on every change

bozaro:
Originally, slot usage metrics were only reported when the slot was released.
In this case, for example, we got a constant value of zero for the `input_fetch_slot_usage` metric with `input_fetch_stage_width: 1`


luxe (approved):
ty.  I've noticed the same thing. This should make the metric appear accurately

PR #1142, opened 2022-08-17
update rules_k8s version

Wyverald:


Wyverald:
gentle ping. Buildfarm remains broken in the downstream pipeline, and if we can't resolve that soon, we'd need to disable Buildfarm on Bazel CI.

luxe (approved):

luxe:
sorry for delay

PR #1143, opened 2022-08-17
[admin] Enable worker indexer scheduler

80degreeswest:
This PR adds buildfarm indexer scheduler to Buildfarm Admin. To enable the scheduler add/modify the following configs to BFAdmin's application.properties. By default, the scheduler is disabled.

buildfarm.scheduled.indexer=true
buildfarm.scheduled.indexer.cron=0 0 0 * * *

### DIFF
--- a/admin/main/src/main/resources/application.properties
+++ b/admin/main/src/main/resources/application.properties
@@ -19,3 +19,7 @@ springdoc.swagger-ui.operationsSorter=method
 
 # REST API Security (TODO: CHANGE IT!!!)
 security.http.auth-token=CHANGEME-1234567890
+
+# Scheduled Tasks
+buildfarm.scheduled.indexer=false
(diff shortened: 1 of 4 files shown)


admin/main/src/main/resources/application.properties
```diff
@@ -19,3 +19,7 @@ springdoc.swagger-ui.operationsSorter=method
 
 # REST API Security (TODO: CHANGE IT!!!)
 security.http.auth-token=CHANGEME-1234567890
```
luxe, line 21: changing this?

luxe (approved):
glad you could get this working.  It would be nice if buildfarm itself could maintain the indexes, but bf admin cronjob is good step forward

admin/main/src/main/resources/application.properties
80degreeswest: It's just a placeholder value.

PR #1144, opened 2022-08-25
Add cas indexer results prometheus metrics

80degreeswest:
Emit prometheus metrics for CAS indexer results per each Redis shard:

- Number of hosts removed from keys (cas_indexer_removed_keys)
- Number of keys removed due to no workers containing them (cas_indexer_removed_hosts)

Above metrics contain a "node" label to separate these metrics per Redis shard.


luxe (approved):
nice

PR #1148, opened 2022-08-29
Separate indexer results by node vs accumulating results

80degreeswest:
Previously, indexer results were accumulating for each node indexed. This change separates the results so they are reported individually per node and not accumulate as the indexer progresses. The end results are still logged as cumulative.


luxe (approved):
ya that makes more sense when looking at labels

jacobmou (approved):

PR #1150, opened 2022-09-01
80degreeswest/convert configs to yaml v2.0

80degreeswest:
Update protobuf configs to yaml. This PR will become a start of v2.0 and will break compatibility with v1.x.
The benefits of moving to yaml configs is to simplify configuration, remove dependency on classes on each other and open up the possibility for future refactoring and simplifying of the codebase.

Ex configuration:

```
server:
  instanceType: MEMORY
  name: memory
worker:
  publicName: "localhost:8981"
  cas:
    type: MEMORY
memory:
  target: "localhost:8980"
  properties:
    - name: "cores"
      value: "1"
```

```
server:
  instanceType: SHARD
  name: shard
backplane:
  redisUri: "redis://localhost:6379"
  queues:
    - name: "cpu"
      properties:
        - name: "min-cores"
          value: "*"
        - name: "max-cores"
          value: "*"
worker:
  publicName: "localhost:8981"
```


_site/docs/architecture/content_addressable_storage.md
```diff
@@ -34,7 +34,7 @@ Since these implementations vary in complexity and storage semantics, a common i
 
 The memory CAS implementation is extremely simple in design, constituting a maximum size with LRU eviction policy. Entry eviction is a registrable event for use as a storage for the delegated ActionCache, and Writes may be completed asynchronously by concurrent independent upload completion of an entry.
 
-This is the example presentation of a CAS in the memory instance available [here](https://github.com/bazelbuild/bazel-buildfarm/blob/85c3fdaf89fedc2faee1172fab01338777de79a1/examples/server.config.example#L40-L45), but for reference, specification in any `cas_config` field for server or worker will enable the creation of a unique instance.
+This is the example presentation of a CAS in the memory instance available [here](https://github.com/bazelbuild/bazel-buildfarm/blob/85c3fdaf89fedc2faee1172fab01338777de79a1/examples/config.memory.yml#L40-L45), but for reference, specification in any `cas_config` field for server or worker will enable the creation of a unique instance.
```
jacobmou, line 37: probably address these doc changes in a followup PR because `config.memory.ym` file won't exist in the snapshot on a previous commit.

src/test/java/build/buildfarm/worker/shard/WorkerTest.java
```diff
@@ -14,21 +14,31 @@
...
+  }
+
+  @Test(expected = NullPointerException.class)
   public void missingWorkerRoot() throws ConfigurationException {
-    new Worker("test", ShardWorkerConfig.newBuilder().build());
+    new Worker("test");
   }
 
-  @Test(expected = ConfigurationException.class)
+  @Test(expected = NullPointerException.class)
```
luxe, line 40: It seems like preserving the ConfigurationException over NullPointer would be better

luxe (approved):
It's obviously a big change, but in terms of overall complexity not too bad.  There's many subtle usages of our configs so I wouldn't be surprised if some default value isn't exactly the same or our users catch some subtle issue on upgrade.  Nonetheless, I feel our CI integration tests do a good job of end-to-end testing so overall I'm good with landing.  I also assume you have internal deployments using v2 for more confidence.

I'd like to test that Tilt local development still works, but I will do that later and make any adjustments as needed.

I saw you had a follow-up diff for some more documentation so that's good too.  Obviously we should make the next buildfarm release explicit about these config changes and how users need to upgrade, etc.

_site/docs/architecture/content_addressable_storage.md
luxe: that's a good point.  Our generated documentation has no version picker.  Whatever lands on main will become the live site.

80degreeswest:
@luxe thanks for the review. A good chunk of the comments are already addressed in my follow up PRs which are ready to go right behind this one. I will take a note of some that are not and get those fixes in as well.

PR #1152, opened 2022-09-06
[storage worker] add storage worker type

jacobmou:
### Context:
To better support Build Without the Bytes, we plan to add a new type of worker `storage worker`, which will host CAS but no execution pipeline so that it will be able to store enough cache to avoid cache expiration. To be more specific, eventually, we want 1) execution worker be able to upload their cache to storage workers and 2) storage worker enforce a replication policy.

### This PR
The PR will add a formal definition of `WorkerType`, which will be inferred from the `Capabilities` of a configured shard worker.
1. Execute with no CAS --> `WorkerType.EXECUTE` --> `executeWorkers`
2. CAS with no Execute --> `WorkerType.STORAGE` --> `storageWorkers`
3. Execute and CAS --> `WorkerType.EXECUTE_AND_STORAGE` --> `executeAndStorageWorkers`

New worker maps `executeWorkers`, `storageWorkers` and `executeAndStorageWorkers` are created. A worker will be able to register itself to one of the pool of backplane based on its `WorkerType`


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -772,10 +772,20 @@ message ShardWorkerConfig {
   int32 input_fetch_deadline = 39;
 }
 
+enum WorkerType {
+  EXECUTE_ONLY = 0;
```
jacobmou, line 776:
```suggestion
  EXECUTE = 0;
```

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -772,10 +772,20 @@ message ShardWorkerConfig {
...
+
+  STORAGE_ONLY = 1;
```
jacobmou, line 778:
```suggestion
  STORAGE = 1;
```

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -939,6 +942,19 @@ public void run() {
...
 
+  private WorkerType determineWorkerType(WorkerCapabilities capabilities) {
+    boolean hasStorage = capabilities.getCas();
+    boolean hasExecute = capabilities.getExecution();
+
+    if (hasStorage && hasExecute) {
+      return WorkerType.EXECUTE_AND_STORAGE;
+    }
+    if (hasStorage) {
+      return WorkerType.STORAGE_ONLY;
```
jacobmou, line 953:
```suggestion
      return WorkerType.STORAGE;
```

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -939,6 +942,19 @@ public void run() {
...
+    }
+    return WorkerType.EXECUTE_ONLY;
```
jacobmou, line 955:
```suggestion
    return WorkerType.EXECUTE;
```

src/main/java/build/buildfarm/instance/shard/DistributedState.java
```diff
@@ -28,11 +28,27 @@
...
-   * @brief All of the workers register themselves to the cluster.
+   * @brief All of the execute-only workers register themselves to the cluster.
    * @details This is done to keep track of which machines are online and known by the rest of the
    *     cluster.
    */
-  public RedisHashMap workers;
+  public RedisHashMap executeWorkers;
+
+  /**
+   * @field workers
```
luxe, line 38:
```suggestion
   * @field storageWorkers
```

src/main/java/build/buildfarm/instance/shard/DistributedState.java
```diff
@@ -28,11 +28,27 @@
...
+   * @brief All of the storage-only workers register themselves to the cluster.
+   * @details This is done to keep track of which machines are online and known by the rest of the
+   *     cluster.
+   */
+  public RedisHashMap storageWorkers;
+
+  /**
+   * @field workers
```
luxe, line 46:
```suggestion
   * @field executeAndStorageWorkers
```

src/main/java/build/buildfarm/instance/shard/DistributedState.java
```diff
@@ -28,11 +28,27 @@
 public class DistributedState {
   /**
    * @field workers
```
luxe, line 30:
```suggestion
   * @field executeWorkers
```

luxe (approved):
thank you!  I know there is some follow-up work to do, but its good to get workers properly annotated and registered to the backplane first.  

I was thinking about having a single redis container for the workers (instead of 3).  Each element would contain worker properties about what type they are.  Having separate containers for now accomplishes what we need.  So if you see a need for combining the 3 redisHashMaps into a single redisHashMap in the future, I would be good with that also.

PR #1154, opened 2022-09-08
[scheduler] Bump buildfarm-server base image

jerrymarino:
This fixes a number of issues flagged in the corretto base


80degreeswest (approved):
Thanks! I actually had it on my list to bump the source images.

PR #1155, opened 2022-09-08
Add project lombok for boiler plate code maintenance

80degreeswest:
Add project lombok to allow for automated generation of getters/setters/toString/etc with simple annotations.


luxe (approved):
so clean!

PR #1163, opened 2022-09-12
Convert all relevant strings to enums

80degreeswest:
Convert all relevant strings to enums. Some of these used to be enums to begin with but were converted to strings during config conversion to yaml. This PR corrects this.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisQueueFactory.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisQueueFactory.java
@@ -14,6 +14,8 @@
 
 package build.buildfarm.common.redis;
 
+import build.buildfarm.common.config.yml.Queue;
+
 /**
  * @class RedisQueueFactory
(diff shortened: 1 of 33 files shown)


src/main/java/build/buildfarm/common/redis/RedisQueueFactory.java
```diff
@@ -23,9 +25,9 @@ public QueueInterface getQueue(String queueType, String name) {
     if (queueType == null) {
       return null;
     }
-    if (queueType.equalsIgnoreCase("regular")) {
+    if (queueType.equalsIgnoreCase(Queue.QUEUE_TYPE.standard.name())) {
```
jacobmou, line 28: Can we also make `BalancedRedisQueue.queueType` also using the new `QUEUE_TYPE` type instead of `String`?

jacobmou (approved):

PR #1169, opened 2022-09-15
call getFileStore once

liyuyun-lyy:
When we use Buildfarm in our project, we use cpu profiler and find out that it takes too much time to call getFileStore when walking file tree. We don't need to call it recursively, so I use double checked  locking to call it once. 
image

image


werkt:
This needs something of a redesign - the static fields will not work for this. Instead, we would do better with an object that constrains to a single fileStore if the lookup is too costly (which is a justifiable premise).

PR #1173, opened 2022-09-19
Fix default configs path

80degreeswest:
Adjust the default configs path to allow external inclusion of bazel-buildfarm into another repo (for CI use case).


src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
```diff
@@ -36,7 +36,7 @@ public static BuildfarmConfigs getInstance() {
     if (buildfarmConfigs == null) {
       buildfarmConfigs = new BuildfarmConfigs();
       try {
-        loadDefaultConfigs("examples/config.yml");
+        loadDefaultConfigs("../build_buildfarm/examples/config.yml");
```
jacobmou, line 39: wondering if we want to make the path an argument to the program so that people wont have to change it from time to time.

80degreeswest: This is for default path to the example config file which shouldn't change. The base math worked fine when building directly, but broke when I was trying to build and test as a dependency of anothe repo as bazel was attempting to load it from my custom named workspace and not the workspace of build_buildfarm. This corrects it for this case. Not sure if there is a better way to do it and if so I'd like to get it fixed properly vs this solution which seems a bit hacky.

jacobmou (approved):

src/test/java/build/buildfarm/server/BuildFarmServerTest.java
```diff
@@ -459,6 +459,7 @@ public void grpcMetricsOffByDefault() {
   @Test
   public void grpcMetricsEnabled() {
     configs.getServer().getGrpcMetrics().setEnabled(true);
+    System.out.println("CONFIGS: " + configs.toString());
```
jacobmou, line 462: delete?

jacobmou (approved):

luxe:
@80degreeswest close?

80degreeswest:
> @80degreeswest close?

No keep it open. Id like to get this working eventually so it works like this:

Use defaults from config.yaml in examples.
Override those defaults with a user provided yaml. Currently its not working and the whole yaml gets overridden (defaults clear out) which is the reason why the defaults are set in Java classes (not ideal).

werkt:
@80degreeswest can you move this to draft state if it's still being worked on?

PR #1178, opened 2022-09-26
Remove memory instance implementation

80degreeswest:
Remove memory implementation for instance and worker, leaving shard as a single and default implementation.


jacobmou (approved):

.bazelci/cache_test.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# This is a CI integation test for a typical "shard deployment" of buildfarm.
+# This is a CI integation test for a typical eployment of buildfarm.
```
jerrymarino, line 2: Minor - `deployment`

PR #1180, opened 2022-09-29
Remove server dependencies from worker

80degreeswest:
Currently there are server dependencies in the worker implementation. Server and worker should be independent of each other and any shared functionality reside in the common package. This change removes all server dependencies from worker and moves shared resources into common.

Config change: move prometheusPort from server to top level.


jacobmou (approved):
LGTM!

PR #1181, opened 2022-09-30
Provide fake-network execution property

### LINKED ISSUE #1176: The linux-sandbox cli doesn't set hostname to `localhost`
Looking at https://github.com/bazelbuild/bazel-buildfarm/blob/9b6ff2506554b7cac4c1e7864f44391314b6bf0e/src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java#L954-L989, we don't set `-H` anywhere.

From the `linux-sandbox` help:
```
  -H  if set, make hostname in the sandbox equal to 'localhost'
```

What would it take to add this? Can we add it unconditionally or does it need to be configurable?

In local bazel actions the `-H` option is controlled via `--sandbox_fake_hostname`.

werkt:
Invokes the sandbox with -H

Fixes #1176

### DIFF
--- a/_site/docs/execution/execution_properties.md
+++ b/_site/docs/execution/execution_properties.md
@@ -42,6 +42,9 @@ Tests that exceed their memory requirements will be killed.
 ### `linux-sandbox`
 **description:** Use bazel's linux sandbox as an execution wrapper.
 
+### `fake-hostname`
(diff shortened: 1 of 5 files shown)


80degreeswest (approved):
Thanks for getting this in!

PR #1188, opened 2022-10-06
Add Spring Framework to the worker

80degreeswest:
Add basic spring framework usage to the worker module. Refactor worker to make it simpler, by removing inner classes and moving admin and config helper functions into their respective packages.


src/main/java/build/buildfarm/admin/aws/AwsAdmin.java
```diff
@@ -47,17 +53,24 @@
...
+  private AmazonAutoScaling scale;
+  private AmazonEC2 ec2;
+  private AWSSimpleSystemsManagement ssm;
+
+  public AwsAdmin() {
+    String region = configs.getServer().getCloudRegion();
+    if (region != null) {
+      scale = AmazonAutoScalingClientBuilder.standard().withRegion(region).build();
+      ec2 = AmazonEC2ClientBuilder.standard().withRegion(region).build();
+      ssm = AWSSimpleSystemsManagementClientBuilder.standard().withRegion(region).build();
```
jacobmou, line 72: Do we want to log something when `region == null` ?

src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
```diff
@@ -46,4 +55,68 @@ public void loadConfigs(Path configLocation) throws IOException {
       buildfarmConfigs = yaml.load(inputStream);
     }
   }
+
+  public static BuildfarmConfigs loadServerConfigs(String[] args) {
```
jacobmou, line 59: Seems like there are just a lot of code duplication here. I feel we might want to create some helper functions.

src/main/java/build/buildfarm/common/config/Worker.java
```diff
@@ -63,4 +68,41 @@ public ExecutionPolicy[] getExecutionPolicies() {
...
+    return WorkerType.EXECUTE;
+  }
+
+  public Path getValidRoot() throws ConfigurationException {
+    addMissingRoot();
+    verifyRootConfiguration();
+    return Paths.get(root);
+  }
+
+  private void addMissingRoot() throws ConfigurationException {
```
jacobmou, line 88: How about `addRootIfMissing()`?

src/main/java/build/buildfarm/common/config/Worker.java
```diff
@@ -63,4 +68,41 @@ public ExecutionPolicy[] getExecutionPolicies() {
...
+    }
+    if (getCapabilities().isCas()) {
+      return WorkerType.STORAGE;
+    }
+    return WorkerType.EXECUTE;
+  }
+
+  public Path getValidRoot() throws ConfigurationException {
+    addMissingRoot();
+    verifyRootConfiguration();
```
jacobmou, line 84: I might misunderstand, but the order here seem to be wrong. 
`verifyRootConfiguration()` would verify if the configs for root is valid and `addMissingRoot` is creating the root if it's missing. It seems to me we want to do verify root option from config first first and then create

src/main/java/build/buildfarm/worker/shard/LocalCasWriter.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.protobuf.ByteString;
+import io.grpc.Status;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.UUID;
+
+class LocalCasWriter implements CasWriter {
```
jacobmou, line 19: The new class can use some java docs.

src/main/java/build/buildfarm/worker/shard/RemoteCasWriter.java
```diff
@@ -0,0 +1,183 @@
...
+import java.util.Iterator;
+import java.util.Random;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.logging.Level;
+import lombok.extern.java.Log;
+
+@Log
+public class RemoteCasWriter implements CasWriter {
```
jacobmou, line 33: Also need some docs

jacobmou (approved):
The refactor makes sense to me.

 I left some thoughts. 

This seems to be a great opportunity to add some java docs for new classes, but i understand that the PR is more focusing on refactor, so we can do that later. 

I dont get what benefits we get from using `@component`, maybe i should start my spring framework 101. And because of that, i would hope a little more details could be added to the PR summary. 

Not blocking on any comments above.

src/main/java/build/buildfarm/common/config/Worker.java
80degreeswest: The current verify method checks for both configuration error and missing dir on the filesystem which is why it's called after the adding of a directory. Perhaps it could be split into multiple verification methods.

80degreeswest:
> The refactor makes sense to me.
> 
> I left some thoughts.
> 
> This seems to be a great opportunity to add some java docs for new classes, but i understand that the PR is more focusing on refactor, so we can do that later.
> 
> I dont get what benefits we get from using `@component`, maybe i should start my spring framework 101. And because of that, i would hope a little more details could be added to the PR summary.
> 
> Not blocking on any comments above.

Yes, I agree on adding javadocs later. 

The @Component annotation lets Spring manage it.

PR #1195, opened 2022-10-14
Introduce Persistent Workers

wiwa:
Bazel Persistent Workers for Buildfarm
===
### Goal: 
Buildfarm workers will be able to spin up and execute actions on persistent workers,
similar to how Bazel itself uses persistent workers.

See: https://bazel.build/remote/persistent

In our testing, this sped up JVM builds (i.e. with mnemonics of `Javac`, `Scalac`, etc.) by ~70-90%.
Tested with a modified Bazel `5.3.0` binary: https://github.com/wiwa/bazel/commit/07d40cef1019dc510aacf49de22ae39831bd647e

The Bazel modification is based off of `--experimental_mark_tool_inputs` [design](https://github.com/bazelbuild/proposals/pull/219) (original issue [here](https://github.com/bazelbuild/bazel/issues/10091)), with a PR currently open: https://github.com/bazelbuild/bazel/pull/16362

We also use a modified Buildfarm (1.16.0): https://github.com/wiwa/bazel-buildfarm/tree/p/16pw  
Presumably, the changes needed to actually _use_ persistent workers would come after this PR.
___

We split persistent workers into two targets:  
`persistentworkers/src/main/java/persistent/common:persistent-common`  
`persistentworkers/src/main/java/persistent/bazel:bazel-persistent-workers`

`persisent-common` aims to provide an interface for general persistent worker processes,
as well as a sort of "code as documentation". Bazel's own `libworker` is not publicly visible: [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/worker/BUILD) and  [here](https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/BUILD)

`bazel-persistent-workers` implements `persistent-common` in the context of Bazel's persistent workers model.

Co-authored-by: Shane Delmore <shane@delmore.io>

### DIFF
--- a/persistentworkers/examples/src/main/java/adder/Adder.java
+++ b/persistentworkers/examples/src/main/java/adder/Adder.java
@@ -1,0 +1,50 @@
+package adder;
+
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+import persistent.bazel.processes.WorkRequestHandler;
+
+/**
+ * Example of a service which supports being run as a persistent worker
+ */
+public class Adder {
+
+  public static void main(String[] args) {
+    if (args.length == 1 && args[0].equals("--persistent_worker")) {
+      WorkRequestHandler handler = initialize();
+      int exitCode = handler.processForever(System.in, System.out, System.err);
+      System.exit(exitCode);
+    }
+  }
+
+  private static WorkRequestHandler initialize() {
+    return new WorkRequestHandler((actionArgs, pw) -> {
+      String res = work(actionArgs);
+      pw.write(res);
+      return 0;
+    });
+  }
+
+  private static String work(List<String> args) {
+    if (args.size() == 1 && args.get(0).equals("stop!")) {
+      System.err.println("exiting!");
+      System.exit(0);
+      return "";
+    } else if (args.size() != 2) {
+      System.err.println("Cannot handle args: " + ImmutableList.copyOf(args).toString());
+      System.exit(2);
+      return "";
+    }
+    else {
+      int a = Integer.parseInt(args.get(0));
+      int b = Integer.parseInt(args.get(1));
+      return String.valueOf(compute(a, b));
+    }
+  }
+
+  private static int compute(int a, int b) {
+    return a + b;
+  }
+}
--- a/persistentworkers/examples/src/main/java/adder/Adder.java
+++ b/persistentworkers/examples/src/main/java/adder/Adder.java
@@ -31,17 +31,13 @@ public class Adder {
     if (args.size() == 1 && args.get(0).equals("stop!")) {
       System.err.println("exiting!");
       System.exit(0);
-      return "";
     } else if (args.size() != 2) {
       System.err.println("Cannot handle args: " + ImmutableList.copyOf(args).toString());
       System.exit(2);
-      return "";
     }
-    else {
-      int a = Integer.parseInt(args.get(0));
-      int b = Integer.parseInt(args.get(1));
-      return String.valueOf(compute(a, b));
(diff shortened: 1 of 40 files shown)


persistentworkers/examples/src/main/java/adder/Adder.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private static String work(List<String> args) {
+    if (args.size() == 1 && args.get(0).equals("stop!")) {
+      System.err.println("exiting!");
+      System.exit(0);
+      return "";
+    } else if (args.size() != 2) {
+      System.err.println("Cannot handle args: " + ImmutableList.copyOf(args).toString());
+      System.exit(2);
+      return "";
```
werkt, line 38: Not much need for a return after the System.exit...

werkt (commented):
With all of your changes under persistentworkers, what is the integration mechanism for actual workers?

persistentworkers/examples/src/main/java/adder/Adder.java
wiwa: Interesting, IntelliJ(?) complained about lack of return after a `System.exit`. In any case, good catch!

wiwa:
Pushed a commit directly to the PR branch showing the integration.  

tl;dr: Add a new path in `worker/Executor.java`, kind of like `DockerExecutor`.
Only run it (`PersistentExecutor`) if the platform properties of `persistentWorkerKey` and `persistentWorkerCommand` are filled in. 
These are only filled in if the host `bazel` command is called with `--experimental_mark_tool_inputs`.  
Specifically, only `persistentWorkerKey` is filled in by mainstream Bazel.  
`persistentWorkerCommand` is our minor tweak to `bazel` in order to also pass along the information needed to start *any* persistent process. Otherwise, we would need to hardcode the processes for each `persistentWorkerKey` we need. 

If the default `dequeue_match_settings` is `accept_everything: True`, then the worker configuration does not need to change at all (at least on 1.16.0; testing on 2.0.0 soon).

Otherwise, the settings (for 1.16.0) would be:

```
dequeue_match_settings: {
  platform: {
    property: {
      name: "persistentWorkerKey"
      value: "*"
    }
    property: {
      name: "persistentWorkerCommand"
      value: "*"
    }
  }
}
```

i.e. accept any value for the key and command. 

This integration doesn't (or at least _shouldn't_) change anything for existing users.

wiwa:
@werkt Pinging again.

Although this is highly unlikely to be used at Twitter anymore, I think it would be pretty beneficial to JVM-based builds.

luxe:
This is awesome!  I'd really like to get it into main.

> tl;dr: Add a new execution pathway in worker/Executor.java, kind of like DockerExecutor.

Yeah, that would be the best integration imo.

> Although this is highly unlikely to be used at Twitter anymore

What changed that made this no longer feasible?

wiwa:
That integration is actually already implemented (in the different branch I linked).
Re: Twitter, I (and my collaborator @ShaneDelmore) were both affected by layoffs, which I _believe_ also removed most of the developer productivity teams at Twitter.

wiwa:
@werkt what's needed for this PR?

werkt:
Just gotta get to a review, sorry for the delay. Will try to do it tonight.

werkt (commented):
The change is still authored under you from twitter, without a CLA. Can you resubmit under a CLA-signed user?

ShaneDelmore:
I'll sign the cla under my normal github account.

werkt (approved):

PR #1199, opened 2022-10-20
Allow Redis password to be empty 

shirchen:
Fixing a small regression b/n 1.x and 2.x to treat "" password as empty.


80degreeswest (approved):

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -59,11 +59,14 @@ public static Supplier<JedisCluster> create() throws ConfigurationException {
               : configs.getBackplane().getRedisPassword(),
           createJedisPoolConfig());
     }
+    // support "" as redis password.
     return createJedisClusterFactory(
         parseUri(configs.getBackplane().getRedisUri()),
         configs.getBackplane().getTimeout(),
         configs.getBackplane().getMaxAttempts(),
-        configs.getBackplane().getRedisPassword(),
+        configs.getBackplane().getRedisPassword().isEmpty()
```
80degreeswest, line 67: You'll need to do a null check here as well. Just use this. Can you also use the same on line 57 above please.
```
import com.google.common.base.Strings;

Strings.isNullOrEmpty(configs.getBackplane().getRedisPassword()) ? null : configs.getBackplane().getRedisPassword(),
```

80degreeswest (changes requested):
Needs an extra null check and run formatter. Also, just go ahead and remove that extra comment - I think that was added to force rebuild.

80degreeswest (approved):

PR #1203, opened 2022-10-21
Use bazelisk for the formatter invocation

80degreeswest:
Switch to bazelisk was issed in the format.sh. Update to fix it.


luxe:
Agree.  I think all of the CI has been switched to use `bazel` now because bazel team likes to swap in different versions for upstream testing.  We were using bazelisk wrapper before, but it defeated the purpose of their tests since we always used the same version.

PR #1204, opened 2022-10-21
Add basic batch scripts to start server and worker

80degreeswest:
Add  two basic scripts to start buildfarm server and buildfarm worker with default logging and yaml configuration.

```
./run_server
./run_worker
```

### DIFF
--- a/run_worker
+++ b/run_worker
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Run Redis docker container
+docker start buildfarm-redis || docker run -d --rm --name buildfarm-redis -p 6379:6379 redis:5.0.9
+redis-cli config set stop-writes-on-bgsave-error no
+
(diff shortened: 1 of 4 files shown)


run_worker
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Run Redis docker container
+docker start buildfarm-redis || docker run -d --rm --name buildfarm-redis -p 6379:6379 redis:5.0.9
```
jacobmou, line 6: we have the same started redis docker container command in `run_server`, would this work if we have already run `run_server`? I'm under the impression this would fail if we have started the container with the same name. 

We might want to put this command in a separate file.

80degreeswest: Yeah I thought about it but then it's a single line so I decided to just keep it as is. These scripts are fairly primitive at this point and will most likely grow in the future so we can start pulling some common functionality out as they get more advanced.

PR #1208, opened 2022-10-26
avoid crash with paths starting with double separators

adrianimboden:
This is the error I get on the worker node:
```
Exception in thread "Thread-190" java.lang.IllegalArgumentException: double separator in output directory
        at build.buildfarm.worker.OutputDirectory.parseDirectories(OutputDirectory.java:181)
        at build.buildfarm.worker.OutputDirectory.parse(OutputDirectory.java:109)
        at build.buildfarm.worker.shard.CFCExecFileSystem.createExecDir(CFCExecFileSystem.java:354)
        at build.buildfarm.worker.shard.ShardWorkerContext.createExecDir(ShardWorkerContext.java:713)
        at build.buildfarm.worker.InputFetcher.fetchPolled(InputFetcher.java:191)
        at build.buildfarm.worker.InputFetcher.runInterruptibly(InputFetcher.java:101)
        at build.buildfarm.worker.InputFetcher.run(InputFetcher.java:280)
        at java.base/java.lang.Thread.run(Thread.java:829)
```

To be honest, I don't know why it happens in the first place, but this fixes my problem.

The paths somehow get like this: `//tmp/bazeltest/_tmp/e062d84144817ba3a0e56f6760a7f161/` when e.g. using `bazel build --test_tmpdir=/tmp/bazeltest/`.

It also happens for normal filepaths, but only under certain circumstances (build of tools for genrules for example).

The bazel client that produces this problem is built from this commit: https://github.com/bazelbuild/bazel/commit/4779553eb5814025a465ebadbbced5c158996660 (I built it myself from main).

My remote build runs smoothly when I apply this change

### DIFF
--- a/src/main/java/build/buildfarm/worker/OutputDirectory.java
+++ b/src/main/java/build/buildfarm/worker/OutputDirectory.java
@@ -165,6 +165,9 @@ public class OutputDirectory {
     String prefix = "/";
     for (OutputDirectoryEntry entry : sortedOutputDirs) {
       String outputDir = entry.outputDirectory;
+      if (outputDir.startsWith("//")) {
+        outputDir = outputDir.substring(1);
+      }
       while (!outputDir.startsWith(prefix)) {
         currentBuilder = stack.pop();
         int upPathSeparatorIndex = prefix.lastIndexOf('/', prefix.length() - 2);


luxe:
That is strange.  Maybe we aren't doing path joining correctly somewhere.  We could land this, but I'm also thinking we should probably fix the path so that when it gets to this point its always correct.

werkt:
Seconded on this being the wrong location to address this - the execute validation should be handling this server side, and the worker should be validated to trigger a failed precondition execution if it reaches it. Filed #1272 to track this separately, recommending that we don't land this.

chill389cc:
Sorry to bump this, but we've also started running into this issue when trying to use the `--test_tmpdir` flag. Is there a mitigation or recommended ways of working around this? Thanks in advance.

werkt:
Can someone spell out the use case for this to be something functional remotely?

The option only affects test runs. Test runs have to be specifically invoked. What are you hoping to get, remotely, where the point is to disregard this option? The behavior of adding TEST_TMPDIR to the output directories list for creation can be ignored, but I want someone to describe why they want this to be effective for builds, and to understand that this is a risk for pollution on the workers outside of the execution root, and that the directories are not guaranteed to be created by the time execution time is reached.

werkt:
@chill389cc did you want to spell out a use case in a separate issue?

chill389cc:
For now we're getting around this by simply only using the flag with tests that we do not run remotely, as the flag was only necessary for us when running without remote-build-execution. Before, were trying to use the flag on a bazel test command that ran both local tests and some tests that used dynamic execution, but now we're splitting that into two separate commands, one for remote and one for local. We'll keep the flag on the local test and drop it for the remote test.

PR #1209, opened 2022-10-28
Provide digests for images and http files to use cache effectively

bshashank:
No digests for the containers especially trigger a ping to docker.io which can be slow. Also avoids the debug messages like:
```
  DEBUG: Rule 'amazon_corretto_java_image_base' indicated that a canonical reproducible form can be obtained by modifying arguments digest = "sha256:81d0df4412140416b27211c999e1f3c4565ae89a5cd92889475d20af422ba507"
```

on first run.

### DIFF
--- a/deps.bzl
+++ b/deps.bzl
@@ -153,5 +153,6 @@ def buildfarm_dependencies(repository_name = "build_buildfarm"):
 
     http_file(
         name = "tini",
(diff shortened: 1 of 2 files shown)


jacobmou (approved):
Looks good! Thanks!

PR #1210, opened 2022-11-02
Consolidate maxEntrySizeBytes config to avoid duplicaton

80degreeswest:
This change consolidates maxEntrySizeBytes config to a single location.

Previous config:
```
server:
  maxEntrySizeBytes: 2147483648 # 2 * 1024 * 1024 * 1024
worker:
  cas:
    maxEntrySizeBytes: 2147483648 # 2 * 1024 * 1024 * 1024
```

New config:
```
maxEntrySizeBytes: 2147483648 # 2 * 1024 * 1024 * 1024
server:
worker:
```

### DIFF
--- a/src/main/java/build/buildfarm/common/config/Cas.java
+++ b/src/main/java/build/buildfarm/common/config/Cas.java
@@ -17,7 +17,6 @@ public class Cas {
   private TYPE type = TYPE.FILESYSTEM;
   private String path = "cache";
   private long maxSizeBytes = 2147483648L; // 2 * 1024 * 1024 * 1024
-  private long maxEntrySizeBytes = 2147483648L; // 2 * 1024 * 1024 * 1024
   private boolean fileDirectoriesIndexInMemory = false;
   private boolean skipLoad = false;
   private String target;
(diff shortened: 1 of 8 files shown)


src/main/java/build/buildfarm/common/config/Cas.java
```diff
@@ -19,8 +19,7 @@ public enum TYPE {
 
   private TYPE type = TYPE.FILESYSTEM;
   private String path = "cache";
-  private long maxSizeBytes = 0;
-  private long maxEntrySizeBytes = 2147483648L; // 2 * 1024 * 1024 * 1024
+  private long maxSizeBytes = 2147483648L; // 2 * 1024 * 1024 * 1024
```
luxe, line 22:
```suggestion
  private long maxSizeBytes = 0;
```

luxe (approved):

werkt:
I'd like to see things like this come in with a little less aggressive output: we should be validating configs for changes like these, and putting something out better than the parse error highlighted by a user who pulled this in early:

```
Unable to find property 'maxEntrySizeBytes' on class: build.buildfarm.common.config.Server
 in 'reader', line 20, column 22:
      maxEntrySizeBytes: 4294967296 # 4 * 1024 * 1024 * 1024
```

Error message, fail to start, something that doesn't look like a pure parsing error.

PR #1212, opened 2022-11-03
Automatically set CAS size based on available disk space

80degreeswest:
Automatically set CAS size based on available disk space.

Set preference:

1. User configured in yaml
2. Attempt to automatically set to 90% of available disk size on the worker CAS path
3. Default pre-configured value of 2GB

### DIFF
--- a/src/main/java/build/buildfarm/common/config/Cas.java
+++ b/src/main/java/build/buildfarm/common/config/Cas.java
@@ -1,12 +1,15 @@
 package build.buildfarm.common.config;
 
 import com.google.common.base.Strings;
+import java.io.File;
 import java.nio.file.Path;
 import javax.naming.ConfigurationException;
 import lombok.Data;
 
 @Data
 public class Cas {
+  private static final long DEFAULT_CAS_SIZE = 2147483648L; // 2 * 1024 * 1024 * 1024
+
   public enum TYPE {
     FILESYSTEM,
     GRPC,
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/common/config/Cas.java
```diff
@@ -16,12 +19,30 @@ public enum TYPE {
...
+  /*
+  Automatically set disk space to 90% of available space on the worker volume.
+  User configured value in .yaml will always take presedence.
+   */
+  public long getMaxSizeBytes() {
+    if (maxSizeBytes == 0) {
+      try {
+        maxSizeBytes =
+            (long)
+                (new File(BuildfarmConfigs.getInstance().getWorker().getRoot()).getFreeSpace()
```
jacobmou, line 37: Seems like most people recommend `java.io.File.getUsableSpace()` more than instead of `java.io.File.getFreeSpace()` in https://stackoverflow.com/questions/21439596/difference-between-getfreespace-and-getusablespace-of-file

src/main/java/build/buildfarm/common/config/Cas.java
```diff
@@ -16,12 +19,30 @@ public enum TYPE {
...
+                    * 0.9);
+      } catch (Exception e) {
+        maxSizeBytes = DEFAULT_CAS_SIZE;
+      }
+    }
+    return maxSizeBytes > 0 ? maxSizeBytes : DEFAULT_CAS_SIZE;
```
jacobmou, line 43: `return maxSizeBytes;` should be more than enough.

src/main/java/build/buildfarm/common/config/Cas.java
80degreeswest: Nice, good catch here!

src/main/java/build/buildfarm/common/config/Cas.java
80degreeswest: With getFreeSpace, I was getting 0 return on a specific non root path. With getUsableSpace it works fine!

PR #1219, opened 2022-11-14
Add optional GRPC metrics for worker

shirchen:
In order to understand slow responses from server, optionally instrument worker with GRPC metrics as well.
Ref #1205


examples/config.yml
```diff
@@ -81,6 +81,9 @@ backplane:
 worker:
   port: 8981
   publicName: "localhost:8981"
+  grpcMetrics:
```
80degreeswest, line 84: I think we should pull this config up to the top level so that it can control both server and worker together instead of duplicating it.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -226,10 +228,31 @@ private Server createServer(
           new WorkerProfileService(
               storage, inputFetchStage, executeActionStage, context, completeStage, backplane));
     }
+    handleGrpcMetricIntercepts(serverBuilder);
 
     return serverBuilder.build();
   }
 
+  public void handleGrpcMetricIntercepts(ServerBuilder<?> serverBuilder) {
```
80degreeswest, line 236: This could go into a common helper class to avoid duplication with the server.

80degreeswest (changes requested):

examples/config.yml
shirchen: Not everyone wants to instrument both: I am planning on not instrumenting workers w/histograms, but may change my mind, so I think we should allow knobs for server and worker to be separate. An example is that someone may want to use metrics for SLA enforcement only, and they wont want to take a performance hit for adding metrics to workers.

src/main/java/build/buildfarm/worker/shard/Worker.java
shirchen: 👍

examples/config.yml
80degreeswest: You'll still be able to configure them separately by providing a different config file to server/worker. The example file has both configurations but only for simplicity.

src/main/java/build/buildfarm/server/BuildFarmServer.java
```diff
@@ -121,7 +120,7 @@ public synchronized void start(ServerBuilder<?> serverBuilder, String publicName
         .addService(new PublishBuildEventService())
         .intercept(TransmitStatusRuntimeExceptionInterceptor.instance())
         .intercept(headersInterceptor);
-    handleGrpcMetricIntercepts(serverBuilder);
+    GrpcMetrics.handleGrpcMetricIntercepts(serverBuilder, configs.getServer().getGrpcMetrics());
```
80degreeswest, line 123: We already have a BuildfarmConfigs object which contains GrpcMetrics. You can simply do configs.getGrpcMetrics.handleGrpcMetriIntercepts here and make the method non-static.

PR #1223, opened 2022-11-29
Synchronize for WSO out modifications

werkt:
The out stream member may be nulled during expected non-null processing via onNext. Guard it properly with this.


shirchen:
Did a regression test and I think this fixes NPEs. I did the following:
- on staging, run with buildfarm 2.1.0 to reproduce the error in #1222 
- deployed this branch, and reran the same steps and did not observe any NPEs.

PR #1220, opened 2022-11-20
fix(sec): upgrade org.json:json to 20180130

ShaoNianL:
### What happened？
There are 1 security vulnerabilities found in org.json:json 20080701
- [MPS-2022-13520](https://www.oscs1024.com/hd/MPS-2022-13520)

### What did I do？
Upgrade org.json:json from 20080701 to 20180130 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

PR #1230, opened 2022-12-22
fix(sec): upgrade org.json:json to 20180130

luxe:
Copy of: https://github.com/bazelbuild/bazel-buildfarm/pull/1220
CLA was needed.


jacobmou (approved):
in case the we cannot land https://github.com/bazelbuild/bazel-buildfarm/pull/1220, let's land this.

PR #1235, opened 2022-12-27
Workaround for test timeout

bozaro:
Currently `timeout` and `pollDeadline` points to the same time. Almost always, when the test timeout expires `pollDeadline` triggers before `timeout` and bazel try retry action instead of report test `TIMEOUT` failure.

For issue reproduce I used rule:
```
sh_test(
    name = "timeout_test",
    size = "small",
    timeout = "short",
    srcs = ["test.sh"],
)
```
with `test.sh` script:
```
#!/bin/sh
echo "$(date -Is): BEGIN"
sleep 90
echo "$(date -Is): END"
exit 1 
```


luxe (approved):

luxe:
We should make the poll offset a part of `TimeoutSettings`

PR #1241, opened 2023-01-13
Use non-blocking UUID generation in CASFileCache

zpzjzj:
https://github.com/apache/openwhisk/issues/2747


werkt:
Can you provide some context on the reason for this? What is the benefit to applying a non blocking uuid gen?

werkt:
> Can you provide some context on the reason for this? What is the benefit to applying a non blocking uuid gen?

Do you really have enough write generation that this matters (i.e. dominates)?

luxe:
Good job profiling this.  It's nice to see you got a performance boost as well.  My only suggestion is to make the UUID fetching something like a [Supplier](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html).  That way, we can use dependency injection and choose which UUID implementation we want.

werkt:
@zpzjzj can you rebase and revalidate tests with this?

PR #1243, opened 2023-01-15
[alternative backplane] Create interfaces for distributed maps & hashmaps

luxe:
I'd like to use an alternative backend instead of redis. To make that possible, we need all of the distributed state to be redis agnostic.  I've converted the `RedisMap` and `RedisHashMap` to interface types-- this will allow swapping in different data structure implementations.

See this discussion around the effort - https://github.com/bazelbuild/bazel-buildfarm/issues/665
There is more to do, like isolating the Redis pub/sub from the backplane, ensuring the queues are behind an interface, etc.


src/main/java/build/buildfarm/common/distributed/DistributedMap.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @brief Set key to hold the string value and set key to timeout after a given number of seconds.
+   * @details If the key already exists, then the value is replaced.
+   * @param client A client used by the implementation.
+   * @param key The name of the key.
+   * @param value The value for the key.
+   * @param timeout_s Timeout to expire the entry. (units: seconds (s))
+   * @note Overloaded.
+   */
+  void insert(DistributedClient client, String key, String value, long timeout_s);
```
werkt, line 48: why is this method necessary? you anticipate exceeding a billenium for an insert timeout?

src/main/java/build/buildfarm/common/distributed/DistributedHashMap.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Set;
+
+/**
+ * @class DistributedHashMap
+ * @brief A distributed hashmap.
+ * @details A distributed hashmap is an implementation of a hashmap data structure which can be
+ *     shared across a distributed system. For example, implementations of this interface might
+ *     choose to use redis, hazelcast, postgres, etc.
+ */
+public interface DistributedHashMap<DistributedClient> {
```
werkt, line 27: shouldn't this extend the DistributedMap interface? Also the Hash seems like an implementation detail that doesn't mandate a separate interface.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @brief Set key to hold the string value and set key to timeout after a given number of seconds.
+   * @details If the key already exists, then the value is replaced.
+   * @param client A client used by the implementation.
+   * @param key The name of the key.
+   * @param value The value for the key.
+   * @param timeout_s Timeout to expire the entry. (units: seconds (s))
+   * @note Overloaded.
+   */
+  void insert(DistributedClient client, String key, String value, int timeout_s);
```
werkt, line 37: If you're going to timeout, this carries no facility that isn't a RuntimeException for delivering that timeout - for the timeout being part of the interface, it should throw a checked exception.

werkt (changes requested):
It looks like you're trying to create a capture interface with the least involvement in changes on the Redis side - I'd prefer fixes to the usage to make a smaller/simpler interface spec than just the min spec.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
luxe: This parameter is poorly named.  It's the expire time when calling `setex`.
`Timeout to expire the entry. (units: seconds (s))`.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
luxe: It looks like clients were often calling with longs and this was done to avoid updating all the callers.  I agree we should remove this and tighten the interface.

luxe:
> It looks like you're trying to create a capture interface with the least involvement in changes on the Redis side - I'd prefer fixes to the usage to make a smaller/simpler interface spec than just the min spec.

haha that's exactly what I was trying to do.  Okay, let me follow up with smaller interfaces.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
luxe: this API has been removed.

src/main/java/build/buildfarm/common/distributed/DistributedHashMap.java
luxe: That was a good idea.  There was more overlap than I thought.  Now there is only `DistributedMap` and we can create `RedisMap`, `RedisHashMap` from it.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
luxe: renamed to `expiration_s`

luxe:
> > It looks like you're trying to create a capture interface with the least involvement in changes on the Redis side - I'd prefer fixes to the usage to make a smaller/simpler interface spec than just the min spec.
> 
> haha that's exactly what I was trying to do. Okay, let me follow up with smaller interfaces.

I've combined the interfaces and made everything just a `DistributedMap`.

luxe:
@werkt ready for re-review

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Set;
+
+/**
+ * @class DistributedMap
+ * @brief A distributed map.
+ * @details A distributed map is an implementation of a map data structure which can be shared
+ *     across a distributed system. For example, implementations of this interface might choose to
+ *     use redis, hazelcast, postgres, etc.
+ */
+public interface DistributedMap<DistributedClient> {
```
werkt, line 27: This is a Map with an extra parameter added to every method. It only makes sense if you're trying to refer to a different client in the same instance Map, which I rather doubt, since you're hiding implementations for it in the concretes. The only different method from Map is the timeout insert, which still doesn't have a checked path, and the details for which can be hidden in a concrete class implementation, unless again, you want to have a different timeout for different circumstances.

In my blatantly biased opinion, you don't need this.

src/main/java/build/buildfarm/common/redis/RedisHashMap.java
```diff
@@ -45,6 +51,22 @@ public RedisHashMap(String name) {
...
+   * @param key The name of the key.
+   * @param value The value for the key.
+   * @param expiration_s Timeout to expire the entry. (units: seconds (s))
+   * @return Whether a new key was inserted. If a key is overwritten with a new value, this would be
+   *     false.
+   * @note Overloaded.
+   */
+  @Override
+  public boolean insert(JedisCluster jedis, String key, String value, int expiration_s) {
+    return jedis.hset(name, key, value) == 1;
```
werkt, line 67: And the unique timeout insert has its expiration ignored. Call the timeout ignorant insert here.

src/main/java/build/buildfarm/common/redis/RedisHashMap.java
```diff
@@ -27,7 +33,7 @@
  *     before and after the map data structure is created (since it exists in redis). Therefore, two
  *     redis maps with the same name, would in fact be the same underlying redis map.
  */
-public class RedisHashMap {
+public class RedisHashMap implements DistributedMap<JedisCluster> {
```
werkt, line 36: `implements Map<String, String>`

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
```diff
@@ -0,0 +1,139 @@
...
+   */
+  Set<String> keys(DistributedClient client);
+
+  /**
+   * @brief Convert the backend map to a java map.
+   * @details This would not be efficient if the map is large. * @param client A client used by the
+   *     implementation.
+   * @return The backend map represented as a java map.
+   */
+  Map<String, String> asMap(DistributedClient client);
```
werkt, line 138: Obviated when you present Map<String, String> in implementations

werkt (changes requested):
Sorry for the back and forth. You've done an excellent job of refactoring this to be a much clearer Map<String, String> implementation where you don't need to gild the method lilly with the client.

src/main/java/build/buildfarm/common/distributed/DistributedMap.java
luxe: Are you suggesting we hold a copy of `JedisCluster` inside the implementation of `RedisHashMap` instead?  That seems like the only way I can inherit from `Map<String, String>` and remove it as an input parameter from all the methods.

werkt: Precisely. Make this an inherited implementation of Map<String, String>

werkt:
@luxe did you want to follow up with an updated version of this with Map<...> inheritance, or should this be closed?

PR #1244, opened 2023-01-16
[REAPI] add support for output_paths (REAPI 2.1)

luxe:
Addresses part of https://github.com/bazelbuild/bazel-buildfarm/issues/376, by sourcing from `Command::output_paths`.  Provides  backwards compatibility to `output_files` and `output_directories` when the attribute is empty.  [Implementation](https://github.com/bazelbuild/bazel-buildfarm/pull/1244/files#diff-101096a75b4ec7962db18bbb0038a2ad0f3927f5bcfdc36ad8a4d0bb120ce973R49) points to relevant REAPI details.

One of the interesting things we need to consider is that buildfarm may no longer know if an output is a file or directory until it executes the action and looks at the filesystem.  Any upfront assumptions about file type based on the `Command` cannot be relied on.  This is an assumption I made when extracting files with the [DockerExecutor](https://github.com/bazelbuild/bazel-buildfarm/blob/54525e71ec73e142cf76fbfed6efb3386e561514/src/main/java/build/buildfarm/worker/DockerExecutor.java#L235-L242).


src/main/java/build/buildfarm/common/CommandUtils.java
```diff
@@ -34,4 +37,50 @@ public static boolean isTest(Command command) {
...
+    // https://github.com/bazelbuild/remote-apis/issues/75
+    // That's unfortunate for server implementations, because now buildfarm has to distinguish
+    // between files and paths itself.
+
+    // To quote REAPI:
+    // "New in v2.1: this field supersedes the DEPRECATED `output_files` and
+    // `output_directories` fields. If `output_paths` is used, `output_files` and
+    // `output_directories` will be ignored!"
+    // https://github.com/bazelbuild/remote-apis/blob/3a21deee813d0b98aaeef9737c720e509e10dc8b/build/bazel/remote/execution/v2/remote_execution.proto#L651-L653
+    List<Path> resolvedPaths = new ArrayList<>();
```
werkt, line 63: Small nit here: if getOutputPathsCount() != 0, just return the getOutputPathsList().

werkt (approved):
Seems like we're covering the API well there.

src/main/java/build/buildfarm/common/CommandUtils.java
luxe: fixed

PR #1250, opened 2023-01-24
[metrics] capture CAS TTL on expired blobs

luxe:
We don't know how long our CAS blobs live across workers.  
This add a metric for how long CAS entries exist before getting expiring / deleted.


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -314,6 +320,14 @@ public CASFileCache(
     this.delegateSkipLoad = delegateSkipLoad;
     this.directoriesIndexDbName = directoriesIndexDbName;
 
+    if (configs.getWorker().getStorages().get(0).isPublishTtlMetric()) {
+      casTtl =
+          Histogram.build()
+              .name("cas_ttl")
+              .help("The amount of time CAS entries live in L1 storage")
```
shirchen, line 327: Can we clarify the units of this metric here?

src/main/java/build/buildfarm/common/config/Cas.java
```diff
@@ -29,6 +29,9 @@ public enum TYPE {
   // GRPC
   private String target;
 
+  // Metrics
+  private boolean publishTtlMetric = false;
```
shirchen, line 33: shall we add this option to the example?

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: Fixed.  I put the units in the metric name as well (`cas_ttl_ms`).  I think a lot of our time metrics are missing units-- so that's something we might want to fix also.

src/main/java/build/buildfarm/common/config/Cas.java
luxe: added

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -314,6 +320,15 @@ public CASFileCache(
     this.delegateSkipLoad = delegateSkipLoad;
     this.directoriesIndexDbName = directoriesIndexDbName;
 
+    if (configs.getWorker().getStorages().get(0).isPublishTtlMetric()) {
```
werkt, line 323: We really should be taking a config object as a param if you're going to use it this way. We don't know that we're the 0-indexed config for that tree, and this makes things lots less testable (without a config instance setup in every test class).

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2535,6 +2550,21 @@ private CancellableOutputStream putOrReference(
     }
   }
 
+  private void deleteExpiredKey(String expiredKey) throws IOException {
+    Path path = getPath(expiredKey);
+    if (configs.getWorker().getStorages().get(0).isPublishTtlMetric()) {
+      publishExpirationMetric(path);
+    }
+    Files.delete(path);
```
werkt, line 2558: Reverse these. I don't want to delay the file deletion by your publish, and at the time you publish here, we haven't deleted yet...

werkt (changes requested):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: fixed to not use 0 indexing.  Now the CAS gets appropriate settings on construction

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -325,6 +333,15 @@ public CASFileCache(
     this.delegateSkipLoad = delegateSkipLoad;
     this.directoriesIndexDbName = directoriesIndexDbName;
 
+    if (publishTtlMetric) {
+      casTtl =
+          Histogram.build()
```
shirchen, line 338: Do default buckets work for this case? I think we had to make some adjustments for some of the metrics to make them useful.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2585,6 +2602,26 @@ private CancellableOutputStream putOrReference(
...
+    // We don't want the metric publishing to delay the deletion of the file.
+    // We publish the metric only after the file has been deleted.
+    if (publishTtlMetric) {
+      long createdTime = path.toFile().lastModified();
+      Files.delete(path);
+      publishExpirationMetric(createdTime);
+    } else {
+      Files.delete(path);
+    }
+  }
```
werkt, line 2617: Would prefer two blocks checking publishTtlMetric for the population and then publish to two copies of the delete call here.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2585,6 +2602,26 @@ private CancellableOutputStream putOrReference(
     }
   }
 
+  private void deleteExpiredKey(String expiredKey) throws IOException {
```
werkt, line 2605: Just take the path here - you don't concern yourself with the key except to getPath

src/main/java/build/buildfarm/common/config/Backplane.java
```diff
@@ -45,7 +45,6 @@ public enum BACKPLANE_TYPE {
   private boolean cacheCas = false;
 
   public String getRedisUri() {
-
```
werkt, line 48: random whitespace change

src/test/java/build/buildfarm/cas/cfc/CASFileCacheTest.java
```diff
@@ -1108,6 +1110,8 @@ public void newInputThrowsNoSuchFileExceptionWithoutDelegate() throws Exception
             /* maxEntrySizeInBytes=*/ 1024,
             /* hexBucketLevels=*/ 1,
             storeFileDirsIndexInMemory,
+            false,
+            false,
```
werkt, line 1114: provide the /* paramName=*/ helpers here and above

werkt (approved):
Approved pending comments

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: Oh good point.  I'm currently graphing it, and grafana's histogram made it look like we weren't keep cas for very long.  Changing the metric to seconds now and adding bucket sizes 1hr through 2 weeks.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: I think is correct now.  I still have to get the modifed time, so I keep that outside the two blocks.

src/main/java/build/buildfarm/common/config/Backplane.java
luxe: fixed

src/test/java/build/buildfarm/cas/cfc/CASFileCacheTest.java
luxe: added

PR #1255, opened 2023-01-27
[config] make execution wrappers configurable

luxe:
### Context
Buildfarm supports different kinds of execution wrappers based on `exec_properties` and global configuration.  It's always been an optional dependency not fully baked into the images we deploy.  Nonetheless, some of them are very important, like the `linux-sandbox`, which is now my preference for supporting additional behavior.  

In order to make this more accessible to more deployments, let's allow the customization of these paths. 

 ### Compatibility 
Defaults are the same.  Configs are forwards compatible.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -118,3 +118,12 @@ worker:
       executionWrapper:
         path: "/"
         arguments:
+executionWrappers:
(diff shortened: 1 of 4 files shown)


werkt (approved):
Real nice.

PR #1261, opened 2023-02-10
[config] derive public name in config files

luxe:
### Context
Buildfarm configs require that a `publicName` be given to the server and worker.  These names are used as a URI for the nodes to communicate. Since buildfarm is often deployed as many servers and workers, there is a complexity to providing these URIs at the configuration level.  We'd like to provide a derivation in the event that the configuration value is not given.

#### Config value selection
   - allow `pubicName` for server/worker to derive as hostname
   - fixed order of other env variables (`REDIS_URI` and `EXECUTION_SLOT_WIDTH`)

   #### Fix Tilt deployment (For k8s testing)
   - separated out open telemetry jvm flags
   - made open telemetry optional for default containers (spews some errors to logs when enabled)
   - made Tilt refresh optional based on bazel query optional (too slow)
   
   ### Testing
   Tested this with repo k8s files, and was able to perform remote builds.
   
   ### Speculation on k8s compatibility:
   
When looking at the pod I see:
```
kubectl get pods -o wide | grep worker | tr -s " "
shard-worker-56b6cd6d6d-2g96n 1/1 Running 0 73m 10.1.110.85 desktop <none> <none>
```
Which is indeed a different IP than what I see for the service:
```
kubectl get service --all-namespaces | grep worker | tr -s " "
default shard-worker-service ClusterIP 10.152.183.219 <none> 8981/TCP,9091/TCP 74m
```
However, when looking at the endpoint, I see the same pod IP:
```
kubectl get endpoints | grep worker | tr -s " "
shard-worker-service 10.1.110.85:8981,10.1.110.85:9091 74m
```

All this is to say, leaving out publicName resulted in 10.1.110.85:8981 being derived which did work in my local k8s deployment and allow me to perform remote bazel builds. I don't know enough about k8s to claim that this will work for all other deployments (such as yours). I'll have more experience on that when I do a production level k8s deployment and try to leverage this.

I can imagine a future adjustment where we detect that were inside a k8s pod, and query for the correct publicly visible endpoint or something. For now, getLocalHost() seems fairly robust.

### DIFF
--- a/src/main/java/build/buildfarm/common/config/Server.java
+++ b/src/main/java/build/buildfarm/common/config/Server.java
@@ -53,7 +53,7 @@ public class Server {
 
     // derive a value
     try {
-      return InetAddress.getLocalHost().getHostAddress();
+      return InetAddress.getLocalHost().getHostAddress() + ":" + port;
     } catch (Exception e) {
       log.severe("publicName could not be derived:" + e);
       return publicName;
--- a/src/main/java/build/buildfarm/common/config/Server.java
+++ b/src/main/java/build/buildfarm/common/config/Server.java
@@ -1,10 +1,13 @@
 package build.buildfarm.common.config;
 
 import com.google.common.base.Strings;
+import java.net.InetAddress;
 import java.util.UUID;
 import lombok.Data;
+import lombok.extern.java.Log;
 
 @Data
+@Log
 public class Server {
   public enum INSTANCE_TYPE {
     SHARD
@@ -38,10 +41,22 @@ public class Server {
   private String publicName;
 
   public String getPublicName() {
+    // use environment override (useful for containerized deployment)
+    if (!Strings.isNullOrEmpty(System.getenv("INSTANCE_NAME"))) {
+      return System.getenv("INSTANCE_NAME");
+    }
+
+    // use configured value
     if (!Strings.isNullOrEmpty(publicName)) {
       return publicName;
-    } else {
-      return System.getenv("INSTANCE_NAME");
+    }
+
+    // derive a value
+    try {
+      return InetAddress.getLocalHost().getHostAddress();
+    } catch (Exception e) {
+      log.severe("publicName could not be derived:" + e);
+      return publicName;
     }
   }
 
--- a/BUILD
+++ b/BUILD
@@ -115,6 +115,26 @@ sh_binary(
     srcs = ["macos-wrapper.sh"],
 )
 
+SERVER_TELEMETRY_JVM_FLAGS = [
+    "-javaagent:/app/build_buildfarm/opentelemetry-javaagent.jar",
+    "-Dotel.resource.attributes=service.name=server",
+    "-Dotel.exporter.otlp.traces.endpoint=http://otel-collector:4317",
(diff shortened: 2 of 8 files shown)


src/main/java/build/buildfarm/common/config/Server.java
```diff
@@ -38,10 +41,22 @@ public enum INSTANCE_TYPE {
   private String publicName;
 
   public String getPublicName() {
+    // use environment override (useful for containerized deployment)
+    if (!Strings.isNullOrEmpty(System.getenv("INSTANCE_NAME"))) {
```
80degreeswest, line 45: Can you double check other usages of env var override to make sure we have the same order of precedence with them? I thought other configs use yaml first and env var second. This makes sense to override but just want to make sure we are consistent.

src/main/java/build/buildfarm/common/config/Server.java
```diff
@@ -38,10 +41,22 @@ public enum INSTANCE_TYPE {
...
+    // use configured value
     if (!Strings.isNullOrEmpty(publicName)) {
       return publicName;
-    } else {
-      return System.getenv("INSTANCE_NAME");
+    }
+
+    // derive a value
+    try {
+      return InetAddress.getLocalHost().getHostAddress() + ":" + port;
```
80degreeswest, line 56: Have you tested this with k8s? May throw a comment in here that it needs to be tested, if it hasn't as I think the pod IP could be different from the container IP.

80degreeswest (approved):

src/main/java/build/buildfarm/common/config/Server.java
luxe: Yeah, the precedence was not the same. It may have broken when we migrated to yaml-- I noticed the `REDIS_URI` env override did not work for the Tilt deployment as it had before.  These are fixed and consistent now. thanks!

src/main/java/build/buildfarm/common/config/Server.java
luxe: good suggestion.  I've now tried this with the local repo k8s files, and to my surprise, it worked.

When looking at the pod I see:
```
kubectl get pods -o wide | grep worker | tr -s " "
shard-worker-56b6cd6d6d-2g96n 1/1 Running 0 73m 10.1.110.85 desktop <none> <none>
```

Which is indeed a different IP than what I see for the service:
```
kubectl get service --all-namespaces | grep worker | tr -s " "
default shard-worker-service ClusterIP 10.152.183.219 <none> 8981/TCP,9091/TCP 74m
```

However, when looking at the endpoint, I see the same pod IP:
```
kubectl get endpoints | grep worker | tr -s " "
shard-worker-service 10.1.110.85:8981,10.1.110.85:9091 74m
```

All this is to say, leaving out `publicName` resulted in `10.1.110.85:8981` being derived which did work in my local k8s deployment and allow me to perform remote bazel builds.  I don't know enough about k8s to claim that this will work for all other deployments (such as yours).  I'll have more experience on that when I do a production level k8s deployment and try to leverage this.

I can imagine a future adjustment where we detect that were inside a k8s pod, and query for the correct publicly visible endpoint or something.  For now, `getLocalHost()` seems fairly robust.

PR #1262, opened 2023-02-10
[config] allow loading config via CONFIG_PATH

luxe:
For some deployments it is easier to provide the config file via an environment variable instead of a CLI argument.


80degreeswest (approved):
Nice. Would also be cool to go one step further and add some default locations we could check for existence of a config file and if one (and only one) is found in one of those locations we can use it (add the location to the configs printout when the app starts).

PR #1263, opened 2023-02-13
Fix use of wrong image in example

Taldoras:
It seems a copy/paste error has been made when the official bazelbuild images were introduced (in https://github.com/bazelbuild/bazel-buildfarm/pull/1246).


luxe (approved):
thank you!

PR #1264, opened 2023-02-15
[refactor] isolate action fetching during operation queue to a separate function

luxe:
- Isolate action fetching during operation queue to a separate function
 - Use `FluentFuture`
 
 Behavior should be the same.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/ShardInstance.java
+++ b/src/main/java/build/buildfarm/instance/shard/ShardInstance.java
@@ -2115,7 +2115,6 @@ public class ShardInstance extends AbstractServerInstance {
             "ShardInstance(%s): queue(%s): fetching action %s",
             getName(), operation.getName(), actionDigest.getHash()));
 
-    // Fetch an action from the CAS and invalidate it from the AC if requested.
     return FluentFuture.from(expectAction(actionDigest, requestMetadata))
         .transformAsync(
             action -> transformActionViaActionCache(action, metadata, operation),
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -2104,6 +2105,58 @@ public ListenableFuture<Void> queue(ExecuteEntry executeEntry, Poller poller, Du
...
+      PreconditionFailure.Builder preconditionFailure = PreconditionFailure.newBuilder();
+      preconditionFailure
+          .addViolationsBuilder()
+          .setType(VIOLATION_TYPE_MISSING)
+          .setSubject("blobs/" + DigestUtil.toString(actionDigest))
+          .setDescription(MISSING_ACTION);
+      checkPreconditionFailure(actionDigest, preconditionFailure.build());
+    }
+    throw st.asRuntimeException();
+  }
```
werkt, line 2158: This framing of the exception filtering in a method makes this pattern look pretty awkward. Is there a clearer/cleaner way to express what we're dealing with here? The if block here is a noescape/noreturn, and we should not be throwing asRuntimeException from a Status when we declare to throw StatusException (asException() preferred).

werkt (commented):
What is your opinion of the coverage for what you've just stretched this out into? I'm concerned that without tests added in this change that we will risk substantial critical regressions if anything was missed in this conversion.

PR #1265, opened 2023-02-15
[refactor] remove recursion from enumerateActionInputDirectory

luxe:
`enumerateActionInputDirectory` is currently recursive.  Use a `Stack` to avoid stack overflow.


src/main/java/build/buildfarm/common/config/Backplane.java
```diff
@@ -45,7 +45,6 @@ public enum BACKPLANE_TYPE {
   private boolean cacheCas = false;
 
   public String getRedisUri() {
-
```
werkt, line 48: errant ws change, pull this out

werkt (approved):
Nice, this should be safer. Good Job

PR #1266, opened 2023-02-17
[config] allow the sandbox to be chosen by default

luxe:
Using the `linux-sandbox` has been primarily a decision of the client chosen via exec_properties.  For some configurations, we feel that it is best to enforce the sandbox on the service side and prevent any client from stepping outside the sandbox.  This preserves existing behavior but adds configs to make the sandbox selected by default.


werkt (changes requested):
Given all the additions of SandboxSettings() here to the tests, can you add some test validation that they're behaving the way you expect in loop?

luxe:
> Given all the additions of SandboxSettings() here to the tests, can you add some test validation that they're behaving the way you expect in loop?

unit tests added!

_site/docs/configuration/configuration.md
```diff
@@ -283,6 +283,24 @@ worker:
     execution: true
 ```
 
+### Sandbox Settings
+
+| Configuration | Accepted and _Default_ Values | Description                                     |
+|---------------|-------------------------------|-------------------------------------------------|
+| alwaysUse             | boolean, _false_      | Enforce that the sandbox be used on every acion.|
+| selectForBlockNetwork | boolean, _false_      | Only use the sandbox when blocking the network. |
+| selectForTmpFs        | boolean, _false_      | Only use the sandbox when using tmpfs.          |
```
werkt, line 292: The Only language here confuses me: I'd change this to something like "block-network implies actions run under sandbox", if that's the right tag.

src/test/java/build/buildfarm/worker/resources/ResourceDeciderTest.java
```diff
@@ -666,9 +687,128 @@ public void decideResourceLimitationsTestWorkerName() throws Exception {
...
+            "foo",
+            /* defaultMaxCores=*/ 0,
+            /* onlyMulticoreTests=*/ false,
+            /* limitGlobalExecution=*/ false,
+            /* executeStageWidth=*/ 100,
+            /* allowBringYourOwnContainer=*/ false,
+            sandboxSettings);
+
+    // ASSERT
+    assertThat(limits.useLinuxSandbox).isTrue();
```
werkt, line 778: I want to see a negative test for this as well, where useLinuxSandbox is false for selectForBlockNetwork is true for non-network blocking platform

werkt (approved):
Approve pending comment mitigation

PR #1268, opened 2023-02-19
Show client error context when failing to establish exec root.

luxe:
### Context:
Buildfarm's CAS implementation relies heavily on being able to hardlink files.  There is a known limitation for particular filesystems like ext4 in which the hardlinking limit of any individual file can not exceed 65k.  Here is how you can reproduce that limitation by creating multiple targets that reference the same file in their input root:
```
# duplicate.bzl
def sh_test_duplicate(name,srcs,amount):
    for index in range(amount):
        native.sh_test(
            name = name + str(index),
            srcs = srcs,
        )

# BUILD
sh_test_duplicate(
    name = "main",
    srcs = ["main.sh"],
    amount = 65537, # Expose the ext4 hardlinking limit: (2^16)+1
)

# client
bazelisk test --remote_executor=grpc://localhost:8980 //code/programs/example_tests/hardlinking/...
```

### Problem:
When this limitation is reached, the client does not get an appropriate failure message.  The `InputFetchStage` will immediately expire the operation causing the `DispatchMonitor` to requeue it. Eventually the client gets a requeue fail message that looks like this:

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -2042,6 +2042,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
           log.log(Level.FINE, format("directory %s has been unlocked", path.getFileName()));
         },
         service);
+
     return putFuture;
   }
 
@@ -2089,7 +2090,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     private final List<Throwable> exceptions;
 
     PutDirectoryException(Path path, Digest digest, List<Throwable> exceptions) {
-      super(String.format("%s: %d exceptions", path, exceptions.size()));
+      // When printing the exception, show the captured sub-exceptions.
+      super(String.format("%s: %d exceptions: %s", path, exceptions.size(), exceptions));
       this.path = path;
       this.digest = digest;
       this.exceptions = exceptions;
--- a/src/main/java/build/buildfarm/worker/InputFetcher.java
+++ b/src/main/java/build/buildfarm/worker/InputFetcher.java
@@ -27,10 +27,12 @@ import build.bazel.remote.execution.v2.Directory;
 import build.bazel.remote.execution.v2.DirectoryNode;
 import build.bazel.remote.execution.v2.ExecutedActionMetadata;
 import build.bazel.remote.execution.v2.FileNode;
+import build.buildfarm.common.FailedOperationGetter;
 import build.buildfarm.common.ProxyDirectoriesIndex;
 import build.buildfarm.v1test.QueuedOperation;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Iterables;
+import com.google.longrunning.Operation;
 import com.google.protobuf.Duration;
 import com.google.protobuf.util.Durations;
 import com.google.protobuf.util.Timestamps;
@@ -194,7 +196,7 @@ public class InputFetcher implements Runnable {
               queuedOperation.getCommand());
     } catch (IOException e) {
       log.log(Level.SEVERE, format("error creating exec dir for %s", operationName), e);
-      owner.error().put(operationContext);
+      failOperation("Error creating exec dir", e.toString());
       return 0;
     }
     success = true;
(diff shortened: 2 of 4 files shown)


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2042,6 +2042,7 @@ public ListenableFuture<Path> putDirectory(
           log.log(Level.FINE, format("directory %s has been unlocked", path.getFileName()));
         },
         service);
+
```
werkt, line 2045: nit whitespace

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2089,7 +2090,8 @@ static class PutDirectoryException extends IOException {
     private final List<Throwable> exceptions;
 
     PutDirectoryException(Path path, Digest digest, List<Throwable> exceptions) {
-      super(String.format("%s: %d exceptions", path, exceptions.size()));
+      // When printing the exception, show the captured sub-exceptions.
+      super(String.format("%s: %d exceptions: %s", path, exceptions.size(), exceptions));
```
werkt, line 2094: While you're at it, can you fix up the plural here? ChoiceFormat almost looks appropriate for this, if overkill

src/main/java/build/buildfarm/worker/InputFetcher.java
```diff
@@ -194,7 +196,7 @@ private long fetchPolled(Stopwatch stopwatch) throws InterruptedException {
               queuedOperation.getCommand());
     } catch (IOException e) {
       log.log(Level.SEVERE, format("error creating exec dir for %s", operationName), e);
-      owner.error().put(operationContext);
+      failOperation("Error creating exec dir", e.toString());
```
werkt, line 199: Without the delivery to the error() path, there may be existing behavior that we're losing here. The call to failOperation should decorate the context and notify the error() path, where it should perform the putOperation below in failOperation.

werkt (changes requested):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: Using a simple ternary for singular/plural now.

src/main/java/build/buildfarm/worker/InputFetcher.java
luxe: failedOperation updates the context and forwards it to `owner.error().put(operationContext);` as before.

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
```diff
@@ -312,6 +313,12 @@ List<Throwable> getExceptions() {
     }
   }
 
+  private static String getErrorMessage(Path path, List<Throwable> exceptions) {
+    return String.format(
+        "%s: %d %s: %s",
+        path, exceptions.size(), exceptions.size() == 1 ? "exception" : "exceptions", exceptions);
```
werkt, line 319: same as above

werkt (approved):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2152,6 +2153,12 @@ List<Throwable> getExceptions() {
     }
   }
 
+  private static String getErrorMessage(Path path, List<Throwable> exceptions) {
+    return String.format(
+        "%s: %d %s: %s",
+        path, exceptions.size(), exceptions.size() == 1 ? "exception" : "exceptions", exceptions);
```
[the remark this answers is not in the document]

src/main/java/build/buildfarm/worker/shard/CFCExecFileSystem.java
luxe: fixed

src/main/java/build/buildfarm/worker/InputFetcher.java
```diff
@@ -307,4 +309,24 @@ public void run() {
...
+      throws InterruptedException {
+    Operation failedOperation =
+        FailedOperationGetter.get(
+            operationContext.operation,
+            operationContext.queueEntry.getExecuteEntry(),
+            failureMessage,
+            failureDetails);
+
+    try {
+      workerContext.putOperation(failedOperation);
```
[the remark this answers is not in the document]
luxe: afaict, the error path removes the operation and expires the poller so that the operation gets requeued.  It wasn't until I put the operation into the context that I was able to inform the client about the execdir failures.

PR #1271, opened 2023-02-20
[refactor] remove recursion from fetchDirectory

luxe:
`fetchDirectory` is currently recursive. Use a `Stack` to avoid stack overflow.

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -101,6 +101,7 @@ import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.AbstractMap;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -1974,13 +1977,41 @@ private void putDirectoryFiles(
...
   private void fetchDirectory(
-      Path path,
+      Path rootPath,
       Digest digest,
       Map<Digest, Directory> directoriesIndex,
       ImmutableList.Builder<String> inputsBuilder,
       ImmutableList.Builder<ListenableFuture<Path>> putFutures,
       ExecutorService service)
       throws IOException, InterruptedException {
+    Stack<Pair<Path, Directory>> stack = new Stack<>();
```
werkt, line 1987: Why not use Map.Entry (with instantiations via AbstractMap.SimpleEntry) rather than incorporate another apache dependency?

luxe: good idea!  fixed.

werkt (approved):

PR #1273, opened 2023-02-20
[cas] allow copy fallback when out of hardlinks

luxe:


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -1973,6 +1975,26 @@ private void putDirectoryFiles(
...
+    try {
+      Files.createLink(filePath, cacheFilePath);
+    } catch (IOException e) {
+      // propagate the exception if we do not intend a fallback strategy.
+      if (!configs.getWorker().getStorages().get(0).isExecRootCopyFallback()) {
+        throw e;
+      }
+
+      // Instead of hardlinking, copy the file into the exec root.
+      Files.copy(cacheFilePath, filePath, StandardCopyOption.REPLACE_EXISTING);
```
werkt, line 1994: I worry that this is completely silent, and will degrade performance significantly - notification would need to be throttled, but we should figure out some way to inform monitors of this happening, maybe just a counter signal will suffice.

Aside, we should also be able to deal with this by creating additional inodes (i.e. one backing file for a 65k or smaller link count) as a strategy, with pools of the same hash getting replicated.

werkt (approved):

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
luxe: I agree with both those comments.  I've made the following improvements:
 - a counter metric for every time the fallback occurs.
 - added a TODO suggesting the more optimal strategy (minimize copies with extra inodes).
 
 I have an additional concern about CAS expiration.  I'm not sure whether the expiration logic was built around the idea of hardlinks and whether these new copies under `_dir` will cause issues in maintaining the CAS size.

For the time being, this has been very helpful in getting actions to succeed on a cluster that has not yet been configured for xfs.

werkt: Good point. I'm actually less worried about the linked directories and files under reference counting though, and more worried about oversubscription. A proper inode pool would solve both problems, but this is indeed riskier than I realized.

PR #1281, opened 2023-02-28
WriteComplete during fetchBlobs is Success

werkt:
Ensure that completed writes when fetching http blobs are considered successful and delivered with content length responses.

### DIFF
--- a/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
@@ -74,6 +74,7 @@ import build.buildfarm.common.TokenizableIterator;
 import build.buildfarm.common.TreeIterator.DirectoryEntry;
 import build.buildfarm.common.Watcher;
 import build.buildfarm.common.Write;
+import build.buildfarm.common.net.URL;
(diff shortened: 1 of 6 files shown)


src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
```diff
@@ -647,12 +646,26 @@ private static void downloadUri(String uri, ContentOutputStreamFactory getConten
   @Override
   public ListenableFuture<Digest> fetchBlob(
       Iterable<String> uris, Digest expectedDigest, RequestMetadata requestMetadata) {
+    ImmutableList.Builder<URL> urls = ImmutableList.builder();
     for (String uri : uris) {
+      try {
+        urls.add(new URL(new java.net.URL(uri)));
+      } catch (Exception e) {
+        // ignore?
```
luxe, line 654: we could log

luxe (approved):

src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java
werkt: Was temp code for completion. moving to failed future.

PR #1287, opened 2023-03-14
[REAPI] fix buildfarm to support negative priority values

luxe:
Negative priority values are supported according the [REAPI spec](https://github.com/bazelbuild/remote-apis/blob/64cc5e9e422c93e1d7f0545a146fd84fcc0e8b47/build/bazel/remote/execution/v2/remote_execution.proto#L1321).  Our implementation supports this, but we advertised it incorrectly to the client in our capabilities response.


werkt (approved):
Looks good, thanks for the test addition.

PR #1289, opened 2023-03-16
[config] allow configuration of server cache sizes

luxe:
Allow these constants to be overridden in the config.  Backwards compatible; the existing values are set by default.


luxe:
> Great! i want to do this for a while! Thanks Would be great to have some docs about this, but definitely can be done in a separate PR.

thanks! added docs.

PR #1290, opened 2023-03-18
[config] [bugfix] calculate available processors differently for containerized deployments

luxe:
## Problem
Java's `Runtime.getRuntime().availableProcessors()` does not always accurately give the amount of processors available. In some cases it returns 1 due to containerization or virtualization. For example, if you are using k8s with containerd you might see this method give back `1` based on your particular deployment configuration. In reality, the container has access to many more processors. It can cause worker startup times to take ~50 minutes when they would otherwise only take ~5 minutes due to sourcing an inappropriate processor amount.

## Solution
There are other implementations such as [OSHI](https://github.com/oshi/oshi) that use JNA to acquire the native operating system and hardware information which is often more suitable for buildfarm deployments. In order to provide consistency in deriving configuration values, and allocating thread pools, it's best to source the processor count from the same place. We provide an abstraction to derive processor count based on config and environment.

## Config Defaults
~~I've set the default derivation strategy to be `OSHI`, but if we consider that a risk, we can switch it back to `JAVA_RUNTIME` by default.  I feel that the default behaviour of using Java's runtime is more susceptible to giving users a processor count of 1 which is problematic and not easy to detect.~~

There is no configuration changes.  Buildfarm chooses the highest value between implementations.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -3,6 +3,7 @@ defaultActionTimeout: 600
 maximumActionTimeout: 3600
 maxEntrySizeBytes: 2147483648 # 2 * 1024 * 1024 * 1024
 prometheusPort: 9090
+processorsDerive: OSHI
 server:
   instanceType: SHARD
   name: shard
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -28,6 +28,11 @@ server:
   recordBesEvents: false
   clusterId: local
   cloudRegion: us-east-1
+  caches:
+    directoryCacheMaxEntries: 10000
+    commandCacheMaxEntries: 10000
+    digestToActionCacheMaxEntries: 10000
+    recentServedExecutionsCacheMaxEntries: 10000
   admin:
     deploymentEnvironment: AWS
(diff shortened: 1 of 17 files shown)


examples/config.yml
```diff
@@ -3,6 +3,7 @@ defaultActionTimeout: 600
 maximumActionTimeout: 3600
 maxEntrySizeBytes: 2147483648 # 2 * 1024 * 1024 * 1024
 prometheusPort: 9090
+processorsDerive: OSHI
```
80degreeswest, line 6: What if we make this smarter and less complex by doing both and comparing the results? Then picking one we determine to be more accurate (which in this case appears to be OSHI. One thing I'd love to see going forward is to provide for reliable automatic configuration with less manual configs.

luxe: Yeah, exposing a config is not all that important.  I've removed the config.  Buildfarm chooses the highest value between implementations.  Please re-review!

80degreeswest (approved):

PR #1291, opened 2023-03-22
[WIP] Adding auth to fetch urls

brujoand:
This probably requires some discussion, but putting it up here to get that started. 
For one my (ab)use of the URL.java class might need a more specific replacement, and I'm not quite sure how we want to handle Optional vs Null vs "".

### DIFF
--- a/src/main/java/build/buildfarm/common/net/URL.java
+++ b/src/main/java/build/buildfarm/common/net/URL.java
@@ -16,16 +16,28 @@ package build.buildfarm.common.net;
 
 import java.io.IOException;
 import java.net.URLConnection;
+import java.util.Optional;
 
-// solely exists so that we can mock this
 public class URL {
   private final java.net.URL url;
+  private final Optional<String> authHeader;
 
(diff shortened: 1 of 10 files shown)


src/main/java/build/buildfarm/common/net/URL.java
```diff
@@ -16,16 +16,28 @@
...
 import java.io.IOException;
 import java.net.URLConnection;
+import java.util.Optional;
 
-// solely exists so that we can mock this
 public class URL {
   private final java.net.URL url;
+  private final Optional<String> authHeader;
 
+  public URL(java.net.URL url, String authHeader) {
```
brujoand, line 25: So i noticed the comment on line 20, but since this class already had the `openConnection` method I thought hey, this class is already it's own person. So I removed the comment instead. Might need a new name though.

brujoand:
There isn't much I added to what you already had @werkt so feel free to just close this and continue on your implementation, pull my commit into your later PR or let me know another way forward. Although this works locally for us, I'm hesitant to start using it until it's landed on the main branch.

PR #1294, opened 2023-03-22
[Config] Better error when yaml.load fails

jerrymarino:
Today when yaml.load returns a null, it raises a `NullPointerException` when trying to log the config

This gives a better error to make it a bit easier to debug and futher print the path it loaded.


luxe (approved):

luxe:
Nice!
CI will be green after running `./.bazelci/format.sh`

jerrymarino:
Thanks for the review @luxe - loving the new config stuff!

PR #1300, opened 2023-03-29
[refactor] organize thread pools (discuss fixed sizes)

luxe:
### Summary
After organizing all the thread pools we can see that some of them have fixed sizes.  We should consider if they need derived depending on whether they are compute, disk, network bound etc.  For much smaller machines, these thread pools might be too large.  For example, I see servers / workers totaling 400 threads- many of which are in a runnable / blocked state.

### Next
These fixed values should be configurable and derived.  Hoping for some discussion first.


src/main/java/build/buildfarm/common/ExecutorServices.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @class ExecutorServices
+ * @brief Group executor services to easier manage thread counts across software stack.
+ * @details Thread usage is a global concern across the java application. Although these thread
+ *     pools are used in different contexts we group them here so the application's threads can be
+ *     managed and derived from the same class.
+ */
+public class ExecutorServices {
+  public static ExecutorService getScanCachePool() {
```
werkt, line 32: Since all of these create new Executors, please change the name pattern to either `create` or `new` (the latter probably being more fitting).

src/main/java/build/buildfarm/common/ExecutorServices.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.concurrent.Executors;
+
+/**
+ * @class ExecutorServices
+ * @brief Group executor services to easier manage thread counts across software stack.
+ * @details Thread usage is a global concern across the java application. Although these thread
+ *     pools are used in different contexts we group them here so the application's threads can be
+ *     managed and derived from the same class.
+ */
+public class ExecutorServices {
```
werkt, line 31: BuildfarmExecutors? Could we just be the Executors class under this namespace (obviously that will be a pain to reference the util.concurrent version)

werkt (approved):
Ideally we won't have any of these at all. Grpc does its threading thing, IO should be all async, and we only have as many threads as cores in a single per-process pool.

PR #1305, opened 2023-03-30
[images] supply recommended jvm flags

luxe:
This refactors the existing jvm flags used on the repo images and adds additional flags to support buildfarm under containerized environments.   We've seen issues where the RAM was pegged at only 20% of the container, and these flags are provided as recommendations for better ram utilization, a more efficient GC, and minimizing memory usage.


jvm_flags.bzl
```diff
@@ -0,0 +1,79 @@
...
+
+    # By default, the JVM sets the maximum heap size to 25% of the available memory.
+    # It’s quite conservative.  Let's give the heap more space:
+    "-XX:MaxRAMPercentage=80",
+
+    # The -XX:+UseG1GC flag is recommended for use in Kubernetes environments.
+    # This enables the Garbage-First Garbage Collector, which is optimized for
+    # high-throughput, low-latency applications.  It can help reduce
+    # the memory footprint of the JVM.
+    "-XX:+UseG1GC",
```
80degreeswest, line 39: G1 should be the default in all recent versions of Java. Also there is a newer UseZGC one now but I personally haven't tested it with Buildfarm yet.

luxe: We should also consider upgrading past java 11.  Do you use anything higher?  We are on 11 simply because its bazel's default.

80degreeswest: I'm on 19 for a while now. I believe the public images still use 11 though.

luxe: You're right, `G1GC` is used by default in Java 11.  Removing the flag here.

80degreeswest (approved):

PR #1306, opened 2023-03-30
[redis] allow the client to reconnect on redis exceptions

luxe:
- allow the client to reconnect on redis exceptions
 - prevent exceptions from surfacing to other code getting servers / workers stuck.
 - print metrics on redis failures so we can track their occurrence

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisClient.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisClient.java
@@ -28,8 +28,24 @@ import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.jedis.exceptions.JedisException;
 import redis.clients.jedis.exceptions.JedisNoReachableClusterNodeException;
+import java.util.function.Supplier;
+import io.prometheus.client.Counter;
+import lombok.extern.java.Log;
+import java.util.logging.Level;
+import lombok.extern.java.Log;
 
+/**
+ * @class RedisClient
+ * @brief Responsible for making calls to redis.
+ */
+@Log
 public class RedisClient implements Closeable {
+
+  // Metrics to detect any kind of redis failures.
+  // Often due to network issues are the redis cluster going down.
+  private static final Counter redisErrorCounter = Counter.build().name("redis_client_error").help("Count of redis client failures").register();
+  private static final Counter redisClientRebuildErrorCounter = Counter.build().name("redis_client_rebuild_error").help("Count of failures rebuilding redis client").register();
+      
   private static final String MISCONF_RESPONSE = "MISCONF";
 
   @FunctionalInterface
@@ -56,12 +72,21 @@ public class RedisClient implements Closeable {
     }
   }
 
-  private final JedisCluster jedis;
+  // We store the factory in case we want to re-create the jedis client.
+  private Supplier<JedisCluster> jedisClusterFactory;
+
+  // The jedis client.
+  private JedisCluster jedis;
 
   private boolean closed = false;
 
   public RedisClient(JedisCluster jedis) {
     this.jedis = jedis;
+  }
+  
+  public RedisClient(JedisCluster jedis, Supplier<JedisCluster> jedisClusterFactory) {
+    this.jedis = jedis;
+    this.jedisClusterFactory = jedisClusterFactory;
   }
 
   @Override
@@ -81,19 +106,25 @@ public class RedisClient implements Closeable {
   }
 
   public void run(Consumer<JedisCluster> withJedis) throws IOException {
-    call(
+    callEnsureSuccess(
         (JedisContext<Void>)
             jedis -> {
               withJedis.accept(jedis);
               return null;
             });
   }
+  
 
   public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
       throws IOException, InterruptedException {
+    return defaultBlockingCall(withJedis);
+  }
+  
+  private <T> T defaultBlockingCall(JedisInterruptibleContext<T> withJedis)
+      throws IOException, InterruptedException {
     AtomicReference<InterruptedException> interruption = new AtomicReference<>(null);
     T result =
-        call(
+        callEnsureSuccess(
             jedis -> {
               try {
                 return withJedis.run(jedis);
@@ -109,8 +140,41 @@ public class RedisClient implements Closeable {
     return result;
   }
 
-  @SuppressWarnings("ConstantConditions")
   public <T> T call(JedisContext<T> withJedis) throws IOException {
+    return callEnsureSuccess(withJedis);
+  }
+  
+  private <T> T callEnsureSuccess(JedisContext<T> withJedis) {
+    while(true){
+      try {
+        T result = defaultCall(withJedis);
+        return result;
+      }
+      catch (Exception e){
+        redisErrorCounter.inc();
+        log.log(Level.SEVERE,"Failure in RedisClient::call");
+        log.log(Level.SEVERE,e.toString());
+        rebuildJedisCluser();
+      }
+    }
+    
+  }
+  
+  private void rebuildJedisCluser(){
+        try {
+          log.log(Level.SEVERE,"Rebuilding redis client");
+           jedis = jedisClusterFactory.get();
+        }
+        catch (Exception e){
+          redisClientRebuildErrorCounter.inc();
+          log.log(Level.SEVERE,"Failed to rebuild redis client");
+          log.log(Level.SEVERE,e.toString());
+        }
+    
+  }
+
+
+  private <T> T defaultCall(JedisContext<T> withJedis) throws IOException {
     throwIfClosed();
     try {
       try {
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +144,47 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
...
+    }
+  }
+
+  private void rebuildJedisCluser() {
+    try {
+      System.out.println("Rebuilding redis client");
+      jedis = jedisClusterFactory.get();
+    } catch (Exception e) {
+      redisClientRebuildErrorCounter.inc();
+      System.out.println("Failed to rebuild redis client");
```
jerrymarino, line 182: Should this plumb in log?

jerrymarino (commented):
@luxe nice! I was wondering if you able to continue a worker after redis transient failure after this PR end to end?

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +144,47 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
...
+    // Alternatively,
+    // Capture all redis problems at the client level.
+    // Try to re-establish the client and log all issues.
+    // This will block the overall thread until redis can be connected to.
+    // It may be a useful strategy for gaining stability on a poorly performing network,
+    // or a redis cluster that goes down.
+    while (true) {
+      try {
+        return defaultCall(withJedis);
+      } catch (Exception e) {
```
jerrymarino, line 167: Should this loop and the other one break on an `InterruptedException`?

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +144,47 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
...
+      return defaultCall(withJedis);
+    }
+
+    // Alternatively,
+    // Capture all redis problems at the client level.
+    // Try to re-establish the client and log all issues.
+    // This will block the overall thread until redis can be connected to.
+    // It may be a useful strategy for gaining stability on a poorly performing network,
+    // or a redis cluster that goes down.
+    while (true) {
```
80degreeswest, line 164: I would prefer configuration to be a number of a reconnects. If it's 0 or null then we don't retry, otherwise we retry up to that number of times. Bonus if there is some backoff here where we don't spam retries continuously.

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: fixed

luxe:
> @luxe nice! I was wondering if you able to continue a worker after redis transient failure after this PR end to end?

Added a testing section to the PR.  I've observed that the servers/workers can keep running while redis is down.  However, **new** workers fail to start up when redis is down.  We should address that use-case as well.

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: Agree.  Switched to retry amount + duration between retries.  We have a `Retrier.java` class in this repo, but its too specific to grpc so I didn't use it.  Might be nice to use something like https://resilience4j.readme.io/docs/retry in the future.

luxe:
> > @luxe nice! I was wondering if you able to continue a worker after redis transient failure after this PR end to end?
> 
> Added a testing section to the PR. I've observed that the servers/workers can keep running while redis is down. However, **new** workers will fail to start up when redis is down. We should address that use-case as well.

Fixed.  If you also start a worker without redis, it will keep trying to establish the client on startup.  When redis is available, the worker will complete its startup.

jerrymarino:
Yeah I think this will definitely make it easier to deal with redis timeouts. Recovering from `MatchStage` hang I've got due to this elasticache may be handled here if a critical redis call inside of the `MatchStage`? FWIW it took about 1 hour for that update to go through so it might not have been a standard outcome. I'm going to try this PR a bit today too 

```
Apr 05, 2023 1:15:22 PM build.buildfarm.worker.PipelineStage run
SEVERE: MatchStage::run(): stage terminated due to exception
redis.clients.jedis.exceptions.JedisDataException: UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)
        at redis.clients.jedis.Protocol.processError(Protocol.java:132)
        at redis.clients.jedis.Protocol.process(Protocol.java:166)
        at redis.clients.jedis.Protocol.read(Protocol.java:220)
        at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:389)
        at redis.clients.jedis.Connection.getBinaryBulkReply(Connection.java:299)
        at redis.clients.jedis.Connection.getBulkReply(Connection.java:289)
        at redis.clients.jedis.Connection$1.call(Connection.java:283)
        at redis.clients.jedis.Connection$1.call(Connection.java:280)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1589)
```

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +155,57 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
...
+    // It may be a useful strategy for gaining stability on a poorly performing network,
+    // or a redis cluster that goes down.  For example, if you need to perform redis maintance on a
+    // production cluster and do not want to affect build clients.
+    for (int i = 0; i < reconnectClientAttempts; ++i) {
+      try {
+        return defaultCall(withJedis);
+      } catch (Exception redisException) {
+        // Record redis failure.
+        redisErrorCounter.inc();
+        log.log(
```
jerrymarino, line 175: Minor: do we want SEVERE here or ERROR? I wasn’t sure if the semantics of SEVERE weren’t recoverable.

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +155,57 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
     return result;
   }
 
-  @SuppressWarnings("ConstantConditions")
   public <T> T call(JedisContext<T> withJedis) throws IOException {
+    return callImpl(withJedis);
```
jerrymarino, line 159: Minor: do we need to make a separate method for this still?

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -109,8 +155,57 @@ public <T> T blockingCall(JedisInterruptibleContext<T> withJedis)
...
+    }
+
+    // If all attempts to re-establish the redis client has failed,
+    // perform the redis command and allow the exception to be thrown.
+    return defaultCall(withJedis);
+  }
+
+  private void rebuildJedisCluser() {
+    try {
+      log.log(Level.SEVERE, "Rebuilding redis client");
```
jerrymarino, line 199: Is this redundant with the log in thr upstream call? E.g. the caller already logged when it catches?

jerrymarino (approved):
This definitely helps with some of the redis problems I was hitting, working good in general. The only possible thing I thought of was back off: perhaps could propose as a followup: if there was packet loss in a big spike backoff could help recovery, but overall this is a large improvement 🥳

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: switched to ERROR

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: fixed.  callImpl folded into call

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: agree, removed

luxe:
> This definitely helps with some of the redis problems I was hitting, working good in general. The only possible thing I thought of was back off: perhaps could propose as a followup: if there was packet loss in a big spike backoff could help recovery, but overall this is a large improvement partying_face

Backoff is a good idea.  We should do it as followup.  Our redis cluster is currently taking thousands of requests per second.  So when the client enters this "retry" state now its already significantly less traffic than normal.   

There's also a lot of places in buildfarm where we want to rety things, so I like the idea of having the same retry framework everywhere with configurable backoffs; ex:  all the network calls between server/worker.  The queues which have been nonblocking, etc.

src/main/java/build/buildfarm/common/redis/RedisClient.java
luxe: Oh there is no `Level.ERROR`! 
switched to`Level.WARNING`.

jerrymarino:
> There's also a lot of places in buildfarm where we want to rety things, so I like the idea of having the same retry framework everywhere with configurable backoffs; ex: all the network calls between server/worker. The queues which have been nonblocking, etc.

Yeah this would help improve some of the failure modes I've put into recently 👍! Happy to help where I can on it

PR #1308, opened 2023-04-04
[Worker] Catch all in ReportResultStage

jerrymarino:
I've got a situation where a worker is missing directories upon tearing down an action. When it throws an exception the worker becomes stuck forever. This makes it more resilient to benign errors and catches them all for safety to prevent a worker being wedged. Here, wedged meant that the pipeline stage, `ReportResultStage`, thread has exited due to the `IllegalStateException` and worker process remains running.

I get the stack:
```
SEVERE: ReportResultStage::run(): stage terminated due to exception
java.lang.IllegalStateException: inputDirectory dbbcb4da7a1858c1debab42054b872a60a64072fe34349b01047e5930d2e3721/79 is not in directoryStorage
        at build.buildfarm.cas.cfc.CASFileCache.decrementReferencesSynchronized(CASFileCache.java:1651)
        at build.buildfarm.cas.cfc.CASFileCache.decrementReferences(CASFileCache.java:1612)
        at build.buildfarm.worker.shard.CFCExecFileSystem.destroyExecDir(CFCExecFileSystem.java:444)
        at build.buildfarm.worker.shard.ShardWorkerContext.destroyExecDir(ShardWorkerContext.java:719)
        at build.buildfarm.worker.ReportResultStage.after(ReportResultStage.java:203)
        at build.buildfarm.worker.PipelineStage.iterate(PipelineStage.java:129)
        at build.buildfarm.worker.PipelineStage.runInterruptible(PipelineStage.java:44)
        at build.buildfarm.worker.PipelineStage.run(PipelineStage.java:51)
        at java.base/java.lang.Thread.run(Thread.java:1589)
```

As a side note, I've been looking to insulate against this case on a macOS VM based setup. The healthcheck endpoint seems like a possible place we could test pipeline status; for something like when the `ReportResultStage` threads or other vital threads exit but don't terminate the worker.

### DIFF
--- a/src/main/java/build/buildfarm/worker/ReportResultStage.java
+++ b/src/main/java/build/buildfarm/worker/ReportResultStage.java
@@ -203,7 +203,7 @@ public class ReportResultStage extends PipelineStage {
       workerContext.destroyExecDir(operationContext.execDir);
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
-    } catch (IOException e) {
+    } catch (Exception e) {
       log.log(
           Level.SEVERE,
           String.format("error destroying exec dir %s", operationContext.execDir.toString()),
--- a/src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
+++ b/src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
@@ -715,7 +715,8 @@ class ShardWorkerContext implements WorkerContext {
   // might want to split for removeDirectory and decrement references to avoid removing for streamed
   // output
   @Override
-  public void destroyExecDir(Path execDir) throws IOException, InterruptedException {
+  public void destroyExecDir(Path execDir)
+      throws IllegalStateException, IOException, InterruptedException {
     execFileSystem.destroyExecDir(execDir);
   }
 


werkt:
There's a lot wrong here. tl;dr: I want this worker to fall down and exit. You don't want to continue on in this state, similar to other recent conversations around stage resiliency and catch-alls. We can't catch this.

First, did you take a jstack to observe the full wedge? This situation really should be shutting down the worker overall, since the ReportResult stage has died.

The after() call is in a problematic place, where a throw obviously shuts the thread down, but the queue/object is still live, and the claim has been released, allowing something else to acquire it. The jstack might shed some more light on this.

Your state with the directory being reference-decremented not being in the directory storage is problematic indeed: it means that expireDirectory was called (no other way) on that directory between the beginning and end of execution. Only likely way for this would include a SEVERE log message saying "directory {directory} does not exist in cache, purging it with fire and resorting to fetch", which should have resulted in another directory matching its name being placed into storage. See if you can locate that message in your logs.

We've also seen insane things happening on MacOS filesystems ever since they've been tried here: do the best you can here to verify that nothing is changing your filesystem asynchronously and burning directories or files under the cache dir.

jerrymarino:
Thanks for the review and taking the time out of your day for the feedback. OK TLDR; looking to see a if I can reasonably exit them in-band

> You don't want to continue on in this state, similar to other recent conversations around stage resiliency and catch-alls. 

Yeah I agree with this in general; terminating it is definitely ideal. In practice, though dropping workers is caveated according to how the buildfarm-worker process is managed. I imagine there are some potential fixes to get it self-healing back to a an acceptable state and perhaps exiting the worker will help this be more loud ( and better than leaving it stuck if this is more widespread ). 


>Only likely way for this would include a SEVERE log message saying "directory {directory} does not exist in cache, purging it with fire and resorting to fetch", which should have resulted in another directory matching its name being placed into storage. See if you can locate that message in your logs.

Yep I do end up seeing this in the log, and I can confirm the files it complained about e.g. `/tmp/worker/cache/70f0ebb95660873622af37cc9e22efec19b1ef48e307abe660fc4f954fc8ac9c_dir/x/x/x/x/x.runfiles/x/tools/__init__.py` are in the path which was state as not missing. 

The last bits of this logfile I see are:
```
Apr 02, 2023 5:55:54 AM build.buildfarm.cas.cfc.CASFileCache directoryExists
SEVERE: directory file entry /tmp/worker/cache/eabfaefae99767e3052f35f9a0e25b2056bf63c15f9809c4d86a7542c7ee9ff1_dir/x/x/x/x-descriptor-set.proto.bin does not exist
Apr 02, 2023 5:55:54 AM build.buildfarm.cas.cfc.CASFileCache putDirectorySynchronized
SEVERE: directory eabfaefae99767e3052f35f9a0e25b2056bf63c15f9809c4d86a7542c7ee9ff1_dir does not exist in cache, purging it with fire and resorting to fetch
Apr 02, 2023 5:56:36 AM build.buildfarm.cas.cfc.CASFileCache directoryExists
SEVERE: directory file entry /tmp/worker/cache/70f0ebb95660873622af37cc9e22efec19b1ef48e307abe660fc4f954fc8ac9c_dir/x/x/x/x/x.runfiles/x/tools/__init__.py does not exist
Apr 02, 2023 5:56:36 AM build.buildfarm.cas.cfc.CASFileCache putDirectorySynchronized
SEVERE: directory 70f0ebb95660873622af37cc9e22efec19b1ef48e307abe660fc4f954fc8ac9c_dir does not exist in cache, purging it with fire and resorting to fetch
Apr 02, 2023 5:56:36 AM build.buildfarm.worker.PipelineStage run
SEVERE: ReportResultStage::run(): stage terminated due to exception
java.lang.IllegalStateException: inputDirectory dbbcb4da7a1858c1debab42054b872a60a64072fe34349b01047e5930d2e3721/79 is not in directoryStorage
        at build.buildfarm.cas.cfc.CASFileCache.decrementReferencesSynchronized(CASFileCache.java:1651)
        at build.buildfarm.cas.cfc.CASFileCache.decrementReferences(CASFileCache.java:1612)
        at build.buildfarm.worker.shard.CFCExecFileSystem.destroyExecDir(CFCExecFileSystem.java:444)
        at build.buildfarm.worker.shard.ShardWorkerContext.destroyExecDir(ShardWorkerContext.java:719)
        at build.buildfarm.worker.ReportResultStage.after(ReportResultStage.java:203)
        at build.buildfarm.worker.PipelineStage.iterate(PipelineStage.java:129)
        at build.buildfarm.worker.PipelineStage.runInterruptible(PipelineStage.java:44)
        at build.buildfarm.worker.PipelineStage.run(PipelineStage.java:51)
        at java.base/java.lang.Thread.run(Thread.java:1589)

````

One of the workers then has this message - proceeding the common stack:
```

Apr 02, 2023 5:56:49 AM build.buildfarm.worker.shard.ShardWorkerContext$1 onEntry
WARNING: matched duplicate operation shard/operations/d740e928-bb96-4671-9ae7-f68ee0f4d641
Apr 02, 2023 5:56:59 AM build.buildfarm.worker.shard.ShardWorkerContext$1 onEntry
WARNING: matched duplicate operation shard/operations/d740e928-bb96-4671-9ae7-f68ee0f4d641
```

What's more interesting is that I have observed the _identicial_ exception on a large number of workers for this file
```
SEVERE: ReportResultStage::run(): stage terminated due to exception
java.lang.IllegalStateException: inputDirectory dbbcb4da7a1858c1debab42054b872a60a64072fe34349b01047e5930d2e3721/79 is not in directoryStorage
```

Yeah I've been tailing this for a bit, and I I've confirmed a number of workers had a similar jstack - I attached it below. I'll still look to prep a fix in hopes exit when `destroyExecDir` raises something than an ioexception or interrupt exception 
[jstack_worker_hung_report_stage.txt](https://github.com/bazelbuild/bazel-buildfarm/files/11153602/jstack_worker_hung_report_stage.txt)

jerrymarino:
> We've also seen insane things happening on MacOS filesystems ever since they've been tried here: do the best you can here to verify that nothing is changing your filesystem asynchronously and burning directories or files under the cache dir.

For sure, also Insane is a nice way to put what is happening here 👍 . This just doesn't replicate very easily and seems to pop under heavy load. These file systems have had extra care and dark magic in some of our rules to make them work reliably in general

jerrymarino:
This PR https://github.com/bazelbuild/bazel-buildfarm/pull/1309 invokes a callback in any `PipelineStage` when there's an uncaught exemption to term the worker

PR #1309, opened 2023-04-06
[Worker][Resilience] Graceful term on PipelineStage failure

jerrymarino:
Today a pipeline may get into a undefined state but leave a worker in a hung but running state. This PR gracefully terms the worker when catching exeptions benign to a `PipelineStage`

Do this by means of a `ThreadGroup` which has an uncaught exception handler. This design affords catching exceptions from all pipelines and delegating actions to the worker, which makes an attempt to interrupt stage threads to shut down. It necessitated removing the catch all exception swallowing feature of the `PipelineStage`.

Currently a worker will `exitPostPipelineFailure` but I imagine folks can extend this and provide parameter to take other actions.


jerrymarino:
@werkt after our thread the other day I spent more time auditing historical exceptions when a worker would get stuck. There were several other exceptions and failure than I originally thought; a few other stacks for `PipelineStage` and also `MatchStage` has contributed to worker hangs during ElastiCache events:

```
SEVERE: MatchStage::run(): stage terminated due to exception
redis.clients.jedis.exceptions.JedisDataException: UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)
```

I was thinking about a followup fix to provide an experimental option to spin up a new pipeline instance. For some of these failures we may want to govern the restart time to avoid a thundering herd.

luxe:
This PR is similar in spirit to: https://github.com/bazelbuild/bazel-buildfarm/pull/1040
I agree that in both cases, @jerrymarino and I are taping over the problems to keep the workers up an running.  It's just useful to do that from an operational perspective.

However, it sounds like the direction we need to go now is to make sure workers completely shutdown when these problems occur-- so they don't continue to manifest additional problems.

As it stands right now, the worker pipeline can get stuck in such a way that the worker stops processing, but the java program continues running.  It's hard to know when to restart the container since there is not a clear health metric around this degradation, and its less easy to tell k8s to restart containers when throughput dies down on a worker.

---

For @jerrymarino's problem specifically it seems to manifest from redis issues, and we know his elastic cache is very small-- however I'm not sure what to make of that master/replica error.  

Here is a related PR around stability for redis:  https://github.com/bazelbuild/bazel-buildfarm/pull/1306
Can't say it will solve this problem specifically, but I want it personally so that the entire redis database can be taken down and brought back up without affecting ongoing builds.  Also, we've had network problems where redis calls timeout, so establishing the client has been helpful for uptime.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -186,6 +188,16 @@ public void prepareWorkerForGracefulShutdown() {
     }
   }
 
+  private void exitPostPipelineFailure() {
+    // Shutdown the worker if a pipeline fails. By means of the spring lifecycle
+    // hooks - e.g. the `PreDestroy` hook here - it will attempt to gracefully
+    // spin down the pipeline
+    // Consider defining exit codes to better afford out of band instance
+    // recovery
+    int code = SpringApplication.exit(springContext, () -> 1);
```
jerrymarino, line 197: This line needs a timeout incase it gets stuck during the shutdown workflow:
```
        at java.lang.Object.wait0(java.base@20/Native Method)
        - waiting on <no object reference available>
        at java.lang.Object.wait(java.base@20/Object.java:366)
        at java.lang.Thread.join(java.base@20/Thread.java:2109)
        - locked <0x0000000740a01fc0> (a java.lang.Thread)
        at build.buildfarm.worker.Pipeline.join(Pipeline.java:182)
        at build.buildfarm.worker.Pipeline.close(Pipeline.java:96)
        at build.buildfarm.worker.shard.Worker.stop(Worker.java:647)
        at java.lang.invoke.LambdaForm$DMH/0x00000008011f0000.invokeVirtual(java.base@20/LambdaForm$DMH)
        at java.lang.invoke.LambdaForm$MH/0x0000000801516000.invoke(java.base@20/LambdaForm$MH)
        at java.lang.invoke.Invokers$Holder.invokeExact_MT(java.base@20/Invokers$Holder)
        at jdk.internal.reflect.DirectMethodHandleAccessor.invokeImpl(java.base@20/DirectMethodHandleAccessor.java:154)
        at jdk.internal.reflect.DirectMethodHandleAccessor.invoke(java.base@20/DirectMethodHandleAccessor.java:104)
        at java.lang.reflect.Method.invoke(java.base@20/Method.java:578)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleElement.invoke(InitDestroyAnnotationBeanPostProcessor.java:389)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleMetadata.invokeDestroyMethods(InitDestroyAnnotationBeanPostProcessor.java:347)
        at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor.postProcessBeforeDestruction(InitDestroyAnnotationBeanPostProcessor.java:177)
        at org.springframework.beans.factory.support.DisposableBeanAdapter.destroy(DisposableBeanAdapter.java:197)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroyBean(DefaultSingletonBeanRegistry.java:587)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingleton(DefaultSingletonBeanRegistry.java:559)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingleton(DefaultListableBeanFactory.java:1163)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.destroySingletons(DefaultSingletonBeanRegistry.java:520)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.destroySingletons(DefaultListableBeanFactory.java:1156)
        at org.springframework.context.support.AbstractApplicationContext.destroyBeans(AbstractApplicationContext.java:1106)
        at org.springframework.context.support.AbstractApplicationContext.doClose(AbstractApplicationContext.java:1075)
        at org.springframework.context.support.AbstractApplicationContext.close(AbstractApplicationContext.java:1021)
        - locked <0x0000000740a09f08> (a java.lang.Object)
        at org.springframework.boot.SpringApplication.close(SpringApplication.java:1365)
        at org.springframework.boot.SpringApplication.exit(SpringApplication.java:1352)
```

jerrymarino:
> However, it sounds like the direction we need to go now is to make sure workers completely shutdown when these problems occur-- so they don't continue to manifest additional problems.

Yeah this is my general hope - if the workers hit an exception hasn't been handled by a given `PipelineStage` implementation then we'd just exit the worker.  Reason being; each pipeline stage would have to recover from specific exceptions: it's each `PipelineStage`'s responsibility to catch and recover from exceptions such e.g. the `CFCExecFileSystem` cleanup exception we worked around in the execution phase. Otherwise, the existing catch all in the root `PipelineStage` impl we've got today leads to workers getting stuck forever.. 

Today, when this happens I personally have to go in and term the workers with an out of band program. This gets me out of the situation as I can spinup the worker processes and makes it self healing against most of the exceptions; which are happening but at very low rate

jerrymarino:
>Here is a related PR around stability for redis: https://github.com/bazelbuild/bazel-buildfarm/pull/1306
> Can't say it will solve this problem specifically, but I want it personally so that the entire redis database can be taken down and brought back up without affecting ongoing builds.

For the redis issue, today I can post up the stack that I was hitting. I've been doing an incremental rollout and scaling it in real time as we send more traffic/actions through it. Initially, I started with a what turned out to be clown-car sized elasticache: TTL was too large for the action throughput. It worked fine for running 1 jenkins job for months but caused worker hangs during scaling.

My realization is I may hit this again and will need to keep scaling redis as I add many other mono-repos. If the elasticache cluster is facing this severe load; perhaps the "graceful" failover implementation in elasticache is unhandled end to end. IME it falls back to a read replica ( potentially facing degrdation too ).

werkt:
> For the redis issue, today I can post up the stack that I was hitting. I've been doing an incremental rollout and scaling it in real time as we send more traffic/actions through it. Initially, I started with a what turned out to be clown-car sized elasticache: TTL was too large for the action throughput. It worked fine for running 1 jenkins job for months but caused worker hangs during scaling.

These are always the worst conditions: we need to do a better job of allowing you to specify an SLA in units and presenting  the nature of a violation like this, in addition to changes similar to this which will permit seamless (albeit unideal) operation in the face of it. "Too small" is always relative, and has to be described in terms of period/usage.

> My realization is I may hit this again and will need to keep scaling redis as I add many other mono-repos. If the elasticache cluster is facing this severe load; perhaps the "graceful" failover implementation in elasticache is unhandled end to end. IME it falls back to a read replica ( potentially facing degrdation too ).

As you hit these, identify where monitoring would have told you that you hit a limit or degraded behavior or minimum lifetimes here: I can think of a few, but the gold standard is field testing.

jerrymarino:
> specify an SLA in units and presenting the nature of a violation like this, in addition to changes similar to this which will permit seamless (albeit unideal) operation in the face of it.

Exceeding these limits put it into a difficult position as-is; I'd love to know your suggestions to this ( even if not incremental changes ). The few I kicked around are having a mode which starts dynamically scaling eviction policies and extending key TTLs. Having some ability which kicks in to reconcile things might dovetail into handling hot keys. 

Today the safe answer is setting the floor to the worst case which assume no operations are ever reused. Computing the bytes of the data structures we put there will allow for a forecast and capacity plan; I'd like arrive at some specs like:
_Redis will store 1 million results per `$nBytes` of memory. e.g. a build with `$nActions` actions a `$nDay` TTL doing `$nBuilds` per day at a `$percent` cache invalidation rate_


>As you hit these, identify where monitoring would have told you that you hit a limit or degraded behavior or minimum lifetimes here: I can think of a few, but the gold standard is field testing.

Yeah absolutely agree with this. My current thinking is to forecast my runway based on such trends, the achieving auto-scaling with the smallest margins 🙈 - but I think it's orthogonal to the other improvements we can make.

src/main/java/build/buildfarm/worker/shard/Worker.java
jerrymarino: In the latest revision, I've added 2 mechanisms to guard against the condition which I hit in production on the first version of this PR. The [deadlining of a shutdown ensures](https://github.com/bazelbuild/bazel-buildfarm/pull/1309/files#diff-43afc2d02d6395012fade1d0104600664bf1747d9ed8fbffe48f666f396f1f92R202) provides better guarantees we don't exceed a set threshold in this case.

jerrymarino:
Over time I'm still getting a bag of misc exceptions in `ReportResultStage` and others that'd have taken the worker downs and caused indefinite queing .

Do you think this is OK to land as-is or would it be more reasonable to add it behind a switch and off - we could conditionally catch/re-throw to preserve the existing catch-all in the short term? cc @luxe who's also got some workarounds for this 

```
SEVERE: ReportResultStage: stage terminating due to uncaught exception
io.grpc.StatusRuntimeException: UNKNOWN
        at io.grpc.Status.asRuntimeException(Status.java:539)
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:487)
        at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:576)
        at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:70)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:757)
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:736)
        at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1623)

```

src/main/java/build/buildfarm/worker/Pipeline.java
```diff
@@ -27,22 +27,73 @@ public class Pipeline {
...
+            }
+
+            // Bail out to avoid duplicating callbacks in highly concurrent
+            // failure modes
+            synchronized (this) {
+              if (handlingUncaughtException) {
+                return;
+              }
+              handlingUncaughtException = true;
+            }
```
werkt, line 71: Won't be necessary with the future model.

src/main/java/build/buildfarm/worker/Pipeline.java
```diff
@@ -27,22 +27,73 @@ public class Pipeline {
...
+                    ex);
+                break;
+              }
+            }
+            // Interrupt the entire threadgroup
+            try {
+              interrupt();
+            } catch (Exception e) {
+              log.log(Level.SEVERE, "Pipeline stage interrupt caught exception", e);
+            }
```
werkt, line 62: There are no checked exceptions from interrupt() - you shouldn't need this block. Add a finally that sets the future if you're truly concerned about this.

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -185,6 +190,42 @@ public void prepareWorkerForGracefulShutdown() {
...
+                System.exit(1);
+              }
+            },
+            forceShutdownDeadline,
+            SECONDS);
+
+    // Consider defining exit codes to better afford out of band instance
+    // recovery
+    int code = SpringApplication.exit(springContext, () -> 1);
+    termFuture.cancel(false);
```
werkt, line 225: Kill the scheduled executor here as well.

werkt (commented):
I reviewed, but I'm still concerned - this seems very duct-tape like and doesn't enumerate our ongoing issues very well. It may similarly mask them for misbehaving workers, where we can apply effort to resolving them.

jerrymarino (commented):
> I reviewed, but I'm still concerned - this seems very duct-tape like and doesn't enumerate our ongoing issues very well. 

Thanks for the review and feedback, going to cleanup the comments here this afternoon. My thought is if we don't have this; a worker can tilt / enter an undefined state but remain consuming resources and/or appear healthy from the rest of the systems' perspective.

> It may similarly mask them for misbehaving workers, where we can apply effort to resolving them.
I don't think these have to be mutually exclusive; once we do fix the ongoing ones IME it's still ideal to exit if a worker if it isn't doing something useful as a safety/stopgap

jerrymarino:
@werkt WDYT about landing this? Unless we live world w/o bugs in a worker - I've either gotta keep this patched on or write some ugly way to determine they tilted and term them - but here we can atleast be more graceful and intentional.

I get burning the CAS isn't great - maybe we can have a flag to turn this off if that's your main concern

PR #1312, opened 2023-04-10
[CAS] replace SharedLock with ReentrantLock

luxe:


werkt:
Can you explain the benefit here?

luxe:
Probably no benefit.  Trying to simplify CAS more.  The implementation looks like a reentrantlock, but based on upstream CI this may not work.

PR #1319, opened 2023-04-21
[execution] allow tmpfs and cgroups enforcement

luxe:
### tmpfs
Previously we required bazel clients to opt-in with an exec_property.  Now buildfarm has the ability to enforce tmpfs usage as part of deployment configuration.
See:  `alwaysUseTmpFs: true`

### cgroups
cgroups was previously enabled via `limitGlobalExecution`.  However, there was no way to opt-out of cgroups and opt-in to other restrictions.  limitGlobalExecution is still required for accessing restrictions, but it will no longer guarantee cgroups is used.  For that, deployments must set:
See:  `alwaysUseCgroups: true`

### Additional context
We may also want to experiment with sandbox, but not cgroups in our deployment.  There is currently no way to do this because enabling sandbox also requires enabling  `limitGlobalExecution`, but doing so enables cgroups.   By making these settings more granular we are able opt into and enforce particular execution wrappers.

### DIFF
--- a/_site/docs/configuration/configuration.md
+++ b/_site/docs/configuration/configuration.md
@@ -291,11 +291,12 @@ worker:
 
 ### Sandbox Settings
 
-| Configuration | Accepted and _Default_ Values | Description                                       |
-|---------------|-------------------------------|---------------------------------------------------|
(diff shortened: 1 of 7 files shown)


werkt (approved):
Looks good, thanks!

PR #1320, opened 2023-04-22
Incorporate rules_{python,proto} for bazel_tools

werkt:
Since bazelbuild/bazel#17545, bazel_tools apparantly requires rules_python and rules_proto to successfully parse under bazel (load lines only). We only use affected bazel_tools (src/main/protobuf) in persistentworkers. Import these repositories to reinstate downstream CI passing.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -8,23 +8,14 @@ load(":defs.bzl", "buildfarm_init")
 
 buildfarm_init()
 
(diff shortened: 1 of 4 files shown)


werkt:
This will break compatibility with the ubuntu1804 target build image due to python3.6 deprecation from https://github.com/pypa/installer, upgraded through https://github.com/bazelbuild/rules_python. Closing in favor of a fork of the worker_protocol java proto target.

PR #1325, opened 2023-04-24
[UI] Add web frontend to buildfarm servers

luxe:
### Context:
Buildfarm needs a web UI.  For starters, let's implement an optional spring controller on the servers so that users can start navigating from the browser.  We may want a standalone web server in the future, but from a deployment perspective, its easier to keep our already load balanced frontend and extend it to serve http.
### How to Use:
Configure the server:
```
ui:
    enable: true
    port: 80
```
Provide bazel with `--bes_results_url`.  This will inform the developers where to look, and they can start introspecting builds on their own.  This implements the landing page for invocations and links to the operation data.

bfui


werkt (approved):
A nice start. Let's go.

src/main/java/build/buildfarm/server/WebController.java
```diff
@@ -0,0 +1,291 @@
+// Copyright 2023 The Bazel Authors. All rights reserved.
```
80degreeswest, line 1: I think we should put all UI related code in a separate ui module (controllers, etc) as I think this will end up growing and hopefully host bfadmin as well.

80degreeswest (approved):
This is awesome!

src/main/java/build/buildfarm/server/WebController.java
luxe: moved to controllers

PR #1328, opened 2023-04-26
Restore storage-only worker registration.

werkt:
Storage workers were not considered part of the available pool of write targets. Workers in general will be considered at once one or more of EXECUTE, or STORAGE types. This restores functionality, registers execute workers for the first time, and paves the way for cluster awareness of execution.

Of note, execute workers will not be presented currently through the backplane. Further changes to non-read/write behaviors will follow for them once storage-only is proven working.

### DIFF
--- a/src/main/java/build/buildfarm/common/config/Worker.java
+++ b/src/main/java/build/buildfarm/common/config/Worker.java
@@ -46,14 +46,15 @@ public class Worker {
     }
   }
 
-  public WorkerType getWorkerType() {
-    if (getCapabilities().isCas() && getCapabilities().isExecution()) {
-      return WorkerType.EXECUTE_AND_STORAGE;
-    }
+  public int getWorkerType() {
+    int workerType = 0;
     if (getCapabilities().isCas()) {
-      return WorkerType.STORAGE;
+      workerType |= WorkerType.STORAGE.getNumber();
     }
-    return WorkerType.EXECUTE;
+    if (getCapabilities().isExecution()) {
+      workerType |= WorkerType.EXECUTE.getNumber();
+    }
+    return workerType;
   }
 
   public Path getValidRoot() throws ConfigurationException {
--- a/src/main/java/build/buildfarm/instance/shard/DistributedState.java
+++ b/src/main/java/build/buildfarm/instance/shard/DistributedState.java
@@ -28,7 +28,7 @@ import build.buildfarm.common.redis.RedisMap;
 public class DistributedState {
   /**
    * @field executeWorkers
-   * @brief All of the execute-only workers register themselves to the cluster.
+   * @brief All of the execute-capable workers register themselves to the cluster.
    * @details This is done to keep track of which machines are online and known by the rest of the
    *     cluster.
    */
@@ -36,19 +36,11 @@ public class DistributedState {
 
   /**
    * @field storageWorkers
-   * @brief All of the storage-only workers register themselves to the cluster.
+   * @brief All of the storage-capable workers register themselves to the cluster.
    * @details This is done to keep track of which machines are online and known by the rest of the
    *     cluster.
    */
   public RedisHashMap storageWorkers;
-
-  /**
-   * @field executeAndStorageWorkers
-   * @brief All of the workers with both execute and storage register themselves to the cluster.
-   * @details This is done to keep track of which machines are online and known by the rest of the
-   *     cluster.
-   */
-  public RedisHashMap executeAndStorageWorkers;
 
   /**
    * @field prequeue
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -141,7 +141,7 @@ public class RedisShardBackplane implements Backplane {
   private ExecutorService subscriberService = null;
   private @Nullable RedisClient client = null;
 
-  private final Set<String> workerSet = Collections.synchronizedSet(new HashSet<>());
+  private final Set<String> storageWorkerSet = Collections.synchronizedSet(new HashSet<>());
   private long workerSetExpiresAt = 0;
 
   private DistributedState state = new DistributedState();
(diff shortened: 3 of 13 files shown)


src/main/java/build/buildfarm/common/config/Worker.java
```diff
@@ -46,14 +46,15 @@ public ExecutionPolicy[] getExecutionPolicies() {
     }
   }
 
-  public WorkerType getWorkerType() {
-    if (getCapabilities().isCas() && getCapabilities().isExecution()) {
-      return WorkerType.EXECUTE_AND_STORAGE;
-    }
+  public int getWorkerType() {
```
luxe, line 49: It would have been nice to preserve the enum so all the caller code was easier to read.  However, I see the value of bitwise flags here as well.

src/main/java/build/buildfarm/instance/shard/DistributedState.java
```diff
@@ -28,28 +28,20 @@
...
   public RedisHashMap executeWorkers;
 
   /**
    * @field storageWorkers
-   * @brief All of the storage-only workers register themselves to the cluster.
+   * @brief All of the storage-capable workers register themselves to the cluster.
    * @details This is done to keep track of which machines are online and known by the rest of the
    *     cluster.
    */
   public RedisHashMap storageWorkers;
```
luxe, line 43: To echo a previous comment from https://github.com/bazelbuild/bazel-buildfarm/pull/1152,
We may want to store all the workers together and filter as needed based on capabilities.  i.e. reduce this down even further to a single HashMap.  Not necessary to change now, but throwing it out as a suggestion.

luxe (approved):
Thanks.  Its cool to see users trying to build storage-only clusters.  It's a unique configuration that was never really tested and has become feasible now through other initiatives + these fixes.  

"execute-only" workers began specifically to support GPU hosts that should not participate in providing CAS (Although we didn't really have the moniker for it then).  "storage-only" was partially implemented afterward as we thought about bwob and Yuriy wanted to grow the CAS in a cost effective way.  Nice to see this unify further and start being used by the public.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -257,19 +257,19 @@ message S3BucketConfig {
 }
 
 enum WorkerType {
-  EXECUTE = 0;
+  NONE = 0;
```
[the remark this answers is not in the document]
werkt: Right, this being the bitfield. That or we switch whole hog to a set of booleans and abandon the idea of 'WorkerType' (and probably consolidate the config capabilities in the process)

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -604,14 +604,7 @@ public void start() throws ConfigurationException, InterruptedException, IOExcep
...
         HealthStatusManager.SERVICE_NAME_ALL_SERVICES, ServingStatus.SERVING);
     PrometheusPublisher.startHttpServer(configs.getPrometheusPort());
-    // Not all workers need to be registered and visible in the backplane.
-    // For example, a GPU worker may wish to perform work that we do not want to cache locally for
-    // other workers.
-    if (configs.getWorker().getCapabilities().isCas()) {
-      startFailsafeRegistration();
-    } else {
-      log.log(INFO, "Skipping worker registration");
-    }
```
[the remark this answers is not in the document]
werkt: Yes, and fortunately now we have breaking destinations for storage and execute that nobody consuming the Workers hash will see, if there were any external tools that considered this.

src/main/java/build/buildfarm/common/config/Worker.java
werkt: If we stick with the bitfield, I'll put better convenience methods (loaders from the packed bits, final properties for the booleans, and packagers to recreate the packed value).

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -942,7 +942,7 @@ public void getBlob(
     Set<String> workerSet;
     Set<String> locationSet;
     try {
-      workerSet = backplane.getWorkers();
+      workerSet = backplane.getStorageWorkers();
```
jerrymarino, line 945: IME the definition of [`No workers found.`](https://github.com/bazelbuild/bazel-buildfarm/pull/1328/files#diff-7d364025210bb06a44a529f14efcdf9a0bbbd8e715af51786c4f0ca24310fe52R1097) could use some tweaking. Generally if you don't have a storage worker ( which prior happened to mean exec + storage ) it will give an `UNAVAILABLE` message.

Do you think it now makes more sense to rename this `No storage workers` found? If the CAS goes down it's a problematic situation. The reality of "No execution workers available" is probably undesirable to assert here unless we want that to mean "Hey - you don't have any execute workers"; I wonder if one day `GetCapabilities` could be more dynamic

jerrymarino: A buildfarm could theoretically transition from exec <-> cas only depending what's up-and-running - or all your exec nodes popped

jerrymarino: Also the situation of 0 storage workers makes it probably hard to assert: say you have some external CAS

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1390,7 +1396,10 @@ public boolean canPrequeue() throws IOException {
   @Override
   public BackplaneStatus backplaneStatus() throws IOException {
     BackplaneStatus.Builder builder = BackplaneStatus.newBuilder();
-    builder.addAllActiveWorkers(client.call(jedis -> state.executeAndStorageWorkers.keys(jedis)));
+    builder.addAllActiveWorkers(
+        client.call(
+            jedis ->
+                Sets.union(state.executeWorkers.keys(jedis), state.storageWorkers.keys(jedis))));
```
amishra-u, line 1402: Side effect of this change:
Prometheus started reporting incorrect count of workers for us.
Reason: We register workers with execution and cache capabilities both but we only utilize cas capabillities. While `storageWorkers` map is update on every `fetchAndExpireStorageWorker` call but `deregisterWorker` method is not called from `stop` method (as per my understanding, I may be wrong) and workers stored in `executeWorkers` maps remains in redis forever.

PR #1330, opened 2023-04-26
[refactor] move all the execution wrappers to their own dir

luxe:
- Clean up the main BUILD file
- Clean up miscellaneous files on the root of repo.


src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
```diff
@@ -44,42 +44,43 @@ public class ExecutionWrappers {
...
+  public String linuxSandbox = "/app/build_buildfarm/src/main/execution_wrappers/linux-sandbox";
 
   /**
    * @field asNobody
    * @brief The program to use when running actions as "as-nobody".
    * @details This program is expected to be packaged with the worker image. The linux-sandbox is
    *     also capable of doing what this standalone programs does and may be chosen instead.
    */
-  public String asNobody = "/app/build_buildfarm/as-nobody";
+  public String asNobody = "/app/build_buildfarm/src/main/execution_wrappers/as-nobody";
```
werkt, line 55: I dislike the idea that we're referencing src/main in runtime applications. Is there a way to relocate this as a part of the deployment?

luxe: Maybe as a class_path resource.  I'll try.

werkt:
Conventionally, I think this belongs in /src/execution_wrappers, as it's a non-main component non-language module separation

PR #1331, opened 2023-04-28
Giving names to worker threads

jerrymarino:
Use a schema where the thread is named by the class that owns it and a brief description about what it does, based on what people called the variables.

The context is I've been trailing a bug from time to time and this makes it easier to look at dumps. If we like the format I'd be happy to go through and name other threads in the repo using this convention


luxe (approved):
Anything that helps find bugs!

The CI failure looks spurious.  Run again and it should pass.

jerrymarino:
You got it - thanks for the review!

PR #1333, opened 2023-05-05
fix: FindMissingblobs via backplane introduced in #1310

amishra-uber:
PR #1310 added logic to find missing blobs via backplane. Removing expired digests from `casWorkerMap` is not a frequent operation and is mostly performed by workers during put operations or by the server in case of an error during a read operation.

Update the code to take the intersection with the available workers. This is because any expired or inactive worker is automatically removed from the backplane at most 30 seconds after becoming inactive.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RemoteInputStreamFactory.java
+++ b/src/main/java/build/buildfarm/instance/shard/RemoteInputStreamFactory.java
@@ -33,7 +33,6 @@ import build.buildfarm.instance.Instance;
 import build.buildfarm.instance.shard.ShardInstance.WorkersCallback;
 import com.google.common.base.Throwables;
 import com.google.common.cache.LoadingCache;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -148,7 +147,6 @@ public class RemoteInputStreamFactory implements InputStreamFactory {
         compressor, blobDigest, offset, 60, SECONDS, RequestMetadata.getDefaultInstance());
   }
 
-  @SuppressWarnings("SynchronizationOnLocalVariableOrMethodParameter")
   public InputStream newInput(
       Compressor.Value compressor,
       Digest blobDigest,
@@ -160,14 +158,13 @@ public class RemoteInputStreamFactory implements InputStreamFactory {
     Set<String> remoteWorkers;
     Set<String> locationSet;
     try {
-      Set<String> workers = backplane.getStorageWorkers();
-      if (publicName == null) {
-        remoteWorkers = workers;
-      } else {
-        remoteWorkers = Sets.difference(workers, ImmutableSet.of(publicName)).immutableCopy();
+      remoteWorkers = backplane.getStorageWorkers();
+      if (publicName != null) {
+        remoteWorkers.remove(publicName);
       }
       locationSet =
-          Sets.newHashSet(Sets.intersection(backplane.getBlobLocationSet(blobDigest), workers));
+          Sets.newHashSet(
+              Sets.intersection(backplane.getBlobLocationSet(blobDigest), remoteWorkers));
     } catch (IOException e) {
       throw Status.fromThrowable(e).asRuntimeException();
     }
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -2634,10 +2641,22 @@ public void deregisterWorker(String workerName) {
...
+  }
+
   private boolean inDenyList(RequestMetadata requestMetadata) throws IOException {
     if (!useDenyList) {
       return false;
     }
     return backplane.isBlacklisted(requestMetadata);
   }
+
+  private Set<String> getActiveWorkerCopy() throws IOException {
```
werkt, line 2656: This should probably be switched over in the backplane so that it returns a new HashSet copy each time that can safely be observed - cost should be minimal in creating that new object, and we're doing it by ourselves here. It will make that interface more threadsafe rather than exposing the multiply referenced worker set implementation detail.

werkt (changes requested):
Requesting changes to move copy functionality into backplane implementation, and mark the return/response as ThreadSafe

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
amishra-uber: Thanks for feedback, updated as per request.

luxe (approved):

src/main/java/build/buildfarm/instance/shard/RemoteInputStreamFactory.java
```diff
@@ -164,9 +164,7 @@ public InputStream newInput(
       if (publicName == null) {
         remoteWorkers = workers;
       } else {
-        synchronized (workers) {
```
werkt, line 167: Removal of this means that you can remove the SuppressWarnings above

src/main/java/build/buildfarm/instance/shard/RemoteInputStreamFactory.java
```diff
@@ -164,9 +164,7 @@ public InputStream newInput(
...
-          remoteWorkers = Sets.difference(workers, ImmutableSet.of(publicName)).immutableCopy();
```
werkt, line 168: Is this immutableCopy behavior needed now that we have a real copy?

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -652,20 +652,29 @@ public ListenableFuture<Iterable<Digest>> findMissingBlobs(
...
     // missing blobs.
+
     if (configs.getServer().isFindMissingBlobsViaBackplane()) {
       try {
         Map<Digest, Set<String>> foundBlobs = backplane.getBlobDigestsWorkers(blobDigests);
+        Set<String> workerSet = backplane.getStorageWorkers();
         return immediateFuture(
             StreamSupport.stream(blobDigests.spliterator(), false)
-                .filter(digest -> !foundBlobs.containsKey(digest))
+                .filter(
```
werkt, line 673: EOL Comment here: _// best effort to present digests only missing on active workers_

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -652,20 +652,29 @@ public ListenableFuture<Iterable<Digest>> findMissingBlobs(
...
-    // missing blobs".  Assuming the backplane has an accurate accounting of which blobs are hosted
-    // on which workers, the server could instead check the backplane to determine which blobs are
-    // missing.  This may come with risks if the backplane is not up-to-date with the workers or is
-    // not considered an authortiative source of truth.  However, checking workers directly is not a
+    // missing blobs".
+    // Workers register themselves with the backplane for a 30-second window, and if they fail to
+    // re-register within this time frame, they are automatically removed from the backplane. While
+    // this alternative strategy for finding missing blobs is faster and more cost-effective than
+    // the exhaustive approach of querying each worker to find the digest, it comes with a higher
+    // risk of returning expired workers. This is because the strategy may return workers that
```
werkt, line 660: modify: returning expired workers _despite filtering by active workers below_.

werkt (approved):
Some minor comments, please update, good to land with them

amishra-uber:
> Some minor comments, please update, good to land with them

I don't have write permission, please merge it.

PR #1341, opened 2023-05-12
Use a centos version with newer toolchains

meteorcloudy:
Fixing https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3015#01880e23-5928-4838-8574-95e1e698c94a where https://github.com/bazelbuild/bazel/commit/978cd23e80f202a03e99a87f2089a9a0622b5d5d is probably the culprit.


werkt (approved):
Looks good, thanks for the assist!

PR #1342, opened 2023-05-18
Remove empty/dup digest in findMissingBlobs calculation

amishra-u:
**Issue**: While testing `ensureOutputPresent` flag for buildfarm I found that the `getActionResult` most of the time returns empty digests under key `stdout` and `stderr`. The `getActionResult` returns `not_found` error if `findMissingBlobs` don't return empty list. As empty digest are never stored it will be always present in the missing digest list.
The old implementation (querying each worker) also filters out empty digests. 
**Solution**: Query only nonEmptyDigests for findMissingBlobs api.

**Issue**:  I found another problem in the ensureOutputPresent function. When creating a list of digests from the getActionResult output, duplicates are not being removed. This issue leads to failures when trying to retrieve workers for these digests from the backplane. The ImmutableMap class throws an IllegalArgumentException when duplicate keys are encountered, resulting in an "unknown" error. Code related to this issue in the following locations:
1. List of digests creation: [AbstractServerInstance.java, line 272](https://github.com/bazelbuild/bazel-buildfarm/blob/ad2b823a093a3b1f048e959b5a01f2540e932f28/src/main/java/build/buildfarm/instance/server/AbstractServerInstance.java#L272)
2. Fetching workers from the backplane: [JedisCasWorkerMap.java, line 216](https://github.com/bazelbuild/bazel-buildfarm/blob/ad2b823a093a3b1f048e959b5a01f2540e932f28/src/main/java/build/buildfarm/instance/shard/JedisCasWorkerMap.java#L216)

**Solution**: Make sure list of digests are unique while querying missing blobs.


werkt:
Can you provide a rationale for this? We ignore empty blobs because, by definition, all empty blobs are not missing.

amishra-uber:
> Can you provide a rationale for this? We ignore empty blobs because, by definition, all empty blobs are not missing.

Thanks for looking into. Updated diff summary. I converted this diff to draft as I am still running few experiments.

amishra-u:
@werkt can you  please review it. Thanks in advance.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -666,10 +666,12 @@ public ListenableFuture<Iterable<Digest>> findMissingBlobs(
 
     if (configs.getServer().isFindMissingBlobsViaBackplane()) {
       try {
-        Map<Digest, Set<String>> foundBlobs = backplane.getBlobDigestsWorkers(blobDigests);
+        Set<Digest> distinctDigest = new HashSet<>();
```
werkt, line 669: change this to `uniqueDigests` - collection as plural for consistency

src/test/java/build/buildfarm/instance/shard/ShardInstanceTest.java
```diff
@@ -1069,7 +1069,14 @@ public void findMissingBlobsTest_ViaBackPlane() throws Exception {
                 Digest.newBuilder().setHash("missing2").setSizeBytes(1).build(),
                 Digest.newBuilder().setHash("missing3").setSizeBytes(1).build()));
 
-    Iterable<Digest> allDigests = Iterables.concat(availableDigests, missingDigests);
```
werkt, line 1072: Can you also verify the deduping?

werkt (approved):
couple of quick fixes and looks good.

src/test/java/build/buildfarm/instance/shard/ShardInstanceTest.java
amishra-u: Done

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
amishra-u: done

PR #1376, opened 2023-06-16
Enable custom latency buckets

amishra-u:
Prometheus provides a default latency bucket, and that may not be helpful all use case. This change provide capability to define custom latency buckets.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -11,7 +11,7 @@ server:
   grpcMetrics:
     enabled: true
     provideLatencyHistograms: true
-    latencyBuckets: [0.001, 0.005, 0.01, 0.05, 0.075, 0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0]
+    latencyBuckets: [0.001, 0.01, 0.1, 1, 5, 10, 20, 40, 60, +Infinity]
   maxInboundMessageSizeBytes: 0
   maxInboundMetadataSize: 0
   casWriteTimeout: 3600
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -9,8 +9,9 @@ server:
   actionCacheReadOnly: false
   port: 8980
   grpcMetrics:
-    enabled: false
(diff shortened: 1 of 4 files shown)


luxe:
awesome! I'm not sure why the formatter didn't fix the buildifer issue.  If you can get that green, we can land.

amishra-u:
> awesome! I'm not sure why the formatter didn't fix the buildifer issue. If you can get that green, we can land.

Running `buildifier` fixed it.

amishra-u:
> awesome! I'm not sure why the formatter didn't fix the buildifer issue. If you can get that green, we can land.

Fixed it.

examples/config.yml
```diff
@@ -9,8 +9,9 @@ server:
   actionCacheReadOnly: false
   port: 8980
   grpcMetrics:
-    enabled: false
-    provideLatencyHistograms: false
+    enabled: true
```
shirchen, line 12: any reason to have lines 12:13 and 95:96 be different?

examples/config.yml
```diff
@@ -9,8 +9,9 @@ server:
...
+    provideLatencyHistograms: true
+    latencyBuckets: [0.001, 0.005, 0.01, 0.05, 0.075, 0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0]
```
shirchen, line 14: we still get the infinite bucket, right? i also suggest example to include buckets up to 60s, and then 60+ since 60s is default for bazel.

examples/config.yml
amishra-u: Nothing specific, I was testing locally and updated to ensure both combination works.

examples/config.yml
amishra-u: Updated example config with infinity bucket. Also in the example I have used different latency buckets for server and workers as developers can have different latency bucket for server and worker.

luxe (approved):

PR #1395, opened 2023-06-29
Refactor findMissingBlobs method

### LINKED ISSUE #1393: Refactor findMissingBlobs methods. Followup #1381
We implemented two different approaches for findMissingBlobs, one with backplane and other with worker. These two flows are implemented under single method. This makes hard to maintain. Refactor the findMissingBlobs method and move both approaches  under different method.

amishra-u:
Refactor the findMissingBlobs method and move findMissingBlobs via backplane and findMissingBlobs querying each worker in different private methods.

closes #1393


amishra-u:
@werkt Can you please review it.

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -730,7 +691,7 @@ public ListenableFuture<Iterable<Digest>> findMissingBlobs(
       return immediateFuture(nonEmptyDigests);
     }
 
-    // Search through all of the workers to decide how many CAS blobs are missing.
+    // Search through all the workers to decide which CAS blobs are missing.
```
werkt, line 694: nit, leave the of

werkt (approved):

src/main/java/build/buildfarm/instance/shard/ShardInstance.java
amishra-u: Updated

amishra-u:
@werkt Can you please merge the changes

PR #1408, opened 2023-07-22
Create separate grpc channel for write api and other Improvements

amishra-u:
### Problem
Observed that the buildfarm availability occasionally drops due to timeouts, and even when the traffic decreases significantly, the system fails to recover. Upon investigation, it was discovered that the majority of timeouts are caused by a small group of Workers. Even under light loads, a substantial percentage of read and write requests to these affected workers experience timeouts.

This problem originates from a substantial backlog of pending data in the RECV-Q socket buffers of the Established connections on the Buildfarm worker. These hung-up connections are not appropriately cleared, leading to a high occurrence of timeouts for these connections. Sample connection:

| State  |  Recv-Q  | Send-Q | Local Address:Port  | Peer Address:Port | |
| - | - | - | - | - | -|
| ESTAB  | 2022741   | 0  | 172.17.0.2:8981   | <server_ip>:60296 |  (on worker)|
| ESTAB  | 0   | 531246  | 172.17.0.2:8980   | <worker_ip>:60296 |  (on server) |

In addition, during these incidents, the number of open file descriptors (FDs) on the affected workers also rises. This increase in open FDs is primarily caused by the TCP health check that we have configured for the worker cluster. The TCP health check operates by establishing and closing TCP connections. However, on the affected workers, around 5-10% of these TCP health check connections do not terminate correctly and remain stuck in the CLOSE_WAIT state. Each connection in the CLOSE_WAIT state contributes to the growing count of open file descriptors. (Q: why this happens?)

Analyzed multiple such incidents and found a common pattern that the problem starts occurring at very high write QPS. When there is a surge in write requests, the buildfarm worker is unable to process the data at the rate it is being pushed to the network layer. As a result, some data is left unread in the socket buffer because the application may have closed the call before all the requested data reached the worker. This accumulation of unread data in the socket buffers makes the connection very much unusable.

### Solution
1. Created seperate channel for read and write calls, so any issue with write call has no or minimum impact on the read call and vice-versa. This is inline with [Grpc performance best practice](https://grpc.io/docs/guides/performance/) recommendation to create a separate channel for each area of high load in the application.
2. Made gRPC maxconnection age configurable, previously it was using default value which is set to never expire. By setting a specific max connection age duration, if an issue arises with a connection, it will only persist for the designated period. With this configuration, gRPC will automatically close the connection after the max connection age, prompting the creation of a new connection. Additionally, set the maxConnectionAgeGrace to the max grpcTimeout, ensuring that pending RPCs on the connection are not impacted.

### Other improvements
1. Resolved missing prometheus metrics on read cancellation by invoking `blobObserver.onError` on call cancel This ensures that latency and status metrics for cancelled read calls are logged for buildfarm workers.
2. Fix `io_bytes_read` metrics for buildfarm:server. Previously a 1MB (16 * 64) blob logged as 8.5MB ( (16 * (16+1) / 2) * 64 ). 
3. Specify deadline for worker:read requests. 
4. Added and updated logging at few locations.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -135,6 +135,8 @@ worker:
       executionWrapper:
         path: "/"
         arguments:
+  maxConnectionAge: 3600
+  maxConnectionAgeGrace: 60
 executionWrappers:
   cgroups: /usr/bin/cgexec
   unshare: /usr/bin/unshare
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -639,7 +639,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
             log.log(Level.SEVERE, "error closing input stream on cancel", e);
           } finally {
             if (!blobObserver.isCancelled()) {
-              blobObserver.onError(Status.CANCELLED.withDescription("client cancelled").asRuntimeException());
+              blobObserver.onError(
+                  Status.CANCELLED.withDescription("client cancelled").asRuntimeException());
             }
           }
         });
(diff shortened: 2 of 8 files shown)


amishra-u:
@werkt Can you please take a look, thanks

amishra-u:
Merged most of the changes in other smaller prs. Will create another pr for the leftover changes

PR #1414, opened 2023-08-01
Reduce log levels to make log files more meaningful

80degreeswest:
Currently we log a lot of stuff as info, most of which is trivial. This logging is also mixed with meaningful entries which are necessary for troubleshooting. This PR downgrades most frequent log entries to make the more meaningful entries easier to track.

The expected outcome of this change is to set log level to FINE to produce the same output as prior to this change.


luxe:
haha, FINE, FINER, FINEST

luxe (approved):

PR #1416, opened 2023-08-04
Fix io_bytes_read metrics for buildfarm:server

amishra-u:
Fix io_bytes_read metrics for buildfarm:server. Previously  1MB (16 * 64) blob logged as 8.5MB ( (16 * (16+1) / 2) * 64 ).

The first 64KB chunk of read message was correctly as logged 64KB, but subsequent nth 64KB messages were erroneously logged as n * 64KB. As a result, a 1MB blob sent to the client in 16 chunks was being logged as 8.5MB.

```
1*64 + 2*64 + 3* 64 + ... + 16*64 = 8.5MB
```


werkt (approved):
good catch

PR #1417, opened 2023-08-06
Enable prometheus logging for buildfarm:worker read on client cancel

amishra-u:
### Problem
Currently we don't see any metrics with status `cancelled` for buildfarm:worker read. This issue arises because the [onCancelHandler](https://github.com/bazelbuild/bazel-buildfarm/blob/831572372df54c4a7e8539dfd27c75bd7a4d0dac/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java#L633) does not invoke the `responseObserver.onError` method. As a result, the [reportEndMetrics](https://github.com/grpc-ecosystem/java-grpc-prometheus/blob/master/src/main/java/me/dinowernli/grpc/prometheus/MonitoringServerCall.java)  function, responsible for logging prometheus metrics, is not triggered.

### Solution
Added code to ensure that the responseObserver.onError method is called when the client initiates a cancellation request. Additionally, added a check in the [ServerStreamCallObserver](https://github.com/bazelbuild/bazel-buildfarm/blob/831572372df54c4a7e8539dfd27c75bd7a4d0dac/src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java#L131) implementation to avoid the [call already closed](https://github.com/grpc/grpc-java/blob/4049f89e13f3a4f2ee240511f2b23c10bf917cab/core/src/main/java/io/grpc/internal/ServerCallImpl.java#L212) error, which could occur if the buildfarm:worker has already called onComplete or onError (realistically onComplete) by the time the cancellation request from the client (buildfarm:server) arrives.

### DIFF
--- a/src/main/java/build/buildfarm/instance/stub/StubInstance.java
+++ b/src/main/java/build/buildfarm/instance/stub/StubInstance.java
@@ -595,8 +595,7 @@ public class StubInstance implements Instance {
       ServerCallStreamObserver<ByteString> blobObserver,
       RequestMetadata requestMetadata) {
     throwIfStopped();
-    bsStub
-        .get()
+    deadlined(bsStub)
         .withInterceptors(attachMetadataInterceptor(requestMetadata))
         .read(
             ReadRequest.newBuilder()
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -89,6 +89,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.google.protobuf.ByteString;
 import io.grpc.Deadline;
+import io.grpc.Status;
 import io.grpc.StatusException;
 import io.grpc.StatusRuntimeException;
 import io.grpc.stub.ServerCallStreamObserver;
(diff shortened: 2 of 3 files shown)


amishra-u:
@werkt Can you please take a look. This is minor update with no functionality change.

src/main/java/build/buildfarm/instance/stub/StubInstance.java
```diff
@@ -595,8 +595,7 @@ public void getBlob(
       ServerCallStreamObserver<ByteString> blobObserver,
       RequestMetadata requestMetadata) {
     throwIfStopped();
-    bsStub
-        .get()
+    deadlined(bsStub)
```
werkt, line 598: This change isn't in line with the description of the commit.

werkt (changes requested):
This is not a change I'm interested in making. You're accommodating an interceptor for its behavior by changing the usage, inducing a registration with a loop back through an already cancelled handle, and the affected domain for this is the single getBlob, ignoring all the other calls where this takes place.

This is a problem with the prometheus interceptor, I recommend you make the change there - it should handle cancels by presenting metrics.

amishra-u:
> This is not a change I'm interested in making. You're accommodating an interceptor for its behavior by changing the usage, inducing a registration with a loop back through an already cancelled handle, and the affected domain for this is the single getBlob, ignoring all the other calls where this takes place.
> 
> This is a problem with the prometheus interceptor, I recommend you make the change there - it should handle cancels by presenting metrics.

Thanks for feedback, will make the required changes.

PR #1418, opened 2023-08-08
Add download rate metrics for buildfarm:worker

amishra-u:
Download rate metrics for worker was missing. Added metrics for worker download rate.

### DIFF
--- a/src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java
+++ b/src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java
@@ -69,7 +69,7 @@ import lombok.extern.java.Log;
 
 @Log
 public class ShardWorkerInstance extends AbstractServerInstance {
-  private final Counter ioMetric =
+  private static final Counter IO_METRIC =
       Counter.build().name("io_bytes_read").help("Read I/O (bytes)").register();
 
   private final Backplane backplane;
@@ -136,7 +136,7 @@ public class ShardWorkerInstance extends AbstractServerInstance {
           @Override
           public void onNext(ByteString data) {
             blobObserver.onNext(data);
-            ioMetric.inc(data.size());
+            IO_METRIC.inc(data.size());
           }
 
           void removeBlobLocation() {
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java
```diff
@@ -68,6 +69,9 @@
 
 @Log
 public class ShardWorkerInstance extends AbstractServerInstance {
+  private static final Counter ioMetric =
```
werkt, line 72: Not `static`

werkt (approved):
approved pending nit fix

src/main/java/build/buildfarm/worker/shard/ShardWorkerInstance.java
amishra-u: Blindly copied from ShardInstance :( removed

amishra-u: Counter's [register function](https://github.com/prometheus/client_java/blob/main/simpleclient/src/main/java/io/prometheus/client/SimpleCollector.java#L253) appends (registers) metrics to a static instance of the [CollectorRegistry](https://github.com/prometheus/client_java/blob/main/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L28).

After I modified the ioMetric variable to be non-static, the unit tests in ShardWorkerInstanceTest began to fail. This is because each test generates a fresh ShardWorkerInstance instance, and with every instance creation, it tries to register the same metrics with the CollectorRegistry again and fails with [already register error](https://github.com/prometheus/client_java/blob/main/simpleclient/src/main/java/io/prometheus/client/CollectorRegistry.java#L55).

Although buildfarm:worker at runtime only creates single instance of `ShardWorkerInstance` and even making the ioMetric non-static will work but we have not it to be singleton. Also `ioMetric` need to be registered only once not per instance.

amishra-u: @werkt Can you please merge it.

werkt: Alright. We will need to get rid of the statics overall elsewhere.

PR #1419, opened 2023-08-08
Upgrade opentelemetry javaagent to 1.28

tokongs:
We've faced some issues relating to the older version of the opentelemetry javaagent. Maybe you have a reason to run the old one. Just close this PR if that is the case.

I believe there is a bug that was fixed in 1.12 that causes a GRPC context to expire early and interrupt a thread that shouldn't be interrupted. More specifically I think it's related to this https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4169


werkt:
Thanks for the followup!

PR #1424, opened 2023-08-20
Separate channel for write api

amishra-u:
The grpc channel has limit on the number of concurrent streams and currently it is not configurable. As per the [Grpc performance best practice](https://grpc.io/docs/guides/performance/), it is recommended to create separate channel for each area of high load in the application.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/WorkerStubs.java
+++ b/src/main/java/build/buildfarm/instance/shard/WorkerStubs.java
@@ -57,6 +57,7 @@ public final class WorkerStubs {
         worker,
         digestUtil,
         createChannel(worker),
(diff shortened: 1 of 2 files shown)


amishra-u:

werkt (approved):
OK! Finally understand what's really going on here, and with https://github.com/grpc/grpc/issues/21386 referenced as a path to a solution that will open up new connections instead of queuing, this is a fine workaround for now. I would even have preferred this to be a configurable option (# of read + write channels multiplexed/RR'd through), since it will nominally vary with performance (and we don't know how 'many' streams per connection there really is a limit for).

https://github.com/bazelbuild/bazel-buildfarm/issues/1430 will clean up behind this.

PR #1425, opened 2023-08-21
Add request metadata interceptor to Worker

amishra-u:
The RequestMetadata is transferred from BuildfarmServer to BuildfarmWorker;  however, it is not set within the Grpc context. Added interceptor `ServerHeadersInterceptor` which will set the requestMetadata in the GRPC context for BuildfarmWorker.


amishra-u:

werkt (approved):
Nice and tight.

PR #1436, opened 2023-09-07
feat(redis): support rediss:// URIs for Redis-SSL

jasonschroeder-sfdc:
Support "rediss://" (extra 's', like https) to indicate Redis-SSL connections.

For Redis client identifier, we use the `identifier` passed to RedisShardBackplane.

Closes: https://github.com/bazelbuild/bazel-buildfarm/issues/1433


src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -186,12 +198,13 @@ private static Supplier<JedisCluster> createJedisClusterFactory(
             /* soTimeout=*/ Integer.max(2000, timeout),
             Integer.max(5, maxAttempts),
             password,
-            poolConfig);
+            identifier,
+            poolConfig,
+            false);
```
werkt, line 203: nit: please incorporate the following
```suggestion
            /* ssl=*/ false);
```

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -186,12 +198,13 @@ private static Supplier<JedisCluster> createJedisClusterFactory(
...
   }
   /**
    * @brief Create a jedis pool config.
    * @details Use configuration to build the appropriate jedis pool configuration.
-   * @param config Configuration for connecting to a redis cluster server.
```
werkt, line 194: nice

werkt (approved):
Will leave this up for a day, see if you can throw in the nit

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
jasonschroeder-sfdc: Fixed via 169bea9ac46ac69f699377a6475794e9d64b54e4 🚀

PR #1455, opened 2023-09-24
feat: Implement CAS lease extension

### LINKED ISSUE #1428: Fix Eviction Strategy for CAS Entry with findMissingBlobsViaBackplane
### Current Behavior
The decision regarding the access order of CAS entries for cache eviction strategy in BuildfarmWorker relies on the BuildfarmWorker:fmb call. However, when the `findMissingBlobsViaBackplane` flag is set, BuildfarmServer bypasses the BuildfarmWorker:fmb API and relies on the backplane instead. As a result our cache eviction strategy shifts from LRU to FIFO and without the fmb call, BuildfarmWorker treats these cas entries as if they were never accessed.

Currently, CAS entries are not removed upon expiration; they are only removed during write operations, if no space left based on the CAS size configuration. Also, if we start evicting cas entries upon expiration, we would need to extend the cas_expiration time with every access to a cas entry.

### Expected Behavior
1. Regardless of the internal implementation, whenever the client (Bazel) makes an fmb call for a cas entry, it should be recorded as an access. 
2. CAS entries should be evicted once they reach their expiration time. This should replace the current logic of evicting the minimum necessary entries during write calls to create space for new CAS entries.
3. The expiration time for cas entry should be updated with every access.

amishra-u:
### Problem
1. Enabling the findMissingBlobsViaBackplane flag in BuildfarmServer eliminates the need for the BuildfarmWorker's fmb API call. This BuildfarmWorker:fmb call was also responsible for tracking CAS entry access. As result, our CAS cache eviction strategy shifted from LRU to FIFO.
2. When the findMissingBlobsViaBackplane flag is enabled, the buildfarm relies on the backplane as the definitive source for CAS availability. Since we don't update CAS expiry on each access, the backplane will independently expire CAS entries based on the specified cas_expire duration, even if they are actively being read.

### Solution
Updated bfServer:fmb call to perform non-blocking fmb calls to workers, allowing these workers to record access for the relevant CAS entries.

Extended expiry duration for available CAS entries in the backplane on each fmb call.

With these changes, we can utilize Bazel's `experimental_remote_cache_lease_extension` and `experimental_remote_cache_ttl` flags for incremental builds.

Closes #1428


src/main/java/build/buildfarm/instance/shard/ShardInstance.java
```diff
@@ -719,23 +727,33 @@ private ListenableFuture<Iterable<Digest>> findMissingBlobsQueryingEachWorker(
...
-                        .isEmpty();
+                    return Pair.of(
+                        digest,
+                        filterAndAdjustWorkersForDigest(
+                            digest, initialWorkers, workerSet, workersStartTime));
                   })
+              .collect(Collectors.toMap(Pair::getKey, Pair::getValue));
+
+      leaseExtensionService.execute(
+          () -> extendLeaseForDigests(digestAndWorkersMap, requestMetadata));
```
werkt, line 751: The use of a separate executor service for this should not be necessary. Use a cascading request and direct executor to do this with a future chain. You should also create a single grpc context detached from the current to handle all of those requests.

werkt: (not necessary because you can throttle # of reqs with non-blocking calls through a semaphore)

amishra-u: With directExecutor(), I need to make sure that `extendLease` method is called after `responseObserver.onCompleted()`.

I can think of one approach pass a `SettableFuture` and add listener to extendLease. After `responseObserver.onCompleted()` invocation mark `SettableFuture` complete. Are you ok with this, or do you have anything else in mind

```
instance.findMissingBlobs(digestList, requestMetadata, onCompleteFuture /* new argument*/)

....
repsonseObserver.onCompleted()
onCompleteFuture.set(..)
```

In findMissingBlobs

```
ListenableFuture<Iterables>  findMissingBlobs(...) {
....
onCompleteFuture.addListener(() -> extendLease(..), directExecutor())
}
```

amishra-u: Incorporated feedback, can you please check.

amishra-u: @werkt can you please take a look.

werkt (approved):
Looks good, thanks!

werkt:
@amishra-u seems like this needs a rebase

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
```diff
@@ -727,24 +729,40 @@ private ListenableFuture<Iterable<Digest>> findMissingBlobsQueryingEachWorker(
...
           uniqueDigests.stream()
-              .filter( // best effort to present digests only missing on active workers
+              .map(
                   digest -> {
                     Set<String> initialWorkers =
                         foundBlobs.getOrDefault(digest, Collections.emptySet());
-                    return filterAndAdjustWorkersForDigest(
-                            digest, initialWorkers, workerSet, workersStartTime)
-                        .isEmpty();
+                    return Pair.of(
```
werkt, line 745: Map.Entry<Digest, Set<String>> would work just fine here as well, instead of introducing an additional dependency...

werkt (changes requested):

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
amishra-u: Updated

werkt (approved):

PR #1464, opened 2023-09-29
Log write errors with worker address

amishra-u:
For metrics purpose it will be helpful to find number of write timeouts per worker. This will help us figure out which worker is causing issues.


luxe (approved):

werkt (approved):
Hopefully these won't be doubled with the new write logs I added in #1459

PR #1465, opened 2023-10-02
feat: Hot CAS Entries - Implement CAS access metrics recorder

amishra-u:
https://docs.google.com/document/d/12fSpsr_UnOOoeZU0zS4KT0IVQhDoQYUsoIO0f-YA__Q/edit#heading=h.gxowpjqc6bbi

#1443

### DIFF
--- a/src/main/java/build/buildfarm/backplane/Backplane.java
+++ b/src/main/java/build/buildfarm/backplane/Backplane.java
@@ -276,4 +276,20 @@ public interface Backplane {
 
   /** Set expiry time for digests */
   void updateDigestsExpiry(Iterable<Digest> digests) throws IOException;
+
+  /**
+   * Updates the read count for CAS entries based on the provided stream of digest and count.
+   *
+   * @param casReadCountMap A map of Digest and its corresponding read count.
+   * @return A Map containing the updated read counts for the specified CAS entries.
+   */
+  Map<String, Integer> updateCasReadCount(Map<Digest, Integer> casReadCountMap) throws IOException;
+
+  /**
+   * Removes the CAS read count entries from the storage.
+   *
+   * @param digestsToBeRemoved CAS entries for which each read count needs to be removed.
+   * @return total count of cas read count entries removed.
+   */
+  int removeCasReadCountEntries(Iterable<Digest> digestsToBeRemoved) throws IOException;
 }
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -183,6 +183,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
   private final ConcurrentMap<String, Entry> storage;
   private final Consumer<Digest> onPut;
   private final Consumer<Iterable<Digest>> onExpire;
+  private final Consumer<Digest> onReadComplete;
   private final Executor accessRecorder;
   private final ExecutorService expireService;
 
@@ -326,6 +327,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
       String directoriesIndexDbName,
       FixedBufferPool zstdBufferPool,
       Consumer<Digest> onPut,
+      Consumer<Digest> onReadComplete,
       Consumer<Iterable<Digest>> onExpire,
       @Nullable ContentAddressableStorage delegate,
       boolean delegateSkipLoad) {
@@ -338,6 +340,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     this.accessRecorder = accessRecorder;
     this.storage = storage;
     this.onPut = onPut;
+    this.onReadComplete = onReadComplete;
     this.onExpire = onExpire;
     this.delegate = delegate;
     this.delegateSkipLoad = delegateSkipLoad;
@@ -705,6 +708,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
         if (len < 0) {
           in.close();
           blobObserver.onCompleted();
+          onReadComplete.accept(digest);
         }
       }
     }
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -57,6 +57,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Suppliers;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.MultimapBuilder;
@@ -928,6 +929,25 @@ public class RedisShardBackplane implements Backplane {
   public Map<Digest, Set<String>> getBlobDigestsWorkers(Iterable<Digest> blobDigests)
       throws IOException {
     return state.casWorkerMap.getMap(client, blobDigests);
+  }
+
+  @Override
+  public Map<String, Integer> updateCasReadCount(Map<Digest, Integer> casReadCountMap)
+      throws IOException {
+    return client.call(
+        jedis ->
+            state.casReadCount.incrementMembersScore(
+                jedis,
+                casReadCountMap.entrySet().stream()
+                    .collect(Collectors.toMap(e -> e.getKey().getHash(), Map.Entry::getValue))));
+  }
+
+  @Override
+  public int removeCasReadCountEntries(Iterable<Digest> digestsToBeRemoved) throws IOException {
+    return client.call(
+        jedis ->
+            state.casReadCount.removeMembers(
+                jedis, Iterables.transform(digestsToBeRemoved, Digest::getHash)));
   }
 
   public static WorkerChange parseWorkerChange(String workerChangeJson)
--- a/src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
+++ b/src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
@@ -1,0 +1,233 @@
+package build.buildfarm.metrics.cas;
+
+import static java.lang.String.format;
+
+import build.bazel.remote.execution.v2.Digest;
+import build.buildfarm.backplane.Backplane;
+import build.buildfarm.common.DigestUtil;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Stopwatch;
+import com.google.common.base.Supplier;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.logging.Level;
+import lombok.extern.java.Log;
+
+/**
+ * This class records read and write metrics, updating the information to central storage at regular
+ * intervals. The collected data can be utilized to determine the read count of any CAS entry, query
+ * the top "n%" of frequently accessed CAS entries, and rank CAS entries based on their read
+ * frequency, among other uses.
+ */
+@Log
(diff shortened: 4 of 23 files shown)


src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,184 @@
...
+  private long getInitialDelayInMillis() {
+    // Ensure synchronization among all workers, regardless of their start time.
+    return casEntryReadCountUpdateInterval.toMillis()
+        - (Instant.now().toEpochMilli() % casEntryReadCountUpdateInterval.toMillis());
+  }
+
+  private void updateReadCount() {
+    Stopwatch stopwatch = Stopwatch.createStarted();
+    Map<Digest, AtomicInteger> lastIntervalReadCount;
+    lock.writeLock().lock(); // Ensure all threads writes to new interval.
```
werkt, line 147: Use try-with-resources for these blocks

werkt (changes requested):
Please incorporate usage of this new class into this PR - separating them leaves this as an orphaned commit with no obvious introduction of behavior.

amishra-u:
> Please incorporate usage of this new class into this PR - separating them leaves this as an orphaned commit with no obvious introduction of behavior.

Sure, will merge all the PRs. There may be too many files to review :(

amishra-u:
> Please incorporate usage of this new class into this PR - separating them leaves this as an orphaned commit with no obvious introduction of behavior.

@werkt Merge all commits into one. Please take a look.

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
amishra-u: The ReadWriteLock doesn't implements `AutoCloseable`. so moved the closeable logic into separate method. 
i.e `runWithLock`.

src/main/java/build/buildfarm/backplane/Backplane.java
```diff
@@ -281,4 +282,21 @@ boolean pollOperation(QueueEntry queueEntry, ExecutionStage.Value stage, long re
...
+  Map<String, Integer> updateCasReadCount(Stream<Map.Entry<Digest, Integer>> casReadCountStream)
+      throws IOException;
+
+  /**
+   * Removes the CAS read count entries from the storage.
+   *
+   * @param digestsToBeRemoved CAS entries for which each read count needs to be removed.
+   * @return total count of cas read count entries removed.
+   */
+  int removeCasReadCountEntries(Stream<Digest> digestsToBeRemoved) throws IOException;
```
werkt, line 301: Why are you using a Stream as an interface specification? Typically these are Iterables

src/main/java/build/buildfarm/cas/BUILD
```diff
@@ -7,6 +7,7 @@ java_library(
         "@maven//:org_xerial_sqlite_jdbc",
     ],
     deps = [
+        "//src/main/java/build/buildfarm/backplane",
```
werkt, line 10: It is odd to present the backplane interface to the CFC. It only makes use of a small part of it, and other instances where the CFC influences distributed content (i.e. the CAS index) have been accomplished with callbacks, and this bleeds the requirements for a backplane at all into the CAS.

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * This class records read and write metrics, updating the information to central storage at regular
+ * intervals. The collected data can be utilized to determine the read count of any CAS entry, query
+ * the top "n%" of frequently accessed CAS entries, and rank CAS entries based on their read
+ * frequency, among other uses.
+ */
+@Log
+public final class CASAccessMetricsRecorder {
+  private final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
werkt, line 40: This scheduled executor service should be injected.

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
...
+
+  private final Deque<Map<Digest, AtomicInteger>> readIntervalCountQueue;
+
+  private volatile Map<Digest, AtomicInteger> currentIntervalReadCount;
```
werkt, line 44: By what measure is this 'volatile'?

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -3048,6 +3054,15 @@ void commit() throws IOException {
...
         } else if (writeWinner.get()) {
+          RequestMetadata requestMetadata = TracingMetadataUtils.fromCurrentContext();
+          log.log(
+              Level.INFO,
+              format(
+                  "Successfully wrote: %s; metadata: actionId - %s, actionMnemonic - %s, target - %s",
+                  key,
+                  requestMetadata.getActionId(),
+                  requestMetadata.getActionMnemonic(),
+                  requestMetadata.getTargetId()));
```
werkt, line 3065: Why is this a) assumption of grpc context and b) unrelated-to-access metrics log entry in this change?

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
+package build.buildfarm.cas.cfc;
```
werkt, line 1: This does not belong in the cas (let alone cfc) package - a more appropriate location is a peer to the metrics presentation endpoint. Moving it out will alleviate the backplane interface bleed noted above.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -940,6 +941,25 @@ public Map<Digest, Set<String>> getBlobDigestsWorkers(Iterable<Digest> blobDiges
...
+  public Map<String, Integer> updateCasReadCount(
+      Stream<Map.Entry<Digest, Integer>> casReadCountStream) throws IOException {
+    return client.call(
+        jedis ->
+            state.casReadCount.incrementMembersScore(
+                jedis,
+                casReadCountStream.map(
+                    entry ->
+                        new AbstractMap.SimpleEntry<>(
+                            entry.getKey().getHash(), entry.getValue()))));
```
werkt, line 954: There are some Maps transformations that would be suitable in guava for this. The Map is looking like a better candidate for interface parameter typing as well.

werkt (changes requested):

src/main/java/build/buildfarm/backplane/Backplane.java
amishra-u: Changed them to Iterable, as it is more conventional .

Used Stream to leverage its functional programming capabilities and lazy evaluation, as I was performing transformations on each element. These transformations are relatively simple and don't benefit from the features offered by streams.

src/main/java/build/buildfarm/cas/BUILD
amishra-u: Thanks for pointing out. Agreed metrics package is more suitable for this. Moved.

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
amishra-u: Moved under metrics package.

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
amishra-u: Done

src/main/java/build/buildfarm/cas/cfc/CASAccessMetricsRecorder.java
amishra-u: currentIntervalReadCount` is updated - at the end of each interval, it is set to reference a new object. `currentIntervalReadCount` is accessed by all cas read/write threads. Without volatile keyword cas read/write threads may have its stale reference.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
amishra-u: Changed it to traditional forEach. I wasn't able to find any method to transform keys in guava something like `Maps.transformKeys`.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
amishra-u: There is no easy way to reverse map an digest to ActionMenomnic and target pair. I needed this to find targets with indeterminism. Definitely unrelated to cas access metrics, removed will create another PR for this change.

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
...
+  private static void runWithLock(Lock lock, Runnable task) {
+    lock.lock();
+    try {
+      task.run();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @SneakyThrows
```
werkt, line 233: Do not use this annotation. Change the signature to accept a Supplier<T>.

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  /** Stops the scheduler responsible to update read counts periodically. */
+  public void stop() throws InterruptedException {
+    running = false;
+    scheduler.shutdown();
+    if (!scheduler.awaitTermination(10, TimeUnit.SECONDS)) {
+      log.severe("Could not shut down cas access metrics recorder service");
+    }
+    scheduler.shutdownNow();
```
werkt, line 134: This class should not presume ownership of the scheduler it is passed - stop the recorder via this method, and stop the scheduler independently, owned (and maintained) by the caller

werkt (changes requested):
looking better, still needs some changes.

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,242 @@
...
+              log.info(
+                  format(
+                      "Digest %s was not read once after being written in last %s duration",
+                      DigestUtil.toString(digest), casEntryReadCountWindow.toString()));
+            }
+          }
+        });
+    if (digestsToExpire.isEmpty()) {
+      return;
+    }
```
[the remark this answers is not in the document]
amishra-u: Done

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
amishra-u: Done

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
amishra-u: Moved logic to stop the scheduler in Worker class

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Increments the read count by 1 for the given digest.
+   *
+   * @param digest Digest to which the read count is recorded.
+   */
+  public void recordRead(Digest digest) {
+    if (!running) {
+      throw new IllegalStateException("Metrics Recorder is not running");
+    }
```
werkt, line 85: prefer `checkState(running, "Metrics Recorder is not running")`

src/main/java/build/buildfarm/metrics/cas/CASAccessMetricsRecorder.java
```diff
@@ -0,0 +1,234 @@
...
+
+  private final int queueLength;
+
+  private boolean running = false;
+
+  private final Backplane backplane;
+
+  final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+  int intervalNumber = 0; // Assists in unit test.
```
werkt, line 56: I really don't like that we have this counter updated in production that we're only using for unit tests. If you can manage to move this out and keep track of the counters in different ways (Test subclassing with override for the interval operation perhaps), that would be preferred.

werkt (commented):
The recording feels like something we should be leaning on a 3rd party lib for tracking - prometheus and other metrics libs seem more appropriate.
I don't see any way that you're presenting this information collected via an API - I'd like to see that present in this change, as a Backplane method, delivered through a service or message expansion.

PR #1489, opened 2023-10-08
Revert "Use integer ids for Sqlite bidirectional index"

80degreeswest:
This reverts commit f651cdbf7b2203e8cdb5645a1f869589e42f28c3.

This PR causes issues with remove executions where once CAS limit is reached the executions get stuck.


werkt (approved):
Good for revert

PR #1505, opened 2023-10-16
revert: chore: bump rules_jvm_external

luxe:
upstream CI does not seem to pass on Windows without reverting this:
https://github.com/bazelbuild/bazel-buildfarm/commit/9d4c86c8021804aeb5f8723b4faf914963a95386

I'm not sure if something is wrong with the windows VM, or its tied to this dependency upgrade.


werkt:
That has been happening for a while. Not sure if its an issue with new bazel or rules_jvm_external, but a number of commits have been subject to it. I'll retry the build on your other PR and it will likely pass. Definitely not a requirement to revert this particular PR.

PR #1507, opened 2023-10-21
Downgrade log message with high spam rate

80degreeswest:
These log messages have a very high rate and are expected during cache expirations. Downgrade these to FINE to avoid spamming the logs.

### DIFF
--- a/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
+++ b/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
@@ -291,6 +291,12 @@ public class WriteStreamObserver implements StreamObserver<WriteRequest> {
                 requestMetadata.getToolInvocationId(),
                 requestMetadata.getActionId(),
                 name));
+      } else if ("write cancelled".equals(t.getMessage())) {
+        log.log(
+            Level.FINE,
+            format(
+                "write cancelled for %s after %d requests and %d bytes at offset %d",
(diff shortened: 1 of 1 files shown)


werkt (changes requested):
Please switch the exception type emitted in cfc to be detected as canceled and conditionally downgraded

src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
```diff
@@ -291,6 +291,12 @@ private boolean errorResponse(Throwable t) {
                 requestMetadata.getToolInvocationId(),
                 requestMetadata.getActionId(),
                 name));
+      } else if ("write cancelled".equals(t.getMessage())) {
```
werkt, line 294: Prefer that we get a WriteException (`extends IOException`) that specializes to a WriteCancelledException

werkt (changes requested):
Can you please add a test that exceptions indicating cancellations are not logged here?

src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
amishra-u: We were also getting the same error. This pr has fixed the issue for us https://github.com/bazelbuild/bazel-buildfarm/pull/1526.

jasonschroeder-sfdc:
Closing out stale PRs. If this is still valuable, please reopen. Thank you!

PR #1509, opened 2023-10-24
Pin the Java toolchain to `remotejdk_17`

### LINKED ISSUE #1508: Latest worker image fails on startup due to classfile version mismatch
I'm on an M1 Mac running macOS Sonoma and Docker Desktop 4.24.2.

When running `examples/bf-run start`, the worker container stops immediately. Looking at the logs, the output is:
```
Error: LinkageError occurred while loading main class build.buildfarm.worker.shard.Worker
	java.lang.UnsupportedClassVersionError: build/buildfarm/worker/shard/Worker has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
```

If the tag `latest` is replaced with `v2.6.1` the problem goes away. I couldn't find a place in the code where the container is pushed, so I suspect this is done externally (but happy to stand corrected if I missed something).

I suspect the cause is that the process that builds the code that goes in the container has a locally installed Java 17 toolchain, which is picked up by Bazel. If that's the case, the solution should be to make the Java build hermetic with the following:

```diff
diff --git a/.bazelrc b/.bazelrc
index adbb3c68..79b97837 100644
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,9 @@
+build --java_language_version=11
+build --java_runtime_version=remotejdk_11
+
+build --tool_java_language_version=11
+build --tool_java_runtime_version=remotejdk_11
+
 common --enable_platform_specific_config
 
 build:fuse --define=fuse=true
```

Another possibility I considered would be to update the base images to use Java 17, which seem to be already available:

```diff
diff --git a/BUILD b/BUILD
index f552ea9d..688dd0f4 100644
--- a/BUILD
+++ b/BUILD
@@ -148,14 +148,14 @@ oss_audit(
 # Download cgroup-tools so that the worker is able to restrict actions via control groups.
 download_pkgs(
     name = "worker_pkgs",
-    image_tar = "@ubuntu-jammy//image",
+    image_tar = "@ubuntu-mantic//image",
     packages = ["cgroup-tools"],
     tags = ["container"],
 )

 install_pkgs(
     name = "worker_pkgs_image",
-    image_tar = "@ubuntu-jammy//image",
+    image_tar = "@ubuntu-mantic//image",
     installables_tar = ":worker_pkgs.tar",
     installation_cleanup_commands = "rm -rf /var/lib/apt/lists/*",
     output_image_name = "worker_pkgs_image",
diff --git a/images.bzl b/images.bzl
index 9ab0a8b0..c6c76739 100644
--- a/images.bzl
+++ b/images.bzl
@@ -27,19 +27,11 @@ def buildfarm_images():
     )

     container_pull(
-        name = "ubuntu-bionic",
-        digest = "sha256:4bc527c7a288da405f2041928c63d0a6479a120ad63461c2f124c944def54be2",
+        name = "ubuntu-mantic",
+        digest = "sha256:1419bba15470a95242e917b3abcd8981ae36707b99df5ac705e1eee4d920c51c",
         registry = "index.docker.io",
         repository = "bazelbuild/buildfarm-worker-base",
-        tag = "bionic-java11-gcc",
-    )
-
-    container_pull(
-        name = "ubuntu-jammy",
-        digest = "sha256:da847ee259ebe7f00631a2f0146d9add60ff0f94b031a2e522ce94c78b1335c2",
-        registry = "index.docker.io",
-        repository = "bazelbuild/buildfarm-worker-base",
-        tag = "jammy-java11-gcc",
+        tag = "mantic-java17-gcc",
     )

     container_pull(
```

Still, even in this case I think it would probably be advisable to pin the Java toolchain to use Java 17.

Happy to hear back from the maintainers and contribute with a PR based on that. 🙂

stefanobaghino:
Closes #1508

Cleanups:
- remove the unused `ubuntu-bionic` base image
- replace `ubuntu-jammy:jammy-java11-gcc` with `ubuntu-mantic:mantic-java17-gcc`
- replace `amazoncorretto:19` with `ubuntu-mantic:mantic-java17-gcc`
- swap inverted log file names in a file

### DIFF
--- a/.bazelci/run_server_test.sh
+++ b/.bazelci/run_server_test.sh
@@ -8,11 +8,11 @@ bazel build //src/main/java/build/buildfarm:buildfarm-shard-worker
 bazel build //src/main/java/build/buildfarm:buildfarm-server
 
 # Start a single worker
-bazel run //src/main/java/build/buildfarm:buildfarm-shard-worker $(pwd)/examples/config.minimal.yml > server.log 2>&1 &
+bazel run //src/main/java/build/buildfarm:buildfarm-shard-worker $(pwd)/examples/config.minimal.yml > worker.log 2>&1 &
 echo "Started buildfarm-shard-worker..."
 
 # Start a single server
-bazel run //src/main/java/build/buildfarm:buildfarm-server $(pwd)/examples/config.minimal.yml > worker.log 2>&1 &
+bazel run //src/main/java/build/buildfarm:buildfarm-server $(pwd)/examples/config.minimal.yml > server.log 2>&1 &
 echo "Started buildfarm-server..."
 
 echo "Wait for startup to finish..."
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,9 @@
+build --java_language_version=17
+build --java_runtime_version=remotejdk_17
+
+build --tool_java_language_version=17
+build --tool_java_runtime_version=remotejdk_17
+
 common --enable_platform_specific_config
 
 build:fuse --define=fuse=true
--- a/BUILD
+++ b/BUILD
@@ -120,7 +120,7 @@ sh_binary(
 java_image(
     name = "buildfarm-server",
     args = ["/app/build_buildfarm/examples/config.minimal.yml"],
-    base = "@amazon_corretto_java_image_base//image",
+    base = "@ubuntu-mantic//image",
     classpath_resources = [
         "//src/main/java/build/buildfarm:configs",
     ],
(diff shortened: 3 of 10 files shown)


stefanobaghino:
The changes in https://github.com/bazelbuild/bazel-buildfarm/pull/1509/commits/4f65781091002ffb8c8fc8cb8ee8590bb27e4ef6 are necessary to make it work on Windows when reflection is used. https://github.com/bazelbuild/bazel-buildfarm/blob/1f9d01fa70f1ecc419c9fca62f5602e5400d66c6/src/main/java/build/buildfarm/common/io/Utils.java#L235-L286

stefanobaghino:
This time around the CI run seems to have hit a flake or bug of some sort. 😕

werkt:
Windows has been kicking out building errors for stamping lombok, it's not clear why. It succeeds on retries typically, doing that now.

werkt (approved):
Glad you fixed the routing for server/worker log, I had seen that and forgotten to fix it.
Hopefully the build against 17 doesn't cause any issues for byzantine OS installs.

PR #1514, opened 2023-10-30
Rename instance types

luxe:
This is the current hierarchy.  Some of the names make it confusing.
 - `ShardInstance` is only used by Server.  It should mention server so we don't confuse it with worker logic.
 - `AbstractServerInstance` is used by workers - not just servers
 
Screenshot from 2023-10-29 23-23-35

After renames:
Screenshot from 2023-10-29 23-25-30

### DIFF
--- a/admin/main/src/main/resources/proto/buildfarm.proto
+++ b/admin/main/src/main/resources/proto/buildfarm.proto
@@ -520,7 +520,7 @@ message RedisShardBackplaneConfig {
   int32 max_attempts = 33;
 }
 
-message ShardInstanceConfig {
+message ServerInstanceConfig {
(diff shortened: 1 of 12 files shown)


werkt (approved):
This is a fine change, though the Server in this naming was intended to be for a gRPC Server, not necessarily a pure buildfarm Server role.
Also, this has reminded me that we should have further abstraction of the action validator - it deserves to be its own interface + implementation with configuration and maybe expected connectivity. The behavior is a) useful to test in absentia of a working NodeInstance, and b) will reduce the complexity of this class which should amount to a collection of injected objects to access AC/CAS/Queues/etc

PR #1524, opened 2023-11-02
Configured output size operation failure

werkt:
Permit installations to control the failure process for operations which produce outputs larger than the maxEntrySizeBytes. A default value of false retains the existing behavior which appears transient and blacklists the executed action key. When enabled, the action will fail under an invalid violation that indicates user error.


80degreeswest (approved):
Nice.

PR #1526, opened 2023-11-02
Check for context cancelled before responding to error

justinwon777:
When a write fails because the write was already cancelled before due to something like deadline exceeded, we get an unknown error. The exception comes from [here](https://github.com/bazelbuild/bazel-buildfarm/blob/main/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java#L227) and when it gets to `errorResponse()`, it only checks if status code is cancelled. In this case the status code is unknown, so we need to check if context is cancelled to prevent `responseObserver` from being invoked

The code change adds checking if context is cancelled and a unit test testing when the exception has context cancelled.

### DIFF
--- a/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
+++ b/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
@@ -271,7 +271,8 @@ public class WriteStreamObserver implements StreamObserver<WriteRequest> {
 
   private boolean errorResponse(Throwable t) {
     if (exception.compareAndSet(null, t)) {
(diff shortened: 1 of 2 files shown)


werkt (changes requested):
Can we guarantee that the throwable that is passed into errorResponse with the noErrorWhenContextCancelled test is *not* a CANCELLED Status exception, i.e. that it hits the cancellation check?

justinwon777:
Made some changes to make sure `onNext` runs without any exceptions. Also set `write.isComplete()` to `True` so it doesn't execute `handleRequest`. Then after that cancel the context and set the future exception. Hopefully that ensures the exception is coming from the future? I checked with debugger and the future is the only exception being thrown

werkt (approved):

PR #1531, opened 2023-11-07
remove doc about `accept_everything`

Stunkymonkey:
sorry, but I find nothing for this in code.


werkt:
These are now camelized: `acceptEverything` is present in config.yml, as is its partner mislabeled above as `allow_unmatched`, present as `allowUnmatched` in config. I'll go rename them in docs, thanks for the pointer on this inconsistency, though!

PR #1543, opened 2023-11-14
Implement Blake3 hash function support

jerrymarino:
Adds the option to configure Blake3 hash function which was recently added to Bazel.

Outside of just adding the reference implementation of Blake3 + java, we need to make some small affordances which were changed in the byestream calls. Importantly they add the name of the hash function now, which is a nice improvement and could make it easier to run multiple hash implementations on the same instance(s).

Summary
- Add Blake3Hasher to DigestUtil via Blake3Provider
- Pull reference implementation of Blake JNI bridge that works with how we hash stuff in buildfarm
- Pull Bazel's JNI loader minues some deps
- From a build perspective, we hinge some of the build settings to handle macOS and linux per Bazel's shims which work very well


jerrymarino:
@werkt putting this up as a draft; this one gained some recent excitement in the community and I'm excited to try it out on our fork!

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -278,7 +278,22 @@ private static BlobInformation parseBlobInformation(String[] segments, MutableIn
...
       builder.setCompressor(Compressor.Value.IDENTITY);
     }
-    String hash = segments[index.getAndIncrement()];
+
+    String firstPart = segments[index.getAndIncrement()];
+    // The resource syntax has changed in the bytestream
+    // requests.
+    // Prior to blake3 we'd have:
+    // "bytestream://$instance/blobs/$hash/$size",
+    // But now, we have the name ofthe blob
```
jasonschroeder-sfdc, line 287:
```suggestion
    // But now, we have the name of the blob
```

jasonschroeder-sfdc:
Fixes: https://github.com/bazelbuild/bazel-buildfarm/issues/1423

src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -62,13 +83,20 @@ public DigestFunction.Value getDigestFunction() {
...
       return DigestFunction.Value.UNKNOWN;
     }
 
     public static HashFunction forHash(String hexDigest) {
       if (SHA256.isValidHexDigest(hexDigest)) {
         return SHA256;
       }
+      if (BLAKE3.isValidHexDigest(hexDigest)) {
+        return BLAKE3;
+      }
```
amishra-u, line 99: BLAKE3 and SHA256 both have 256 bit digest, With this logic, we can't identify the hashfn by digest.

jasonschroeder-sfdc:
BLAKE3 implemented via https://github.com/bazelbuild/bazel-buildfarm/pull/1697

PR #1546, opened 2023-11-16
Revert "Reorganize DequeueMatchEvaluator (#1537)"

80degreeswest:
This reverts commit 339aa131a971c1e541ede590ea8096cfbde97824.

This commit causes actions to not be picked up by workers and sit in the queue indefinitely, causing the entire cluster to be locked up.


werkt:
We've fixed up the regression with this without a revert. Closing.

PR #1549, opened 2023-11-17
fix: Periodically Refresh Active Storage Workers With StartTime

amishra-u:
### Problem
When the `findMissingBlobsViaBackplane` flag is enabled, every fmb or getac call (when the ensureOutputPresent flag is set) retrieves the Worker start time. Currently, each `getWorkerStartTime` triggers a Redis call. As a result, the Redis shard responsible for storing the worker list becomes hot-shard. At peak request this becomes bottleneck and significantly slows down the BuildfarmServer.

### Solution
There is already a mechanism to periodically refresh the set of active workers every 3 seconds. Made the changes to include addition worker information on periodic call. Instead of depending on Redis for fetching worker start times, we can retrieve the worker start time from the local map which is refreshed periodically.

Unrelated change: Adjusted log level from Info to Finer to record the correction of the worker list for the digest. It was polluting the log.


amishra-u:

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
```diff
@@ -795,7 +795,7 @@ private Set<String> filterAndAdjustWorkersForDigest(
             .immutableCopy();
     if (!workersToBeRemoved.isEmpty()) {
       try {
-        log.log(Level.INFO, format("adjusting locations for the digest %s", digest));
+        log.log(Level.FINE, format("adjusting locations for the digest %s", digest));
```
werkt, line 798: Separate this change if you want it

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -275,7 +275,7 @@ message ShardWorker {
 }
 
 message WorkerChange {
-  message Add {}
+  message Add { google.protobuf.Timestamp firstRegisteredAt = 1; }
```
werkt, line 278: change this to `effectiveAt`

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -144,7 +143,8 @@ public class RedisShardBackplane implements Backplane {
   private @Nullable RedisClient client = null;
 
   private Deadline storageWorkersDeadline = null;
-  private final Set<String> storageWorkerSet = Collections.synchronizedSet(new HashSet<>());
+  private final Map<String, ShardWorker> storageWorkers =
+      Collections.synchronizedMap(new HashMap<>());
```
werkt, line 147: Use a ConcurrentHashMap for this

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
amishra-u: Updated

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
amishra-u: Created new PR
https://github.com/bazelbuild/bazel-buildfarm/pull/1573

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -594,13 +591,21 @@ public void observe(Operation operation) {
   @Override
   public void addWorker(ShardWorker shardWorker) throws IOException {
     String json = JsonFormat.printer().print(shardWorker);
+    Timestamp registrationTime =
+        Timestamp.newBuilder()
+            .setSeconds(shardWorker.getFirstRegisteredAt() / 1000)
+            .setNanos((int) ((shardWorker.getFirstRegisteredAt() % 1000) * 1000000))
```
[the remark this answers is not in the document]
amishra-u: Done

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
amishra-u: Done

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -594,13 +591,23 @@ public void observe(Operation operation) {
...
+    long firstRegisteredSeconds = MILLISECONDS.toSeconds(shardWorker.getFirstRegisteredAt());
+    int firstRegisteredNanos =
+        (int)
+            (MILLISECONDS.toNanos(shardWorker.getFirstRegisteredAt())
+                - SECONDS.toNanos(firstRegisteredSeconds));
+    Timestamp registrationTime =
+        Timestamp.newBuilder()
+            .setSeconds(firstRegisteredSeconds)
+            .setNanos(firstRegisteredNanos)
+            .build();
```
werkt, line 603: Refactor this into a single function that takes a ms count and returns a timestamp. Use the var name `effectiveAt`, or even better, don't retain this in a variable, and put it directly into the WorkerChange.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -699,43 +706,35 @@ public void deregisterWorker(String workerName) throws IOException {
...
-        Set<String> newWorkerSet = client.call(this::fetchAndExpireStorageWorkers);
-        storageWorkerSet.clear();
-        storageWorkerSet.addAll(newWorkerSet);
-      }
-      storageWorkersDeadline = Deadline.after(workerSetMaxAge, SECONDS);
-    }
-    return new HashSet<>(storageWorkerSet);
+  public Set<String> getStorageWorkers() throws IOException {
+    refreshStorageWorkersIfExpired();
+    return new HashSet<>(storageWorkers.keySet());
```
werkt, line 711: It was missing before, but add a comment here that indicates that we're making a *copy* of the set for return, and not providing access to the shared set.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -699,43 +706,35 @@ public void deregisterWorker(String workerName) throws IOException {
...
-      return Collections.emptyMap();
-    }
-    List<String> workerList = client.call(jedis -> state.storageWorkers.mget(jedis, workerNames));
+    refreshStorageWorkersIfExpired();
+    Map<String, Long> workerAndStartTime = new HashMap<>();
+    workerNames.forEach(
+        worker -> {
+          ShardWorker workerInfo = storageWorkers.get(worker);
+          if (workerInfo != null) {
+            workerAndStartTime.put(worker, workerInfo.getFirstRegisteredAt() / 1000L);
```
werkt, line 723: unit conversion missing here.

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -275,7 +275,7 @@ message ShardWorker {
 }
 
 message WorkerChange {
-  message Add {}
+  message Add { google.protobuf.Timestamp effectiveAt = 1; }
```
werkt, line 278: Don't inline this, give the block the same spacing as everything else.

amishra-u: Formatter complains if I don't make it inline. This might be why all types with a single property are also defined inline in this file.
https://github.com/bazelbuild/bazel-buildfarm/blob/2097ef997a0d42a3c0290deb104b3660d697de1d/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto#L80

werkt: Huh, good to know we have linting protection for proto.

werkt (approved):
Looks good, can you address the comment add since you're in there?

amishra-u:
> Looks good, can you address the comment add since you're in there?

Created another PR to update the proto styling. Please merge this one.
https://github.com/bazelbuild/bazel-buildfarm/pull/1594/files

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
amishra-u: Added comment on method.

werkt (approved):

PR #1552, opened 2023-11-21
Update Jedis connection error handling for addWorker

amishra-u:
### Problem
In `addWorker` method exception handling logic was implemented as grpc standard. But Jedis library is not built on top of GRPCm, due to which any connection issue with redis goes unhandled and thread responsible to register worker periodically terminates and the corresponding worker is removed from the list of active workers.
Usually this happens when we restart, upgrade, scale up or scale down the redis cluster. Or redis cluster is overloaded.

### Solution
Updated exception handling.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisClient.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisClient.java
@@ -24,6 +24,7 @@ import java.net.SocketTimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import redis.clients.jedis.JedisCluster;
+import redis.clients.jedis.exceptions.JedisClusterOperationException;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.jedis.exceptions.JedisException;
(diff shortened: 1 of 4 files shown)


werkt (changes requested):
The entire point of addWorker is to encapsulate jedis behaviors. If you're finding an exception that is slipping through the existing exception wrappers inside both the backplane and the client, add that detail there. Do not expose the worker to the fact that the backplane is using jedis, that is an implementation detail.

amishra-u:
> The entire point of addWorker is to encapsulate jedis behaviors. If you're finding an exception that is slipping through the existing exception wrappers inside both the backplane and the client, add that detail there. Do not expose the worker to the fact that the backplane is using jedis, that is an implementation detail.

Thanks for feedback, moved the logic to RedisClient.

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -131,7 +132,7 @@ public <T> T call(JedisContext<T> withJedis) throws IOException {
       // we are technically not at RESOURCE_EXHAUSTED, this is a
       // persistent state which can exist long past the error
       throw new IOException(Status.UNAVAILABLE.withCause(e).asRuntimeException());
-    } catch (JedisConnectionException e) {
+    } catch (JedisConnectionException | JedisClusterOperationException e) {
```
werkt, line 135: JedisClusterOperationException is not guaranteed to have suppressions. It is also thrown in contexts that represent invalid arguments. If you're trying to get the UNAVAILABLE throw here for JedisClusterMaxAttemptsException, just catch that exception specifically and manufacture the exception, without any unwrapping.

amishra-u: Updated

src/main/java/build/buildfarm/common/redis/RedisClient.java
```diff
@@ -148,11 +150,15 @@ public <T> T call(JedisContext<T> withJedis) throws IOException {
           status = Status.DEADLINE_EXCEEDED;
         } else if (cause instanceof IOException) {
           throw (IOException) cause;
+        } else if (cause instanceof JedisClusterOperationException) {
+          cause = cause.getSuppressed()[cause.getSuppressed().length - 1];
```
werkt, line 154: This suppression extract is still in place - there are no suppressions for the non-JedisClusterMaxAttemptsException case when throwing.

werkt (approved):

PR #1556, opened 2023-11-30
add github action to package and publish the helm chart as a released artifact

andrewrothstein:
to publish releases with the gh cli please set
> Settings > Actions > General > Workflow Permissions > Read and Write Permissions

### DIFF
--- a/.github/workflows/buildfarm-helm-chart-publish.yml
+++ b/.github/workflows/buildfarm-helm-chart-publish.yml
@@ -1,0 +1,25 @@
+---
+name: Package and Publish Helm Chart
+
+on:
+  push:
+    tags:
+      - '*'
+
+env:
+  GH_TOKEN: ${{ github.token }}
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


werkt (approved):
Looks good I suppose.

andrewrothstein:
> to publish releases with the gh cli please set
> 
> > Settings > Actions > General > Workflow Permissions > Read and Write Permissions

Thanks for the consideration and merge.

Are you able to make this change? The github action that will publish the chart as a github release artifact on the next tag but it needs the authority do so. I'm guessing that authority hasn't been granted since I don't see any github release artifacts published previously.

werkt:
Perms were already set like this before. Not sure how I should further verify the publish.

andrewrothstein:
> Perms were already set like this before. Not sure how I should further verify the publish.

thanks! next tag should package and publish the chart. I did test this on my fork so I do believe it will work as expected. a thousand apologies if I've broken anything.

PR #1282, opened 2023-03-06
Implement local resources for workers

luxe:
This implements local resources as described in this [design doc](https://docs.google.com/document/d/1u0TkmVmdMS53PWR1hgh-a_cj3NmQYE0Favv9aGFfQZs/edit?usp=sharing).

PR #1567, opened 2023-12-13
docs(configuration): update env vars

jasonschroeder-sfdc:
`REDIS_URI` env var was removed in https://github.com/bazelbuild/bazel-buildfarm/pull/1282


luxe:
REDIS_URI should still be possible via:
https://github.com/bazelbuild/bazel-buildfarm/blob/8d6e93fe0798978bff997c78458ac00fc35d0eeb/src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java#L238-L244

If the functionality was removed, it would be by accident and we should make sure REDIS_URI still works.

jasonschroeder-sfdc:
Oops - Thank you @luxe . I'll close this PR 😇

PR #1569, opened 2023-12-13
feat: Redis password from file

jasonschroeder-sfdc:
Support reading the Redis password from:
- the Redis URL (in a format like `redis://username:password@host`, or `redis://:password@host`
- the `redisCredentialFile` from the config YAML. This could be a Kubernetes secret or Hashicorp Vault or whatever. Just a file with a secret inside.


werkt (approved):
Given the prioritization, I assume that there's some change to the documentation that this would merit. Change looks fine overall, I don't think it would break any password-less redis situations inadvertently.

jasonschroeder-sfdc:
Thank you @werkt ! Doc change coming up...

werkt:
Do you want me to wait to merge this for doc changes in this commit, or will you be doing those separately?

jasonschroeder-sfdc:
Please wait - I'll add it to this PR. Thanks for checking!

werkt:
Seems a bit odd that the password file should take precedence over the immediately declared redisPassword field, but OK, as long as it's documented.

werkt (approved):

PR #1571, opened 2023-12-15
Update statefulsets.yaml

werkt:
Request an xfs formatted mount point for worker root.


jasonschroeder-sfdc:
Why XFS? :)

werkt:
Two reasons: Link count, enumerated here: https://bazelbuild.github.io/bazel-buildfarm/docs/architecture/content_addressable_storage/#casfilecache
And the creation of lost+found directories at mount roots for ext4 filesystems, which a non-privileged worker will fail to start due to attempts to remove it.

jasonschroeder-sfdc (approved):

PR #1590, opened 2023-12-21
include (un)zip in images

andrewrothstein:
the stock images are less useful for my usecase without (un)zip


werkt (approved):
Seems fine enough. Some of our other package choices here are dubious for general audiences anyway.

PR #1592, opened 2023-12-22
[CI] skip server/worker audit on mac

luxe:
Avoid python dependency issues on mac CI


werkt (changes requested):
Can you leave a comment referencing ideally a new issue that identifies this?

werkt (approved):

PR #1594, opened 2023-12-26
chore: Update proto file styling

amishra-u:
1. Upgrade `llvm_toolchain` to 16.0.0 and `bazel_toolchain` to 0.9.
2. Updated the proto styling `AllowShortFunctionsOnASingleLine` from `All` to `Empty`


amishra-u:
No idea why suddenly `Unit Tests on :mac: (OpenJDK 11, Xcode)` started failing.

.bazelci/format.sh
```diff
@@ -77,13 +77,13 @@ run_proto_formatter () {
     # This is intended to be done by the CI.
     if [[ "$@" == "--check" ]]
     then
-        find . -name '*.proto' -exec $BAZEL run $CLANG_FORMAT -- -i --dry-run --Werror {} +
+        find ~+ -name '*.proto' -exec $BAZEL run $CLANG_FORMAT -- -i -style='{AllowShortFunctionsOnASingleLine: Empty}' --dry-run --Werror {} +
```
werkt, line 80: why are you changing `.` to `~+` here? I don't want to find all files in someone's home directory...

.bazelci/format.sh
```diff
@@ -77,13 +77,13 @@ run_proto_formatter () {
...
         handle_format_error_check
         return
     fi
 
     # Fixes formatting issues
-    find . -name '*.proto' -exec $BAZEL run $CLANG_FORMAT -- -i {} +
+    find ~+ -name '*.proto' -exec $BAZEL run $CLANG_FORMAT -- -i -style='{AllowShortFunctionsOnASingleLine: Empty}' {} +
```
werkt, line 86: Since you're repeating this, can you please make the style into a file and reference it?

werkt (changes requested):

.bazelci/format.sh
amishra-u: Fixed. 
After update it stopped working with relative path so try to update the find command to return the absolute path.

.bazelci/format.sh
amishra-u: Created `.clang-format` file at root.

.bazelci/format.sh
werkt: After what update? It's nonsensical that `$PWD` would work and `.` would not...

amishra-u: With $PWD find returns absolute path while with `.` it returns relative path. With relative path formatter fails with `No such directory` error.

Build result wtih `.`
https://buildkite.com/bazel/buildfarm-farmer/builds/7583#018ca3c0-ab27-4e02-b8db-cb63802da98a/120-149

Build result wtih $PWD
https://buildkite.com/bazel/buildfarm-farmer/builds/7619#018ccdbc-808f-44f8-b312-d01d2dd43d5f/120-149

werkt: OK, the problem here is that the `bazel run` of clang-format is a bare executable with no inputs. I'm confused as to how it ever worked, since it would always have the working directory of its runfiles. Short of creating a rule that formats with all proto srcs declared as inputs in filegroups, I'm not sure if we'd be able to do better than this.  I'll allow it for now.

werkt (approved):

PR #1604, opened 2024-01-04
fix seating of the shard worker’s extraVolumeMounts

andrewrothstein:
they’re at `.Values.shardWorker.extraVolumeMounts` and not `.Values.extraVolumeMounts`

### DIFF
--- a/kubernetes/helm-charts/buildfarm/values.yaml
+++ b/kubernetes/helm-charts/buildfarm/values.yaml
@@ -140,10 +140,10 @@ shardWorker:
     #   secretName: my-secret
     #   defaultMode: 0600
 
-  extraVolumeMounts: []
-    # - name: customConfig
-    #   mountPath: /mnt/config
-    #   readOnly: true
+  #extraVolumeMounts:
+  # - name: customConfig
+  #   mountPath: /mnt/config
+  #   readOnly: true
   extraEnv:
   - name: JAVABIN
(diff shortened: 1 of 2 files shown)


kubernetes/helm-charts/buildfarm/values.yaml
```diff
@@ -140,10 +140,10 @@ shardWorker:
     #   secretName: my-secret
     #   defaultMode: 0600
 
-  extraVolumeMounts: []
-    # - name: customConfig
-    #   mountPath: /mnt/config
-    #   readOnly: true
+  #extraVolumeMounts:
```
werkt, line 143: Isn't this going to freak out with a now missing field? Not familiar enough with helm templating to know myself

andrewrothstein: nah, the "with" form's block is omitted if referenced value is undefined.
I learned this trick from this helm chart [here](https://github.com/bazelbuild/bazel-buildfarm/blob/main/kubernetes/helm-charts/buildfarm/templates/server/deployment.yaml#L18). `podAnnotations` doesn't have a default [value](https://github.com/bazelbuild/bazel-buildfarm/blob/main/kubernetes/helm-charts/buildfarm/values.yaml) and still the template renders.

werkt:
Resolving left this substantially smaller - @andrewrothstein does this still capture the idea?

PR #1606, opened 2024-01-05
Publish storage worker and execute worker pool size in prometheus

amishra-u:
Publish storage worker and execute worker pool size in prometheus

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/ServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/shard/ServerInstance.java
@@ -210,9 +210,15 @@ public class ServerInstance extends NodeInstance {
   private static final Gauge workerPoolSize =
       Gauge.build().name("worker_pool_size").help("Active worker pool size.").register();
   private static final Gauge storageWorkerPoolSize =
-      Gauge.build().name("storage_worker_pool_size").help("Active storage worker pool size.").register();
+      Gauge.build()
+          .name("storage_worker_pool_size")
+          .help("Active storage worker pool size.")
+          .register();
   private static final Gauge executeWorkerPoolSize =
-      Gauge.build().name("execute_worker_pool_size").help("Active execute worker pool size.").register();
+      Gauge.build()
+          .name("execute_worker_pool_size")
+          .help("Active execute worker pool size.")
+          .register();
   private static final Gauge queueSize =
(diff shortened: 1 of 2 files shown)


amishra-u:
Had dependency on #1605, buildfails due to missing newly added field.

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
```diff
@@ -209,6 +209,16 @@ public class ServerInstance extends NodeInstance {
...
   private static final Gauge workerPoolSize =
       Gauge.build().name("worker_pool_size").help("Active worker pool size.").register();
+  private static final Gauge storageWorkerPoolSize =
+      Gauge.build()
+          .name("storage_worker_pool_size")
+          .help("Active storage worker pool size.")
+          .register();
+  private static final Gauge executeWorkerPoolSize =
+      Gauge.build()
+          .name("execute_worker_pool_size")
```
80degreeswest, line 219: Would be useful to add a labelNames("queue_name") here as you can have multiple pools of execute workers. Also can you please update the config docs with new metrics.

80degreeswest (changes requested):

src/main/java/build/buildfarm/instance/shard/ServerInstance.java
amishra-u: This will require changes at multiple places. 
1. Change type of active_workers property in [BackplaneStatus](https://github.com/bazelbuild/bazel-buildfarm/blob/044597e12210c8cbca6b0b71ba0062b1eba0243a/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto#L444). Which will trigger changes in multiple places.
2. Include worker queues in [ShardWorker](https://github.com/bazelbuild/bazel-buildfarm/blob/044597e12210c8cbca6b0b71ba0062b1eba0243a/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto#L271)

Are we ok with these changes?

80degreeswest: This is fine as is for now. There is a different way to get the workers by type (if running in k8s) with something like this:
`
count(completed_operations_total) by (job)
`

where job is a label assigned to a kubernetes deployment.

80degreeswest (approved):
Thanks for updating the doc!

PR #1612, opened 2024-01-18
upgrade container image to ubuntu:jammy, python3, add default-jdk

andrewrothstein:


Dockerfile
```diff
@@ -1,14 +1,16 @@
...
 RUN apt-get update
-RUN apt-get -y install wget git zip python gcc openjdk-8-jdk g++ redis redis-server
-RUN wget --tries=10 -O get-pip.py https://bootstrap.pypa.io/pip/2.7/get-pip.py
-RUN python2 get-pip.py
-RUN pip install python-dateutil
+RUN apt-get -y install \
+  wget git zip \
+  python3 python3-pip python3-dateutil \
+  gcc g++ \
+  openjdk-8-jdk openjdk-17-jdk \
```
werkt, line 14: Why both -8- and -17-?

PR #1626, opened 2024-02-08
fix(helm-chart): nodeSelector for shard worker

aisbaa:
`shard-worker/statefulsets.yaml` manifest would fails to apply without this fix.

Before:
```
...
             - mountPath: /tmp/worker
               name: buildfarm-shard-worker-data
       node-type: buildfarm-node
       volumes:
...
```

After:
```
...
             - mountPath: /tmp/worker
               name: buildfarm-shard-worker-data
       nodeSelector:
         node-type: buildfarm-node
       volumes:
...
```

### DIFF
--- a/kubernetes/helm-charts/buildfarm/templates/shard-worker/statefulsets.yaml
+++ b/kubernetes/helm-charts/buildfarm/templates/shard-worker/statefulsets.yaml
@@ -79,7 +79,8 @@ spec:
             {{- end }}
 
       {{- with .Values.shardWorker.nodeSelector }}
-      {{- tpl (toYaml .) $ | nindent 6 -}}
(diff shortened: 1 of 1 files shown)


werkt:
@aisbaa Thanks! In order to accept this code, we require a CLA to be signed by the author - can you do this and we'll happily merge!

jasonschroeder-sfdc (approved):

jasonschroeder-sfdc:
Looks like this fix was already merged in https://github.com/bazelbuild/bazel-buildfarm/pull/1631

aisbaa:
> Looks like this fix was already merged in #1631

Yup, lets close this as duplicate.

PR #1646, opened 2024-02-28
build(bzlmod): update rules_pkg

jasonschroeder-sfdc:


container/test/worker_wrappers.yaml
```diff
@@ -15,7 +15,7 @@ fileExistenceTests:
   - name: "as-nobody wrapper"
     path: '/app/build_buildfarm/as-nobody'
     shouldExist: true
-    permissions: '-r-xr-xr-x'
+    permissions: '-rwxr-xr-x'
```
werkt, line 18: Why is this necessary for this change?

MODULE.bazel
```diff
@@ -46,3 +49,19 @@ use_repo(
...
+# https://github.com/bazelbuild/rules_python/pull/713#issuecomment-1885628496
+# Satisfy running tests in Docker as root.
+bazel_dep(name = "rules_python", version = "0.24.0")
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(
+    configure_coverage_tool = False,
+    ignore_root_user_error = True,
+    python_version = "3.11",
+)
```
werkt, line 62: Not sure I see the connection between adding python here and rules_pkg

jasonschroeder-sfdc: `rules_pkg` 0.10.1 depends on `rules_python`
https://registry.bazel.build/modules/rules_pkg , we're configuring the `ignore_root_user_error` for CI, or, we can invest in running CI as non-root user

container/test/worker_wrappers.yaml
jasonschroeder-sfdc: This just seems to be how `rules_pkg` is delivering this particular file. I can try to explicitly set it to `-r-xr-xr-x`

MODULE.bazel
werkt: Got it, I recall this discussion now.

container/test/worker_wrappers.yaml
werkt: There's really no reason for any file deployed under the buildfarm container root as an application file to have writable permissions, at all. Try to drop this if possible

jasonschroeder-sfdc: Fixed!

PR #1647, opened 2024-02-29
Disable action cache write in default config

80degreeswest:
Ability to write to action cache is a security risk. Set it to read only in default configuration.


80degreeswest:
Abandoning as this config is necessary for remote cache only use case (no execution) and would introduce a breaking change.

PR #1649, opened 2024-02-29
chores: bump google-java-format to 1.20.0

jasonschroeder-sfdc:


werkt (approved):
Fine with nits. Kind of annoyed that the argument descriptor can't be stacked without a space, but w/e

werkt (approved):

PR #1650, opened 2024-02-29
docs(metrics): Update prometheus metrics

jasonschroeder-sfdc:
Some of these don't exist in the code anymore.
Also, labeled some metric types.


werkt (approved):
I don't know whether these were aspirational or what. They don't seem particularly useful anyway.

PR #1651, opened 2024-02-29
build(bzlmod): remove bzlmod lockfile

jasonschroeder-sfdc:
`MODULE.bazel.lock` has platform-specific sections that are difficult for developers on different platforms to update (e.g. darwin/amd64 laptop cannot easily update the linux/amd64 bits)


MODULE.bazel
```diff
@@ -1,3 +1,9 @@
+"""Buildfarm build and test"""
+
+# NOTE: When editing this file, also update the lockfile.
+#   bazel mod deps --lockfile_mode=update
+# See also: https://bazel.build/external/lockfile
+
```
werkt, line 6: Should we then remove this line? Don't really see why we need any modification to this file.

werkt (approved):

PR #1564, opened 2023-12-08
build: start adopting bzlmod

jasonschroeder-sfdc:
Only four bazel dependencies were found in the existing bzlmod registry (https://registry.bazel.build/)

Fixes: https://github.com/bazelbuild/bazel-buildfarm/issues/1479

PR #1659, opened 2024-03-02
Revert "build: start adopting bzlmod (#1564)"

80degreeswest:
This reverts commit b62a4ce976e0d932cec52a4adadec7d573a324a0.

This commit has continuously been causing issues with building buildfarm as an external dependency of other project. Since it's a breaking change recommend reverting it to reverse the breaking change until it can be fixed.
We've been reverting it internally in our CI but it's now diverged due to to other commits to not revert cleanly.


werkt (changes requested):
The end goal of @jasonschroeder-sfdc 's work here is to be able to build as an external dependency with bzlmod and bazel 7. The maintenance of that as a non-breaking behavior is not maintained through commits, or even minor versions - breaking changes are those that change functional/interface behaviors without a direct upgrade path.

Requesting that you hold on this at HEAD.

werkt:
OOB comms with @80degreeswest look like he is unblocked to use bzlmod. Closing, feel free to discuss further.

PR #1662, opened 2024-03-06
build: add cgroup-tools to the ubuntu base images

jasonschroeder-sfdc:
This is part 1 of a two-part PR to remove [`download_pkgs()`](https://github.com/bazelbuild/bazel-buildfarm/blob/cf5f08f1c6a643f6d092beb4747449d9a6d6657f/BUILD#L142C1-L142C14) and [`install_pkgs()`](https://github.com/bazelbuild/bazel-buildfarm/blob/cf5f08f1c6a643f6d092beb4747449d9a6d6657f/BUILD#L149) from `/BUILD.bazel`

### DIFF
--- a/ci/base-worker-image/jammy/Dockerfile
+++ b/ci/base-worker-image/jammy/Dockerfile
@@ -3,4 +3,4 @@
 FROM ubuntu:22.04
 
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


werkt (approved):
Very good looking update here.

PR #1665, opened 2024-03-07
fix(common): do not log redis password on start-up

jasonschroeder-sfdc:
Do not log passwords on start-up.

Fixes: https://github.com/bazelbuild/bazel-buildfarm/issues/1664


werkt (commented):
Clean

werkt (approved):

PR #1673, opened 2024-03-19
chore(MODULE.bazel): bump rules_go

jasonschroeder-sfdc:


werkt (approved):
Passed after a retry of abseil. Looks like flaky tests are a problem all over.

PR #1675, opened 2024-03-19
build: declare container_structure_test as dev dependency

jasonschroeder-sfdc:


werkt (approved):
Have to trust you on this. I think I can guess what dev_dependency means.

PR #1683, opened 2024-03-22
publish Helm chart as OCI artifact to ghcr.io

andrewrothstein:
in lieu of publishing a tgz as a GitHub release artifact, this change publishes (on tag `helm/${semver}`) the Helm chart bundled as an OCI artifact to the GitHub container registry, i.e. ghcr.io. I also updated the installation instructions which are now less gory. This change requires granting public read access to the corresponding buildfarm "package" within the bazel-buildfarm project. It defaults to private.

### DIFF
--- a/.github/workflows/buildfarm-helm-chart-publish.yml
+++ b/.github/workflows/buildfarm-helm-chart-publish.yml
@@ -8,7 +8,7 @@ on:
 
 env:
   CHART_ROOT: "${{ github.workspace }}/kubernetes/helm-charts/buildfarm"
-  GCR_IMAGE: ghcr.io/${{ github.repository_owner }}/buildfarm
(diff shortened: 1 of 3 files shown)


werkt:
Does this require up front work regarding granting permissions? I have no familiarity with configuring anything related to ghcr, packages, and I don't see anything in settings that leads to such.

jasonschroeder-sfdc (approved):

andrewrothstein:
When I tested this in my own gh account the "package" had default visibility of private and needed to be modified by the owner to be publicly downloadable.

werkt (approved):
Hopefully this workflow will result in some publications.

werkt:
I've requested org-level permissions to make packages public, hopefully this will be granted tomorrow.

PR #1687, opened 2024-03-27
fix: simplify RedisHashTags

jasonschroeder-sfdc:
Fixes: #1686

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisHashtags.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisHashtags.java
@@ -14,12 +14,8 @@
 
 package build.buildfarm.common.redis;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
 
 /**
  * @class RedisHashtags
@@ -35,8 +31,7 @@ public class RedisHashtags {
    * @return A valid queue name for one of the internal queues.
    * @note Suggested return identifier: queueName.
    */
-  public static String hashedName(@Nonnull String name, @Nullable String hashtag) {
-    checkNotNull(name, "Name is required");
+  public static String hashedName(String name, String hashtag) {
     if (hashtag != null && !hashtag.isEmpty()) return "{" + hashtag + "}" + name;
     return name;
   }
@@ -48,8 +43,7 @@ public class RedisHashtags {
    * @return A valid keyname without hashtags.
    * @note Suggested return identifier: queueName.
    */
-  public static String unhashedName(@Nonnull String name) {
-    checkNotNull(name, "Name is required");
+  public static String unhashedName(String name) {
     return name.replaceAll("\\{.*?\\}", "");
   }
 
@@ -61,8 +55,7 @@ public class RedisHashtags {
    * @return The existing hashtag name found in the string (brackets are removed).
    * @note Suggested return identifier: hashtag.
    */
-  public static String existingHash(@Nonnull String name) {
-    checkNotNull(name, "Name is required");
+  public static String existingHash(String name) {
     String regex = "\\{.*?\\}";
     Pattern pattern = Pattern.compile(regex);
     Matcher matcher = pattern.matcher(name);
--- a/src/main/java/build/buildfarm/common/redis/RedisHashtags.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisHashtags.java
@@ -14,8 +14,12 @@
 
 package build.buildfarm.common.redis;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /**
  * @class RedisHashtags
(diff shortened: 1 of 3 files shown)


src/main/java/build/buildfarm/common/redis/RedisHashtags.java
```diff
@@ -31,8 +35,10 @@ public class RedisHashtags {
    * @return A valid queue name for one of the internal queues.
    * @note Suggested return identifier: queueName.
    */
-  public static String hashedName(String name, String hashtag) {
-    return "{" + hashtag + "}" + name;
+  public static String hashedName(@Nonnull String name, @Nullable String hashtag) {
```
werkt, line 38: we assume that non-@Nullable fields are not accepted to be null - definitely have some outliers, but I don't want to start going down the route of tagging *everything* one way or the other.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1497,16 +1497,16 @@ public boolean canPrequeue() throws IOException {
...
-    builder.setPrequeue(state.prequeue.status(client.call(jedis -> jedis)));
-    return builder.build();
+    return BackplaneStatus.newBuilder()
+        .addAllActiveExecuteWorkers(executeWorkers)
+        .addAllActiveStorageWorkers(storageWorkers)
+        .addAllActiveWorkers(Sets.union(executeWorkers, storageWorkers))
+        .setDispatchedSize(client.call(jedis -> state.dispatchedOperations.size(jedis)))
+        .setOperationQueue(state.operationQueue.status(client.call(jedis -> jedis)))
+        .setPrequeue(state.prequeue.status(client.call(jedis -> jedis)))
+        .build();
```
werkt, line 1509: Aside from the formatting change, I don't see any material modifications to this. Does BackplaneStatus now do something different?

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1497,16 +1497,16 @@ public boolean canPrequeue() throws IOException {
...
-    builder.setDispatchedSize(client.call(jedis -> state.dispatchedOperations.size(jedis)));
-    builder.setOperationQueue(state.operationQueue.status(client.call(jedis -> jedis)));
-    builder.setPrequeue(state.prequeue.status(client.call(jedis -> jedis)));
-    return builder.build();
+    return BackplaneStatus.newBuilder()
+        .addAllActiveExecuteWorkers(executeWorkers)
+        .addAllActiveStorageWorkers(storageWorkers)
+        .addAllActiveWorkers(Sets.union(executeWorkers, storageWorkers))
+        .setDispatchedSize(client.call(jedis -> state.dispatchedOperations.size(jedis)))
+        .setOperationQueue(state.operationQueue.status(client.call(jedis -> jedis)))
```
werkt, line 1507: Apropos of nothing, but... what? This is yielding the jedis instance, that is returned to the pool - this is asking for a desync failure...
I know it's not related to your change, but I cannot believe this was in there.
This needs to be `client.call(jedis -> state.operationQueue.status(jedis))` and likewise for the following line with prequeue.

werkt (changes requested):
If this changes BackplaneStatus to return names without hashes, then I need to see that differentiated from the metrics presentation, as there's no other way to present the redis data structures used for queues per their hash positioning within the key.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
jasonschroeder-sfdc: OK if I make that change in a separate PR?

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
jasonschroeder-sfdc: No, just style change.

jasonschroeder-sfdc:
> If this changes BackplaneStatus to return names without hashes, then I need to see that differentiated from the metrics presentation, as there's no other way to present the redis data structures used for queues per their hash positioning within the key.

This only seems to affect when the hash is empty string.

The name is mangled here:
https://github.com/bazelbuild/bazel-buildfarm/blob/e1ad3ad22875a2462d19db36727b498d9e738f81/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java#L377-L382

```java
379         setName(RedisHashTags.hashedName("cpu", ""))
```

returns:
```
{}cpu
```
after the change, it returns:
```
cpu
```


**If this isn't the right place for a fix, I can do a `RedisHashTags.unhashedName(String)` near the metrics presentation. I can create a separate PR if you'd like to see that option 😄**

src/main/java/build/buildfarm/common/redis/RedisHashtags.java
jasonschroeder-sfdc: No worries, I can revert all of the `javax.annotations.*` in this class.

src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
jasonschroeder-sfdc: Moved to https://github.com/bazelbuild/bazel-buildfarm/pull/1691

werkt:
> **If this isn't the right place for a fix, I can do a `RedisHashTags.unhashedName(String)` near the metrics presentation. I can create a separate PR if you'd like to see that option 😄**

That's what I'd prefer - mangle this downstream if this is the structure consumed by metrics. There has to be some way to get the queue name out - I'd actually prefer if we could associate each of the queue responses in the status with a name in redis, and then also offer the hashed-template name for the pool (or maybe skip it if we had each constituent).

jasonschroeder-sfdc:
Here's a smaller attempt. https://github.com/bazelbuild/bazel-buildfarm/pull/1693

PR #1691, opened 2024-04-01
fix(RedisShardBackplane): fix backplaneStatus

jasonschroeder-sfdc:
- re-factor to use chained builder pattern
- fix the calls to `client.call(jedis -> ...)`

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
@@ -1501,14 +1501,16 @@ public class RedisShardBackplane implements Backplane {
   public BackplaneStatus backplaneStatus() throws IOException {
     Set<String> executeWorkers = getExecuteWorkers();
     Set<String> storageWorkers = getStorageWorkers();
+    OperationQueueStatus operationQueueStatus =
+        client.call(jedis -> state.operationQueue.status(jedis));
+    QueueStatus prequeueStatus = client.call(jedis -> state.prequeue.status(jedis));
     return BackplaneStatus.newBuilder()
         .addAllActiveExecuteWorkers(executeWorkers)
         .addAllActiveStorageWorkers(storageWorkers)
         .addAllActiveWorkers(Sets.union(executeWorkers, storageWorkers))
         .setDispatchedSize(client.call(jedis -> state.dispatchedOperations.size(jedis)))
-        .setOperationQueue(
-            (OperationQueueStatus) client.call(jedis -> state.operationQueue.status(jedis)))
-        .setPrequeue((QueueStatus) client.call(jedis -> state.prequeue.status(jedis)))
+        .setOperationQueue(operationQueueStatus)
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -1497,16 +1499,17 @@ public boolean canPrequeue() throws IOException {
...
-    return builder.build();
+    return BackplaneStatus.newBuilder()
+        .addAllActiveExecuteWorkers(executeWorkers)
+        .addAllActiveStorageWorkers(storageWorkers)
+        .addAllActiveWorkers(Sets.union(executeWorkers, storageWorkers))
+        .setDispatchedSize(client.call(jedis -> state.dispatchedOperations.size(jedis)))
+        .setOperationQueue(
+            (OperationQueueStatus) client.call(jedis -> state.operationQueue.status(jedis)))
+        .setPrequeue((QueueStatus) client.call(jedis -> state.prequeue.status(jedis)))
+        .build();
```
werkt, line 1512: To nitpick: this requires 2 * N + 1 (dispatched, which we've demonstrated also needs to be sharded) cluster members' connections to complete, and it's sequential. For the rates that this has to respond at to observe a cluster, I believe it should be done pipelined for each cluster member and in a perfect world, concurrently.

Also, why do we require the casts?

jasonschroeder-sfdc: If I am reading you correctly, I should "batch" the `client.call(...)` once, instead of the 3x `client.call(...)? I am not sure where the `2 * N` comes in..

The casts are there as `BackplaneStatus` has both `setOperationQueue(OperationQueueStatus.Builder)` and `setOperationQueue(OperationQueueStatus)` and java is confused.

werkt: prequeue and operationQueue (2 logical queues) have to query N redis primaries to get their individual values for 2 * N constituent 'real' queues.

It does [get confused]? Even when the T return for the client call is derived from the JedisContext callable equivalent? That seems... odd.

werkt: To put it bluntly, I don't think we can use a single newBuilder()...build() sequence to make the pipeline and concurrency change I'm suggesting. It needs to a) build a Jedis pipeline for each primary, b) dispatch each in a thread executor, and construct the value into a Future BackplaneStatus to be proper. Might be too big for this change.

werkt: For how this is implemented now, assign the values to typed variables above the BackplaneStatus builder and that will remove the ambiguity - I don't like the random cast solving the ambiguity problem overall, and would rather have it explicit than try to coerce what I thought would be an obvious type resolution.

werkt (approved):
Certainly looks better.

PR #1693, opened 2024-04-01
fix: remove redis hashtags from prometheus metrics

jasonschroeder-sfdc:
Strip the redis hashtags from the `queue_size` metric

Alternate fix is in https://github.com/bazelbuild/bazel-buildfarm/pull/1687

Fixes: https://github.com/bazelbuild/bazel-buildfarm/issues/1686


werkt (approved):
It does bind us up with the redis namespace for the first time in ServerInstance, but I prefer this static usage to anything else.

PR #1705, opened 2024-04-15
Paging for listOperations

werkt:
Removes deprecated operations finder
name remains unused, but could identify instances

Supported filters:
"" => all known operations
"status=dispatched" => currently dispatched operations
"invocationId=<invocation guid>" => operations under invocation

page_size == 0 => defaults to 50
  must be 0 <= page_size <= 100


src/main/java/build/buildfarm/tools/Hist.java
```diff
@@ -55,8 +55,8 @@ private static void printHistogram(Instance instance) {
       ImmutableList.Builder<Operation> operations = new ImmutableList.Builder<>();
       for (; ; ) {
         try {
-          pageToken = instance.listOperations(1024, pageToken, "", operations);
-        } catch (io.grpc.StatusRuntimeException e) {
+          pageToken = instance.listOperations(1024, pageToken, "", operations::add);
+        } catch (Exception e) {
```
jasonschroeder-sfdc, line 59: Can we have a more narrow `Exception` sub-class here? this looks spooky

src/main/java/build/buildfarm/server/services/OperationsService.java
```diff
@@ -38,22 +38,25 @@ public OperationsService(Instance instance) {
   public void listOperations(
       ListOperationsRequest request, StreamObserver<ListOperationsResponse> responseObserver) {
     int pageSize = request.getPageSize();
-    if (pageSize < 0) {
+    if (pageSize < 0 || pageSize > 100) {
       responseObserver.onError(Status.OUT_OF_RANGE.asException());
       return;
     }
+    if (pageSize == 0) {
+      pageSize = 50;
```
jasonschroeder-sfdc, line 46: Consider refactoring this into a constant

src/main/java/build/buildfarm/server/services/OperationsService.java
```diff
@@ -38,22 +38,25 @@ public OperationsService(Instance instance) {
...
-    String nextPageToken =
-        instance.listOperations(pageSize, request.getPageToken(), request.getFilter(), operations);
-
-    responseObserver.onNext(
-        ListOperationsResponse.newBuilder()
-            .addAllOperations(operations.build())
-            .setNextPageToken(nextPageToken)
-            .build());
-    responseObserver.onCompleted();
+    // TODO make async
```
jasonschroeder-sfdc, line 49: TODO for now or later?

werkt: later. This would need async redis interaction on the instance -> backplane, and that's something to be handled when I try lettuce.

src/main/java/build/buildfarm/server/services/OperationsService.java
werkt: Done

jasonschroeder-sfdc (approved):

PR #1706, opened 2024-04-17
feat: change `operation_status` Prometheus metric labels

jasonschroeder-sfdc:
Prefer to use the GRPC status Code name instead of int.


werkt (approved):
Nice.

PR #1710, opened 2024-04-20
Limit Zstd ByteBuffer allocations with BufferPool

werkt:
Implement a limiting BufferPool for Zstd's decompression sources.    
This may mitigate OOMs which occur as a result of many concurrent    
writes, but does nothing to capitalize on commonly short compressed    
blobs.    
Since there is no control over the buffered read size at the zstd-jni    
level, lower-level wrappers must be written to take advantage of this,    
and short-blob short-circuiting in CASFileCache may be necessary as    
well.


src/main/java/build/buildfarm/cas/BUILD
```diff
@@ -1,6 +1,9 @@
 java_library(
     name = "cas",
-    srcs = glob(["**/*.java"]),
+    srcs = glob(
+        ["**/*.java"],
+        exclude = ["ShardCAS.java"],
```
jasonschroeder-sfdc, line 5: Is `ShardCAS.java` built (and used) elsewhere?

src/main/java/build/buildfarm/common/config/Worker.java
```diff
@@ -40,6 +40,7 @@ public class Worker {
   private ExecutionPolicy[] executionPolicies = {};
   private SandboxSettings sandboxSettings = new SandboxSettings();
   private boolean createSymlinkOutputs = false;
+  private int zstdBufferPoolSize = 2048; /* * 128k == 256MiB */
```
jasonschroeder-sfdc, line 43:
```suggestion
  private int zstdBufferPoolSize = 2048; /* A value of 128k will use up to 256MiB of RAM */
```
or otherwise clarify this comment?

src/test/java/build/buildfarm/cas/cfc/CASFileCacheTest.java
```diff
@@ -1185,6 +1187,7 @@ public void newInputThrowsNoSuchFileExceptionWithoutDelegate() throws Exception
             /* accessRecorder= */ directExecutor(),
             storage,
             /* directoriesIndexDbName= */ ":memory:",
+            /* zstdBufferPool= */ null,
```
jasonschroeder-sfdc, line 1190: Do we have/need a test that actually tests the buffer pool?

jasonschroeder-sfdc (commented):
Looks good to me. Consider a Prometheus metric (Gauge) for the Worker to see how busy that BufferPool is, so that a buildfarm operator has enough data to tune the configuration.

src/test/java/build/buildfarm/cas/cfc/CASFileCacheTest.java
werkt: We need some tests for zstd itself. This is just context for the stream that the CFC creates - tests on input stream contract and tests of zstd content should be good enough.

PR #1712, opened 2024-04-20
build: Annotate with units from Checker Framework

jasonschroeder-sfdc:
This is a proof-of-concept as I wanted to learn more about this framework.

It's a bit verbose so I'm not convinced. I didn't notice any build time degredations.

If you want to pull it and try, just switch one of those `@s(Prefix.micro)` to `@s(Prefix.nano)` and watch the `bazel build //src/...` will fail.

If we do like this, I can try to tidy it up by adding a convenience `@microseconds` or `@us`(not very read-able) Annotations per https://checkerframework.org/manual/#extending-units


src/main/java/build/buildfarm/worker/Utils.java
```diff
@@ -36,4 +40,13 @@ public static FileStatus statIfFound(Path path, boolean followSymlinks, FileStor
       throw new IllegalStateException(e);
     }
   }
+
+  @SuppressWarnings("units")
+  public static @s(Prefix.micro) long stopwatchToMicroseconds(Stopwatch sw) {
+    return sw.elapsed(MICROSECONDS);
+  }
```
jasonschroeder-sfdc, line 47: Convenience method to convert a stopwatch to a `long` annotated as microseconds.

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -321,11 +334,14 @@ private long executePolled(
...
       }
     }
-    return stopwatch.elapsed(MICROSECONDS) - executeUSecs;
+    return stopwatchToMicroseconds(stopwatch) - executeUSecs;
   }
 
   public void run(ResourceLimits limits) {
+    @SuppressWarnings("units")
+    @s(Prefix.micro)
     long stallUSecs = 0;
```
jasonschroeder-sfdc, line 343: Assigning constants needed a warning suppression, including in the Checker Framework examples. https://github.com/typetools/checker-framework/blob/master/docs/examples/units-extension/UnitsExtensionDemo.java

src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -101,7 +104,7 @@ private long runInterruptible(Stopwatch stopwatch, ResourceLimits limits)
       metadata = operationContext.operation.getMetadata().unpack(ExecuteOperationMetadata.class);
     } catch (InvalidProtocolBufferException e) {
       log.log(Level.SEVERE, "invalid execute operation metadata", e);
-      return 0;
+      return ZERO_MICROSECONDS;
```
jasonschroeder-sfdc, line 107: This was rather annoying as Checker Framework refused to return "unit-less" 0 constant.

(arguably, 0 probably doesn't need units as 0 seconds == 0 minutes, but I suppose 0 degrees Celsius is _not_ the same as 0 degrees Fahrenheit)

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -103,13 +105,17 @@ synchronized int removeAndRelease(String operationName, int claims) {
...
+      String operationName,
+      int claims,
+      @s(Prefix.micro) long uSecs,
+      @s(Prefix.micro) long stallUSecs,
+      int exitCode) {
     int slotUsage = removeAndRelease(operationName, claims);
-    executionTime.observe(usecs / 1000.0);
-    executionStallTime.observe(stallUSecs / 1000.0);
+    executionTime.observe(TimeUnit.MILLISECONDS.convert(uSecs, TimeUnit.MICROSECONDS));
+    executionStallTime.observe(TimeUnit.MILLISECONDS.convert(stallUSecs, TimeUnit.MICROSECONDS));
```
jasonschroeder-sfdc, line 115: This wasn't part of the CheckerFramework, but I just didn't trust myself to do unit conversions :) Can revert..

PR #1719, opened 2024-05-01
Dependency bumps - May 2025

jasonschroeder-sfdc:
_Work in progress_


werkt (approved):
Thanks for the chores.

PR #1721, opened 2024-05-02
fix: metrics math was not including seconds portion of Timestamp

jasonschroeder-sfdc:
The metrics math was only looking at the `getNanos()` part of com.google.protobuf.Timestamp . This adds a helper and adds a unit test.


werkt (approved):
Oof. This was bad. I am fine getting this in as is, but if you could pull getExecutionMetadata() *once* for these 4 uses, I think that would be a lot easier read

werkt (approved):

PR #1722, opened 2024-05-02
fix(metrics): metric type for cas_copy_fallback

jasonschroeder-sfdc:
Since this is incremented, a Counter is more appropriate. Also, added to `metrics.md`


werkt (approved):
"It's a *good* change..."

werkt (approved):

PR #1729, opened 2024-05-10
fix: avoid String ==

jasonschroeder-sfdc:
Use `.equals()` when comparing strings.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/Operations.java
+++ b/src/main/java/build/buildfarm/instance/shard/Operations.java
@@ -154,7 +154,7 @@ public class Operations {
 
     // We also store a mapping from invocationID -> operationIDs
     // This is a common lookup that needs to be performant.
-    if (!"".equals(invocationId) && jedis.sadd(invocationId, name) == 1) {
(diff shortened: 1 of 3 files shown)


src/main/java/build/buildfarm/instance/shard/Operations.java
```diff
@@ -154,7 +154,7 @@ public void insert(UnifiedJedis jedis, String invocationId, String name, String
 
     // We also store a mapping from invocationID -> operationIDs
     // This is a common lookup that needs to be performant.
-    if (invocationId != "" && jedis.sadd(invocationId, name) == 1) {
+    if (!"".equals(invocationId) && jedis.sadd(invocationId, name) == 1) {
```
werkt, line 157: I'm not quite ready to use ruby-like behavior here (const.method()), let's change this to `!invocationId.isEmpty()`

jasonschroeder-sfdc, line 157: That assumes `invocationId != null` :)
```suggestion
    if (!invocationId.isEmpty() && jedis.sadd(invocationId, name) == 1) {
```

PR #1735, opened 2024-05-14
fix(common): guard against empty RedisURI

jasonschroeder-sfdc:
If a customer does not use `redisUri:` in the configuration, and instead uses `redisNodes:`, this can throw a NPE.


werkt (approved):
Nice, this happens actually any time there's a syntax error in the config as well, so good to have it cleaned up.

PR #1677, opened 2024-03-21
Switch to jedis-5

werkt:
Fixes #1661 

Calculate slot lookup table once per app instance

The slot key index is invariant for all usage, calculate it once and
avoid a costly lookup for successive calls to RedisSlotToHash.correlate.
Evidenced in slow unit tests, but may improve production runtime as
well. Provide an initial capacity to the slots ArrayList to boot.

Switch jedis dequeue over to milli granularity

Speeds up tests dramatically.

PR #1736, opened 2024-05-14
fix: Memorystore compatibility mode

jasonschroeder-sfdc:
Google Memorystore's `CLUSTER SHARDS` reply is confusing Jedis. Fall back to using `CLUSTER SLOTS` if `CLUSTER SHARDS` throws any `RuntimeException`s.

The bulk of the implementation is resurrected from the implementation pre-https://github.com/bazelbuild/bazel-buildfarm/pull/1677

See also: https://github.com/redis/jedis/issues/3837

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
@@ -44,7 +44,7 @@ import redis.clients.jedis.util.SafeEncoder;
 @Log
 public class RedisNodeHashes {
   private static final Iterable<List<List<Long>>> SINGLETON_NODE_SLOT_RANGES =
-      ImmutableList.of(ImmutableList.of(ImmutableList.of(0l, CLUSTER_HASHSLOTS - 1l)));
+      Collections.singleton(ImmutableList.of(ImmutableList.of(0l, CLUSTER_HASHSLOTS - 1l)));
 
   /**
    * @brief Get a list of evenly distributing hashtags for the provided redis cluster.
@@ -57,7 +57,7 @@ public class RedisNodeHashes {
     if (jedis instanceof JedisCluster) {
       try {
         Iterable<List<List<Long>>> nodeSlotRanges = getNodeSlotRanges(jedis);
-        ImmutableList.Builder hashTags = ImmutableList.builder();
+        ImmutableList.Builder<String> hashTags = ImmutableList.builder();
         for (List<List<Long>> slotRanges : nodeSlotRanges) {
           // we can use any slot that is in range for the node.
           // in this case, we will use the first slot in the first range.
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
```diff
@@ -104,7 +111,15 @@ public static List<String> getEvenlyDistributedHashesWithPrefix(
   private static Iterable<List<List<Long>>> getNodeSlotRanges(UnifiedJedis jedis) {
     if (jedis instanceof JedisCluster) {
       // get slot range information for each shard
-      return transform(getClusterShards((JedisCluster) jedis), ClusterShardInfo::getSlots);
+      try {
+        return transform(getClusterShards((JedisCluster) jedis), ClusterShardInfo::getSlots);
+      } catch (RuntimeException re) {
+        log.log(
+            Level.WARNING,
+            "Unable to get Redis Cluster shards. Fall-back to deprecated implementation",
```
werkt, line 119: nit: "CLUSTER SHARDS failed. Falling back to CLUSTER NODES (deprecated)"

There's enough context in the backtrace to implicate redis.

src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
```diff
@@ -44,7 +44,7 @@
 @Log
 public class RedisNodeHashes {
   private static final Iterable<List<List<Long>>> SINGLETON_NODE_SLOT_RANGES =
-      ImmutableList.of(ImmutableList.of(ImmutableList.of(0l, CLUSTER_HASHSLOTS - 1l)));
+      Collections.singleton(ImmutableList.of(ImmutableList.of(0l, CLUSTER_HASHSLOTS - 1l)));
```
werkt, line 47: That rather makes the next interior ImmutableList.of a useful singleton candidate, no?

src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
```diff
@@ -86,18 +86,18 @@ public static List<String> getEvenlyDistributedHashesWithPrefix(
...
           hashTags.add(RedisSlotToHash.correlateRangesWithPrefix(slotRanges, prefix));
         }
         return hashTags.build();
       } catch (JedisException e) {
-        return ImmutableList.of();
+        return List.of();
       }
     } else {
-      return ImmutableList.of(prefix);
+      return List.of(prefix);
```
werkt, line 100: I don't agree with this (or really the List response - it should be Iterable) - I don't want the response to be a mutable object.

src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
jasonschroeder-sfdc: Curiously, `Collections.singleton(...)` returns a `java.util.Set` of one item! I would not have guessed that.

src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
jasonschroeder-sfdc: No worries. IntelliJ keeps recommending this refactor so I thought I'd try it. I'll just fixup this commit.

jasonschroeder-sfdc:
Google Memorystore says they will fix this "soon":
https://github.com/redis/jedis/issues/3837#issuecomment-2130059005

jasonschroeder-sfdc:
Google has fixed Memorystore, this patch is no longer needed.

PR #1739, opened 2024-05-15
chore: remove un-implemented interface

jasonschroeder-sfdc:
This isn't used anywhere. Can we remove it?


werkt:
There are commented-out usages in Executor.java that I wanted to restore (they were only active for the legacy operationqueue worker). I'd prefer to leave these in as they represent the proper behavior for the operation presented.

jasonschroeder-sfdc:
Thanks for confirming. Closing as working-as-designed...

PR #1740, opened 2024-05-15
ci: add multi-arch worker images

jasonschroeder-sfdc:
Adding `linux/arm64/v8` variants for the worker images.

The GitHub Actions changes were based on https://docs.docker.com/build/ci/github-actions/multi-platform/ , QEMU and Docker Buildx needed for the ARM variant.

Also, bumping `ci-base-worker-image/mantic/Dockerfile` and `ci-base-worker-image/jammy/Dockerfile` (unused) to trigger a github action to run after merging to the default branch.

### DIFF
--- a/ci/base-worker-image/mantic/Dockerfile
+++ b/ci/base-worker-image/mantic/Dockerfile
@@ -1,6 +1,6 @@
 # A minimal container for building a base worker image.
 # Buildfarm public releases are build using this image as a starting point.
-FROM ubuntu:23.04
+FROM ubuntu:mantic-20240427@sha256:565d62d2283a7cc4b3d759d9a97a5bfcebeb341166f9076a4df504f8f106cd54
 
 RUN apt-get update
 RUN apt-get -y install default-jre default-jdk build-essential libfuse2 cgroup-tools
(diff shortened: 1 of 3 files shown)


ci/base-worker-image/mantic/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # A minimal container for building a base worker image.
 # Buildfarm public releases are build using this image as a starting point.
-FROM ubuntu:23.04
+FROM ubuntu:mantic-20240427@sha256:565d62d2283a7cc4b3d759d9a97a5bfcebeb341166f9076a4df504f8f106cd54
```
werkt, line 3: Why is this and the pin above for jammy desirable for this change in particular? I understand the reason to pick a non-moving target, but I'm not sure it's applicable to this PR (at least not in the same commit, without some reference in the description).

jasonschroeder-sfdc: ah sorry, should have documented it! Per the github action, it only builds this image if the `ci/base-worker-image/mantic/Dockerfile` is updated. To trigger a rebuild, I bumped the image. 
I can just sha256-pin 23.04 if you prefer.

jasonschroeder-sfdc:
Closing in favor of https://github.com/buildfarm/buildfarm/pull/1946

PR #1743, opened 2024-05-16
feat: support custom CA for redis TLS

jasonschroeder-sfdc:
This is to support Google MemoryStore in-transit encryption.

Fixes: https://github.com/bazelbuild/bazel-buildfarm/issues/1643


src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -290,4 +310,70 @@ private static Set<HostAndPort> list2Set(String[] nodes) throws ConfigurationExc
...
   }
+
+  /**
+   * @brief Attach ssl factory to builder.
+   * @details Create an SSL context and attach it to the builder.
+   * @param builder
+   * @return A builder with the SSL context attached.
+   * @note Suggested return identifier: builder.
+   */
+  private static DefaultJedisClientConfig.Builder attachSslSocketFactoryIfNeeded(
```
werkt, line 321: can we break this out into a separate RedisSSL.java - it pollutes the namespace of this file substantially and has only static methods to show for it - I'm picturing one road in with the cert file, and one road out with the authentication adapter.

jasonschroeder-sfdc: Sure - shall I place it under `src/main/java/build/buildfarm/common/redis/RedisSSL.java` ?

werkt: Yup, that will work best.

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -127,7 +132,7 @@ private static void deleteNodeExistingKeys(UnifiedJedis node) {
     ScanParams params = new ScanParams();
     params.match("*");
 
-    // get all of the keys for the particular node
+    // get all the keys for the particular node
```
werkt, line 135: an odd choice? in Pittsburgh we typically only omit 'to be', not 'of'

werkt (approved):
Some odd grammatical semantics in there, but looks good. I'll wait until tomorrow to merge.

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
jasonschroeder-sfdc: The LanguageTool plugin bundled with intellij was complaining about wordiness 🙃 https://languagetool.org/insights/post/wordiness/
I'll revert this and squash.

werkt (approved):

PR #1752, opened 2024-05-30
refactor(common): Collections refactor

jasonschroeder-sfdc:
Prefer `Collections.singleton()` and apply more types.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
+++ b/src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
@@ -18,7 +18,6 @@ import static com.google.common.collect.Iterables.transform;
 import static redis.clients.jedis.Protocol.CLUSTER_HASHSLOTS;
 
 import com.google.common.collect.ImmutableList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import redis.clients.jedis.ConnectionPool;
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
```diff
@@ -37,7 +38,7 @@
  */
 public class RedisNodeHashes {
   private static final Iterable<List<List<Long>>> SINGLETON_NODE_SLOT_RANGES =
-      ImmutableList.of(ImmutableList.of(ImmutableList.of(0l, CLUSTER_HASHSLOTS - 1l)));
+      Collections.singleton(ImmutableList.of(ImmutableList.of(0L, CLUSTER_HASHSLOTS - 1L)));
```
werkt, line 41: This can be another layer of singleton, no?

src/main/java/build/buildfarm/common/redis/RedisNodeHashes.java
```diff
@@ -78,7 +79,7 @@ public static List<String> getEvenlyDistributedHashesWithPrefix(
     if (jedis instanceof JedisCluster cluster) {
       Iterable<List<List<Long>>> nodeSlotRanges = getNodeSlotRanges(cluster);
       try {
-        ImmutableList.Builder hashTags = ImmutableList.builder();
+        ImmutableList.Builder<String> hashTags = ImmutableList.builder();
```
werkt, line 82: oof on both of these. Would be great if we had static analysis that told us when we misuse generics like this.

werkt (approved):

PR #1761, opened 2024-06-10
feat: add Google IAM credentials for Redis password

jasonschroeder-sfdc:
This is to support Google Memorystore Redis Cluster to use IAM authentication.
See also: https://cloud.google.com/memorystore/docs/cluster/manage-iam-auth

To provide auth, you can use `GOOGLE_APPLICIATION_CREDENTIALS` env variable to buildfarm containers, or, if you are deployed to Google Cloud, you can use native identity binding.


src/main/java/build/buildfarm/common/redis/GoogleCredentialProvider.java
```diff
@@ -0,0 +1,152 @@
+// Copyright 2024 The Bazel Authors. All rights reserved.
```
werkt, line 1: Buildfarm Authors

werkt (changes requested):
Clever use of a runnable. One nit, then good

werkt (approved):

PR #1772, opened 2024-06-13
Expand Asset Fetch Qualifier support

### LINKED ISSUE #1768: Support http_header: prefix in Fetch Qualifiers
These can be used to reflect Authorizations and other useful bits for the http request that the fetch api will perform on behalf of the client.

werkt:
Support the now-default `bazel.canonical_id` presented by the bazel client intended to identify the url content uniquely - we currently discard this, as our only association is content digest based.
Support http headers specified by client via the http_headers: prefix.

Bump to Java 21 to support Guarded patterns in switch with `when` for cleaner code.

Fixes #1768, #1769

### DIFF
--- a/.bazelci/format.sh
+++ b/.bazelci/format.sh
@@ -5,8 +5,8 @@
 
 FORMAT_JAVA=true
 REMOVE_NEWLINES_AFTER_START_BRACKET=true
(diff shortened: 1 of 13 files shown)


jasonschroeder-sfdc:
Could you please bump the java version in `ij.bazelproject` like https://github.com/bazelbuild/bazel-buildfarm/pull/1760?

jasonschroeder-sfdc (approved):
Awesome!

PR #1780, opened 2024-06-20
Support missing content-length and checksum.sri

### LINKED ISSUE #1779: Fetch service does not properly handle missing content-length
for an HTTP archive like this:
```starlark
http_archive(
    name = "googleapis",
    urls = ["https://github.com/googleapis/googleapis/archive/ca1372c6d7bcb199638ebfdb40d2b2660bab7b88.tar.gz"],
)
```

Github sends a redirect to `https://codeload.github.com/googleapis/googleapis/tar.gz/ca1372c6d7bcb199638ebfdb40d2b2660bab7b88` . The redirect is followed correctly, but the 

```java
    long contentLength = connection.getContentLengthLong();
```
 in [NodeInstance.java](https://github.com/bazelbuild/bazel-buildfarm/blob/4891ffd30b83319e46696ebb22d935a1d4bd37d2/src/main/java/build/buildfarm/instance/server/NodeInstance.java#L636) is now `-1` which prevents the Output from being created and sized.

werkt:
Load fetch url targets in when either content-length is missing or the checksum was not provided.
Will fail for exceedingly large contents and require additional solutions to page contents/stream and refetch.

Fixes #1779

### DIFF
--- a/src/main/java/build/buildfarm/instance/server/NodeInstance.java
+++ b/src/main/java/build/buildfarm/instance/server/NodeInstance.java
@@ -76,6 +76,7 @@ import build.buildfarm.common.TokenizableIterator;
 import build.buildfarm.common.TreeIterator.DirectoryEntry;
 import build.buildfarm.common.Watcher;
 import build.buildfarm.common.Write;
+import build.buildfarm.common.function.IOSupplier;
 import build.buildfarm.common.net.URL;
 import build.buildfarm.common.resources.BlobInformation;
 import build.buildfarm.common.resources.DownloadBlobRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/build/buildfarm/instance/server/NodeInstance.java
```diff
@@ -647,10 +652,33 @@ private static void downloadUrl(
       throw new IOException(message);
     }
 
-    try (InputStream in = connection.getInputStream();
-        OutputStream out = getContentOutputStream.create(contentLength)) {
+    IOSupplier<InputStream> inSupplier;
+    if (expectedHash.isEmpty() || contentLength < 0) {
+      // not great, plenty risky for large objects
```
jasonschroeder-sfdc, line 657: Should there be a cap on this inputStream? like, abort after some value? At least some logging. I'd hate to try to pull a (large) OCI layer through this path 😬

werkt: There's not currently any storage that would permit us to flow through to mitigate heap usage configured on servers.
I'd be fine with this, or another mechanism that cuts it off and decides that we need to redownload for the actual write. More engineering required.

PR #1785, opened 2024-06-26
test(persistent): update assertions

jasonschroeder-sfdc:
Clarified logic and fixed a test near persistent worker filesystem manipulations.

### DIFF
--- a/src/test/java/build/buildfarm/worker/util/WorkerTestUtils.java
+++ b/src/test/java/build/buildfarm/worker/util/WorkerTestUtils.java
@@ -218,7 +218,7 @@ public class WorkerTestUtils {
     List<Path> listedPaths = listFilesRec(root);
     for (Path filePath : listedPaths) {
       assertWithMessage("Path not match prefix of any expected file: " + filePath)
-          .that(expectedFiles.stream().anyMatch(p -> p.startsWith(p)))
+          .that(expectedFiles.stream().anyMatch(p -> p.startsWith(filePath)))
           .isTrue();
     }
     assertThat(listedPaths).containsAtLeastElementsIn(expectedFiles);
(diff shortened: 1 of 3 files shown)


src/test/java/build/buildfarm/worker/util/WorkerTestUtils.java
```diff
@@ -218,7 +218,7 @@ public static void assertFilesExistExactly(Path root, List<Path> expectedFiles)
     List<Path> listedPaths = listFilesRec(root);
     for (Path filePath : listedPaths) {
       assertWithMessage("Path not match prefix of any expected file: " + filePath)
-          .that(expectedFiles.stream().anyMatch(p -> p.startsWith(p)))
+          .that(expectedFiles.stream().anyMatch(p -> p.startsWith(filePath)))
```
werkt, line 221: Definitely a bug here, but what is the nature of the assertion failing?

jasonschroeder-sfdc: Flipping it to `filePath.startsWith(p)` causes more failures:
```
1) testProtoCoordinator(build.buildfarm.worker.persistent.ProtoCoordinatorTest)
Path not match prefix of any expected file: /workRootsDir/work-root_executionName_-286016731/tool_inputs/tools_dir
expected to be true
	at build.buildfarm.worker.util.WorkerTestUtils.assertFilesExistExactly(WorkerTestUtils.java:222)
	at build.buildfarm.worker.persistent.ProtoCoordinatorTest.testProtoCoordinator(ProtoCoordinatorTest.java:115)

```

`p.startsWith(p)` should always be `true`

werkt: I think this wants a reversed logic: where all listed paths in the outputPaths/Files/Directories are prefixes of the paths observed on the filesystem. I see you've done a lot of refactoring here, and assuming this does proper verification from the tests, I'm fine landing as is and mopping up later.

werkt (approved):

PR #1786, opened 2024-06-30
fix: prefer Timestamps.now() to add nanosecond granularity

jasonschroeder-sfdc:
`Timestamps.fromMillis(System.currentTimeMillis())` is missing nanosecond field for the Timestamp data structure. Prefer `Timestamps.now()` to add it.


werkt (approved):
There was probably not a good reason for this.

PR #1791, opened 2024-07-04
1787-Support_multiple_hash_functions

Robinbux:
Feature to support multiple Hash Functions as described [here](https://github.com/bazelbuild/bazel-buildfarm/issues/1787).

Some uncertain parts I expressed in comments now. Once discussed I will remove those and adjust the unit tests before a potential merge .

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -44,14 +44,8 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
+import build.bazel.remote.execution.v2.*;
 import build.bazel.remote.execution.v2.BatchReadBlobsResponse.Response;
-import build.bazel.remote.execution.v2.Compressor;
-import build.bazel.remote.execution.v2.Digest;
-import build.bazel.remote.execution.v2.Directory;
-import build.bazel.remote.execution.v2.DirectoryNode;
-import build.bazel.remote.execution.v2.FileNode;
-import build.bazel.remote.execution.v2.RequestMetadata;
-import build.bazel.remote.execution.v2.SymlinkNode;
 import build.buildfarm.cas.ContentAddressableStorage;
 import build.buildfarm.cas.DigestMismatchException;
 import build.buildfarm.common.BuildfarmExecutors;
@@ -71,6 +65,8 @@ import build.buildfarm.common.io.FeedbackOutputStream;
 import build.buildfarm.common.io.FileStatus;
 import build.buildfarm.common.io.NamedFileKey;
 import build.buildfarm.v1test.BlobWriteKey;
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Throwables;
 import com.google.common.cache.CacheBuilder;
@@ -247,6 +243,13 @@ public abstract class CASFileCache implements ContentAddressableStorage {
   private final transient Entry header = new SentinelEntry();
   private volatile long unreferencedEntryCount = 0;
 
+  private static final int MAX_ENTRIES = 10000;
+  private static final Cache<String, DigestFunction.Value> digest_to_function_map =
+          Caffeine.newBuilder()
+                  .expireAfterWrite(2, TimeUnit.MINUTES)
+                  .maximumSize(MAX_ENTRIES)
+                  .build();
+
   @GuardedBy("this")
   private long removedEntrySize = 0;
 
@@ -375,7 +378,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
 
     String hashComponent = components[0];
 
-    return digestUtil.build(hashComponent, size);
+    DigestUtil corrDigestUtil = DigestUtil.forDigestFunction(digest_to_function_map.getIfPresent(hashComponent));
+    return corrDigestUtil.build(hashComponent, size);
   }
 
   /**
@@ -775,7 +779,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
 
   @Override
   public Write getWrite(
-      Compressor.Value compressor, Digest digest, UUID uuid, RequestMetadata requestMetadata)
+      Compressor.Value compressor, Digest digest, DigestFunction.Value digestFunction,
+      UUID uuid, RequestMetadata requestMetadata)
       throws EntryLimitException {
     if (digest.getSizeBytes() == 0) {
       return new CompleteWrite(0);
@@ -787,6 +792,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
       return writes.get(
           BlobWriteKey.newBuilder()
               .setDigest(digest)
+              .setDigestFunction(digestFunction)
               .setIdentifier(uuid.toString())
               .setCompressor(compressor)
               .build());
@@ -981,6 +987,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
               throws IOException {
             // caller will be the exclusive owner of this write stream. all other requests
             // will block until it is returned via a close.
+            // Map identifier to digest function
+            digest_to_function_map.put(key.getDigest().getHash(), key.getDigestFunction());
             if (closedFuture != null) {
               try {
                 while (!closedFuture.isDone()) {
@@ -2861,6 +2869,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     Path writePath = getPath(key).resolveSibling(writeKey);
     final long committedSize;
     HashingOutputStream hashOut;
+    DigestUtil corrDigestUtil = DigestUtil.forDigestFunction(digest_to_function_map.getIfPresent(key));
     if (!isReset && Files.exists(writePath)) {
       committedSize = Files.size(writePath);
       try (InputStream in = Files.newInputStream(writePath)) {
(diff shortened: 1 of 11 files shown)


src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -71,20 +74,21 @@ public DigestFunction.Value getDigestFunction() {
       return DigestFunction.Value.UNKNOWN;
     }
 
+    // Correct me if I'm wrong but this function doesn't make any sense?
```
Robinbux, line 77: * I am referring to the old function here

src/main/java/build/buildfarm/common/BUILD
```diff
@@ -80,6 +80,20 @@ java_library(
     ],
 )
 
+java_library(
+    name = "digestUtil",
+    srcs = ["DigestUtil.java"],
```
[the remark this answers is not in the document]
Robinbux: Yea it it, but if I add common to `src/main/java/build/buildfarm/common/resources/BUILD` we have a circular dependency problem. So I created a separate instance for it.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -247,6 +243,13 @@ public SettableFuture<Long> load(Digest digest) {
   private final transient Entry header = new SentinelEntry();
   private volatile long unreferencedEntryCount = 0;
 
+  private static final int MAX_ENTRIES = 10000;
+  private static final Cache<String, DigestFunction.Value> digest_to_function_map =
```
Marzi1904, line 247: why we need to introduce this cache?

src/main/java/build/buildfarm/instance/server/NodeInstance.java
```diff
@@ -1906,7 +1898,7 @@ public boolean pollOperation(String operationName, ExecutionStage.Value stage) {
 
   protected CacheCapabilities getCacheCapabilities() {
     return CacheCapabilities.newBuilder()
-        .addDigestFunctions(digestUtil.getDigestFunction())
+        .addAllDigestFunctions(Arrays.asList(DigestFunction.Value.BLAKE3, DigestFunction.Value.SHA256, DigestFunction.Value.SHA1, DigestFunction.Value.MD5))
```
Marzi1904, line 1901: Let's keep them in digestUtil, and use the function to return them

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
Robinbux: Without it this change actually requires a quite huge refactoring of changing many chaining functions, the corresponding interfaces and as a result also all other implementations of those. This seems to be the cleanest solution to me.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -981,6 +987,8 @@ public synchronized FeedbackOutputStream getOutput(
               throws IOException {
             // caller will be the exclusive owner of this write stream. all other requests
```
amishra-u, line 988: misplaced comment

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -981,6 +987,8 @@ public synchronized FeedbackOutputStream getOutput(
...
             // will block until it is returned via a close.
+            // Map identifier to digest function
+            digest_to_function_map.put(key.getDigest().getHash(), key.getDigestFunction());
```
amishra-u, line 991: This seems like a work around to store the hashing algorithm for digests. It also introduces extra overhead to maintain a cache.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -247,6 +243,13 @@ public SettableFuture<Long> load(Digest digest) {
   private final transient Entry header = new SentinelEntry();
   private volatile long unreferencedEntryCount = 0;
 
+  private static final int MAX_ENTRIES = 10000;
+  private static final Cache<String, DigestFunction.Value> digest_to_function_map =
+          Caffeine.newBuilder()
+                  .expireAfterWrite(2, TimeUnit.MINUTES)
```
amishra-u, line 249: What happens if there are more than 10k write requests in 2 mins.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
```diff
@@ -308,6 +309,11 @@ private static BlobInformation parseBlobInformation(String[] segments, MutableIn
     if (DIGEST_FUNCTIONS.containsKey(maybeDigestFunction)) {
       builder.setDigestFunction(DIGEST_FUNCTIONS.get(maybeDigestFunction));
       index.getAndIncrement();
+    } else {
+      // Infer the digest function based on the hash format
+      // Should go here when it's not BLAKE3
+      String hash = maybeDigestFunction;
+      builder.setDigestFunction(DigestUtil.HashFunction.forHash(hash).getDigestFunction());
```
amishra-u, line 316: DigestFunction is always included in the request, this additional handling might not be needed.
There are several other hashing algorithms with the same hash length as BLAKE3. It might be better to remove the HashFunction.forHash method entirely and always rely on the request.

src/main/java/build/buildfarm/instance/server/NodeInstance.java
```diff
@@ -1906,7 +1898,7 @@ public boolean pollOperation(String operationName, ExecutionStage.Value stage) {
 
   protected CacheCapabilities getCacheCapabilities() {
     return CacheCapabilities.newBuilder()
-        .addDigestFunctions(digestUtil.getDigestFunction())
+        .addAllDigestFunctions(DigestUtil.getSupportedDigestFunctions())
```
amishra-u, line 1901: Current change mostly focus on supporting multiple hashing algorithm for write requests but this change will signal client that buildfarm support multiple hashing algorithm for all apis including `ExecutionService.Execute` API.

amishra-u (changes requested):
Lets first target supporting multiple hashing algorithm for write request and accordingly update the pr summary.

amishra-u (commented):
Please make sure these checks passes. 
https://buildkite.com/bazel/buildfarm-farmer/builds/8542

src/main/java/build/buildfarm/common/resources/ResourceParser.java
Robinbux: I don't see it always included in the request. For blake it is, for sha it isn't.
I assume if we want to change that we would need to change the Bazel OSS first

src/main/java/build/buildfarm/instance/server/NodeInstance.java
Robinbux: Then how would we make the change just for write? If get Capabilities doesn't say it accepts SHA and BLAKE, the client will cancel the request if it uses a different algorithm

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
Robinbux: Doesn't matter. This is basically a LRU cache. As long as there are no 10k request (or 2 minutes) between the methods `getOutput` and `putOrReferenceGuarded` it works. But we can also increase the size if you think that is not sufficient.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
Robinbux: Not my comment, but I can remove it 🤷‍♂️

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
amishra-u: You may increase it, but doing so will add extra overhead. Using an additional data structure (accessible by multiple threads) to pass a value from one method to another may seem like a shortcut to avoid further code changes. However, it's better to make the necessary changes rather than introduce a new data structure.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
amishra-u: I thought that hash function is not part of request only for SHA256 as SHA256 is default hashing algorithm for bazel. However, upon checking the Bazel code, I found that it classifies hashing algorithms into two groups: old and new. For the old ones, where the hashing algorithm can be identified using the hash length, the hash function is not part of the request. For the new ones, the hash function is included in the request. The hash detection logic should complement this Bazel method.
https://github.com/bazelbuild/bazel/blob/1dd5c20513b6ff479bf796fe2541bd492c661c94/src/main/java/com/google/devtools/build/lib/remote/ByteStreamUploader.java#L183

src/main/java/build/buildfarm/instance/server/NodeInstance.java
amishra-u: This could be internal patch until support for multiple hash function is enabled for all APIs.

src/main/java/build/buildfarm/common/resources/ResourceParser.java
Robinbux: It is complementing that now no? For new one we extract if it's part of the request, otherwise we determine by length with this function I added:
```java
public static HashFunction forHash(String hexDigest) {
      int hashLength = hexDigest.length();
        return switch (hashLength) {
            case 64 -> // 32 bytes * 2 characters per byte
                    HashFunction.SHA256;
            case 40 -> // 20 bytes * 2 characters per byte
                    HashFunction.SHA1;
            case 32 -> // 16 bytes * 2 characters per byte
                    HashFunction.MD5;
            default -> throw new IllegalArgumentException("hash type unrecognized: " + hexDigest);
        };
```

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
Robinbux: Fair, but it requires quite a large code change then, even more if we don't want to create a new overload method like you mentioned but change the existing interface methods as well as all implementations of those

src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -71,20 +76,21 @@ public DigestFunction.Value getDigestFunction() {
...
+    // It is always calling the same isValidHexDigest function with the same parameters.
+    // So it will always either return SHA256, or throw an exception.
+    // With the new method we still can't distinguish between SHA256 and BLAKE3, but at least it's better...
```
amishra-u, line 82: This doesn't seem correct. Yes it calls the same function but on different instance of DigestFunction, so for each call, the value for `hash.bits` will be different. 

You may leave this code as it is, as it doesn't change the code behavior.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
amishra-u: Yes, there will be small changes at multiple places in the code. Lets try to do it proper way so it can be upstreamed.

werkt:
This is no longer relevant, as we've landed multiple digest functions for over a year now.

PR #1799, opened 2024-08-01
build: change some `build` flags to be `common`

jasonschroeder-sfdc:
so that they apply to tests, too.


werkt:
`build` is a subset of `test` in bazel configs - can you explain what you're adding to `test` here by switching to `common`?

jasonschroeder-sfdc:
Today I learned!
>(For example, test inherits from build.)

PR #1801, opened 2024-08-01
feat(persistentworker): update minimal config, fix worker key

jasonschroeder-sfdc:
- add persistentWorkerKey and persistentWorkerCommand to `config.minimal.yml`
- persistentWorkerCommand

Test with `bazel build --remote_executor=... --experimental_remote_mark_tool_inputs`

### DIFF
--- a/examples/config.minimal.yml
+++ b/examples/config.minimal.yml
@@ -7,5 +7,17 @@ backplane:
           value: "*"
         - name: "max-cores"
           value: "*"
+        - name: "persistentWorkerKey"
+          value: "*"
+        - name: "persistentWorkerCommand"
+          value: "*"
 worker:
   publicName: "localhost:8981"
+  dequeueMatchSettings:
+    allowUnmatched: true
+    properties:
+      - name: "persistentWorkerKey"
+        value: "*"
+      - name: "persistentWorkerCommand"
+        value: "*"
+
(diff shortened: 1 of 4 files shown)


examples/config.minimal.yml
```diff
@@ -7,5 +7,17 @@ backplane:
           value: "*"
         - name: "max-cores"
           value: "*"
+        - name: "persistentWorkerKey"
+          value: "*"
+        - name: "persistentWorkerCommand"
+          value: "*"
```
werkt, line 13: We should switch to having the default minimal queue be:
```
properties:
  - name: "*"
```
Rather than extending with these.

jasonschroeder-sfdc: I'm not sure what you mean. Are you suggesting to replace `config.minimal.yml` with this?

```yaml
backplane:
  redisUri: "redis://localhost:6379"
  queues:
    - name: "cpu"
      properties:
        - name: "*"
worker:
  publicName: "localhost:8981"
  dequeueMatchSettings:
    allowUnmatched: true
```

werkt: Yes. I'm only assuming that this works, but it looks like it should with the current code, and grants exactly the same level of acceptance as your previous change.

werkt (approved):
Let's go!

PR #1803, opened 2024-08-01
fix(helm): fix statefulset port for worker grpc

jasonschroeder-sfdc:
We pass `--public_name=$(POD_IP):8982` as an arg to the worker, so this should be 8982. There's also reference to 8982 in the configmap.

Fixes: #1796


werkt (approved):
Nice fix

PR #1813, opened 2024-08-14
fix(RedisShardBackplane): Don't return execute-only workers in getStorageWorkers

congt:
BuildFarm now registers all workers in RedisBackplane. When an execute-only worker is registered, RedisShardSubscriber will add it to RedisBackplane's storageWorkers. Then it's possible for getStorageWorkers() to reurn execute-only workers.

To fix this bug, We add a field worker_type to the proto message WorkerChange.Add so that RedisShardSubscriber can tell if the worker type matches the expected value.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
@@ -62,19 +62,19 @@ class RedisShardSubscriber extends JedisPubSub {
 
   private final ListMultimap<String, TimedWatchFuture> watchers;
   private final Map<String, ShardWorker> workers;
-  private final int workerType;
+  private final int workerChangeTypeMask;
   private final String workerChannel;
   private final Executor executor;
 
   RedisShardSubscriber(
       ListMultimap<String, TimedWatchFuture> watchers,
       Map<String, ShardWorker> workers,
-      int workerType,
+      int workerChangeTypeMask,
       String workerChannel,
       Executor executor) {
     this.watchers = watchers;
     this.workers = workers;
-    this.workerType = workerType;
+    this.workerChangeTypeMask = workerChangeTypeMask;
     this.workerChannel = workerChannel;
     this.executor = executor;
   }
@@ -235,7 +235,7 @@ class RedisShardSubscriber extends JedisPubSub {
   }
 
   void addWorker(WorkerChange workerChange) {
-    if ((workerChange.getAdd().getWorkerType() & workerType) > 0) {
+    if ((workerChange.getAdd().getWorkerType() & workerChangeTypeMask) != 0) {
       synchronized (workers) {
         workers.put(
             workerChange.getName(),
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
```diff
@@ -232,13 +235,16 @@ void onWorkerChange(WorkerChange workerChange) {
...
 
   void addWorker(WorkerChange workerChange) {
-    synchronized (workers) {
-      workers.put(
-          workerChange.getName(),
-          ShardWorker.newBuilder()
-              .setEndpoint(workerChange.getName())
-              .setFirstRegisteredAt(Timestamps.toMillis(workerChange.getAdd().getEffectiveAt()))
-              .build());
+    if ((workerChange.getAdd().getWorkerType() & workerType) > 0) {
```
werkt, line 238: != 0 here please - nothing saying that the mask has to be in non-negative range.

src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
```diff
@@ -62,16 +62,19 @@ void complete() {
 
   private final ListMultimap<String, TimedWatchFuture> watchers;
   private final Map<String, ShardWorker> workers;
+  private final int workerType;
```
werkt, line 65: Rename this to workerChangeTypeMask given the behavior.

src/test/java/build/buildfarm/instance/shard/RedisShardSubscriberTest.java
```diff
@@ -178,7 +185,7 @@ public void flush() {
 
   RedisShardSubscriber createSubscriber(
       ListMultimap<String, TimedWatchFuture> watchers, Executor executor) {
-    return new RedisShardSubscriber(watchers, /* workers= */ null, "worker-channel", executor);
+    return new RedisShardSubscriber(watchers, /* workers= */ null, /*WorkerType.NONE*/ 0, "worker-channel", executor);
```
werkt, line 188: Call WorkerType.NONE.getNumber()

werkt (changes requested):

src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
congt: Done

src/main/java/build/buildfarm/instance/shard/RedisShardSubscriber.java
congt: Done. workerChangeTypeMask is better.

src/test/java/build/buildfarm/instance/shard/RedisShardSubscriberTest.java
congt: Done

werkt (approved):
Thanks for the fix!

congt:
> Thanks for the fix!

No problem and thanks for the quick review!

PR #1821, opened 2024-08-21
fix(helm): fix statefulset {ready,live}ness probe port for worker grpc

### LINKED ISSUE #1820: Helm chart won't deploy workers because {ready,live}ness probes are using the wrong port
I normally provide detailed issue descriptions, but I believe in this case $SUBJECT and upcoming PR should be quite self-explanatory.  Please ping me otherwise.

ahachete:
Commit
https://github.com/bazelbuild/bazel-buildfarm/commit/ba057003761cbe7a21a505ddad2e694a0c6d5c6b introduced a change to move the named "worker-comm" to port 8982. However, both the livenessProbe and the readinessProbe for that container were still pointing to 8981, which prevented the container to start.

Fixes #1820


jasonschroeder-sfdc (approved):
Thank you!

PR #1822, opened 2024-08-22
Default ensureOutputsPresent to true

werkt:
BwoB (toplevel) is now the default in bazel. Make this option default to true to reduce the support burden of new users and the default client behavior.


jasonschroeder-sfdc (approved):
😎

PR #1824, opened 2024-08-22
refactor: Swap netrc lib

jasonschroeder-sfdc:
Swap from https://github.com/cdjones32/java-netrc/
to https://github.com/bearlinux/owteam_engutils_netrc

Reasons being:
- the current library is only on jitpack.io; the replacement is on Maven Central
- the replacement is clearly licensed MIT license per the POM.xml in git
  and on Maven Central metadata.


src/main/java/build/buildfarm/instance/server/NodeInstance.java
```diff
@@ -154,6 +154,7 @@ public abstract class NodeInstance extends InstanceBase {
   protected final DigestUtil digestUtil;
   protected final boolean ensureOutputsPresent;
 
+  private static final Netrc NETRC_INSTANCE = Netrc.getInstance();
```
werkt, line 157: Rather negates the singleton principle if we have to keep a reference around.

jasonschroeder-sfdc: Yoops you're right. I was thinking of avoiding all the (re-)parsing of the default `$HOME/.netrc` by holding an instance var. 🤦

werkt (approved):
Nice horizontal swap

PR #1830, opened 2024-08-28
Bug: Check WriteComplete Before Write

amishra-u:
### Issue:
The current flow allows, under specific circumstances, the write method to be called on a blob that has already been written. This leads to a failure later in the stack, resulting in a [WriteCompleteException](https://github.com/bazelbuild/bazel-buildfarm/blob/635d9fed03bedc076d019dcf4f430f20744eafe8/src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java#L267).

```
build.buildfarm.common.Write$WriteCompleteException
	at build.buildfarm.common.grpc.StubWriteOutputStream.write(StubWriteOutputStream.java:267)
	at java.base/java.io.OutputStream.write(OutputStream.java:127)
	at com.google.protobuf.ByteString$LiteralByteString.writeTo(ByteString.java:1457)
	at build.buildfarm.common.services.WriteStreamObserver.writeData(WriteStreamObserver.java:419)
	at build.buildfarm.common.services.WriteStreamObserver.handleWrite(WriteStreamObserver.java:391)
	at build.buildfarm.common.services.WriteStreamObserver.handleRequest(WriteStreamObserver.java:323)
	at build.buildfarm.common.services.WriteStreamObserver.initialize(WriteStreamObserver.java:238)
	at build.buildfarm.common.services.WriteStreamObserver.onUncommittedNext(WriteStreamObserver.java:127)
	at build.buildfarm.common.services.WriteStreamObserver.onNext(WriteStreamObserver.java:106)
	at build.buildfarm.common.services.WriteStreamObserver.onNext(WriteStreamObserver.java:54)
```

### Fix:
Ensure that the write method is not called on a blob that has already been written.

### When:
When compression is enabled, the write-completion check does not function properly based on size. The current [logic](https://github.com/bazelbuild/bazel-buildfarm/blob/635d9fed03bedc076d019dcf4f430f20744eafe8/src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java#L337) compares committed_size with expected_size. However, when compression is enabled, expected_size is set to [∞](https://github.com/bazelbuild/bazel-buildfarm/blob/635d9fed03bedc076d019dcf4f430f20744eafe8/src/main/java/build/buildfarm/instance/stub/StubInstance.java#L754), causing the check to fail.

### DIFF
--- a/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
+++ b/src/main/java/build/buildfarm/common/services/WriteStreamObserver.java
@@ -365,7 +365,7 @@ public class WriteStreamObserver implements StreamObserver<WriteRequest> {
                       "request resource_name %s does not match previous resource_name %s",
                       resourceName, name))
               .asException());
-    } else {
+    } else if (!write.isComplete()) {
       if (earliestOffset < 0 || offset < earliestOffset) {
         earliestOffset = offset;
       }
--- a/src/test/java/build/buildfarm/common/services/WriteStreamObserverTest.java
+++ b/src/test/java/build/buildfarm/common/services/WriteStreamObserverTest.java
@@ -6,6 +6,7 @@ import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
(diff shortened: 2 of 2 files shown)


amishra-u:
> You've only narrowed the opportunity to receive a completion without throwing an exception.

Yes, I agree that this change only reduces the likelihood of an exception being thrown. The main issue is that for any blob that already exists remotely, with compression enabled the write call will consistently log this error. However, with this change, this will only occur when two concurrent requests attempt to write the same blob and one succeeds—which, in practice, is rare. Additionally, I could update [this line](https://github.com/bazelbuild/bazel-buildfarm/blob/main/src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java#L337) to `return checkComplete() || writeStatus.get().getComplete() || getCommittedSize() == expectedSize;`, which would also address the excessive WriteCompleteException logging issue.


I attempted to catch the `WriteCompleteException` and return `ALREADY_EXISTS`, but I couldn't find an elegant solution. I was tempted to refactor the WriteStreamObserver code. So  made this change to mitigates the excessive logging for us, but it doesn't fully resolve the issue.


Uber android repo uses forked bazel and it seems in some case it is not making FMB call before write. Due to which buildfarm getting many duplicate write reqeusts. (this also need to be debugged and fixed)

amishra-u:
Apologies, the unit test didn’t sufficiently cover the negative case. I will update it accordingly.

PR #1833, opened 2024-09-03
Make `logging.properties` configurable via ConfigMap

### LINKED ISSUE #1794: Verbose Logging for Servers and Workers through Helm Chart
I am trying to use the HelmChart present [here](https://github.com/bazelbuild/bazel-buildfarm/tree/main/kubernetes/helm-charts/buildfarm).

My requirement is that I want verbose logging enabled for the servers and workers. And also to get some logs when remote build is triggered.
I have set the below configuration in the `values.yaml`:

```yaml
server:
    name: "shard"
    recordBesEvents: true
    grpcMetrics:
      enabled: true
      provideLatencyHistograms: true
      latencyBuckets: [0.001, 0.01, 0.1, 1, 5, 10, 20, 40, 60, +Infinity]
      labelsToReport: []
    metrics:
      #publisher: log
      logLevel: FINEST
worker:
  port: 8982
  grpcMetrics:
    enabled: true
    provideLatencyHistograms: true
    latencyBuckets: [0.001, 0.005, 0.01, 0.05, 0.075, 0.1, 0.25, 0.5, 1.0, 2.0, 5.0, 10.0]
```

By looking at the config [here](https://bazelbuild.github.io/bazel-buildfarm/docs/configuration/configuration/#metrics)

Do let me know if this is correct?

I also found that we can use debugging like [this](https://github.com/bazelbuild/bazel-buildfarm/tree/main?tab=readme-ov-file#debugging)
But I do not see any way to pass the `jvm_flag` flags in the `values.yaml`

I tried uploading the `logging.properties` through the deployment file of server and it was also placed at the appropriate place, but it still does not give any logs.

ynqa:
Resolves #1794

## What

- Manage `logging.properties` as ConfigMap and configured it to be mounted by Pod.
- Configure `JAVA_TOOL_OPTIONS` to be managed in Pod's environment variables.

## Why

> Manage `logging.properties` as ConfigMap and configured it to be mounted by Pod.

This allows `logging.properties` to be modified externally.

> Configure `JAVA_TOOL_OPTIONS` to be managed in Pod's environment variables.

This is to specify the value of `-Djava.util.logging.config.file` property representing the path of file mounted from ConfigMap.

### DIFF
--- a/kubernetes/helm-charts/buildfarm/templates/shard-worker/statefulsets.yaml
+++ b/kubernetes/helm-charts/buildfarm/templates/shard-worker/statefulsets.yaml
@@ -43,6 +43,8 @@ spec:
           env:
             - name: CONFIG_PATH
               value: /config/config.yml
+            - name: JAVA_TOOL_OPTIONS
+              value: -XX:+UseContainerSupport -XX:MaxRAMPercentage=80.0 -XX:+UseStringDeduplication -XX:+UseCompressedOops -XX:+HeapDumpOnOutOfMemoryError -Djava.util.logging.config.file=/shard-worker-log-props/logging.properties
             - name: POD_IP
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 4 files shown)


kubernetes/helm-charts/buildfarm/templates/shard-worker/statefulsets.yaml
```diff
@@ -70,8 +72,8 @@ spec:
             - mountPath: /config
               name: config
               readOnly: true
-            - mountPath: /tmp/worker
```
werkt, line 73: Was this removal intentional?

ynqa: @werkt No (Sorry for not checking). I reverted the modification.

werkt (approved):

PR #1870, opened 2024-09-28
Support multiple digest functions

werkt:
Declare new Digest type in buildfarm proto as container
Where aggregates are used, REAPI Digests are passed with decorator
Naming makes things awkward for use, lots of cleanup possible


src/main/java/build/buildfarm/backplane/Backplane.java
```diff
@@ -120,6 +120,7 @@ ScanResult<String> scanCorrelatedInvocationIndexEntries(String cursor, int count
   /** Returns a set of the names of all active storage workers. */
   Set<String> getStorageWorkers() throws IOException;
 
+  // TODO this is just a namespace, but kind of jank for just digest function as a string...
```
jasonschroeder-sfdc, line 123: Does this go here? What's the namespace?

src/main/java/build/buildfarm/cas/GrpcCAS.java
```diff
@@ -69,6 +71,7 @@ public class GrpcCAS implements ContentAddressableStorage {
   private final boolean readonly;
   private final Channel channel;
   private final ByteStreamUploader uploader;
+  // TODO
```
jasonschroeder-sfdc, line 74: Leftover TODO

src/main/java/build/buildfarm/common/DigestUtil.java
```diff
@@ -18,24 +18,43 @@
...
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Set;
 import lombok.Getter;
 
 /** Utility methods to work with {@link Digest}. */
 public class DigestUtil {
```
jasonschroeder-sfdc, line 39: This class could probably benefit from more unit tests to prevent any regressions

src/main/java/build/buildfarm/backplane/Backplane.java
werkt: 3am venting. This was going to look much more complicated, and the namespace is (iirc) the action result space mapping by digest key. Ended up being exactly the same, likely due to buildfarm Digest message.

PR #1874, opened 2024-09-30
feat: add k6 performance scripts

jasonschroeder-sfdc:
Adding some useful k6 scripts to benchmark RBE implementations, such as BuildFarm.

See also: https://k6.io


werkt (approved):

werkt:
Something to consider might be a suite of external benchmark/stress tests elsewhere under the buildfarm org.

PR #1877, opened 2024-10-02
Changed handling of JedisException in RedisShardSubscription::stop

### LINKED ISSUE #1838: RedisShardSubscription::stop is unsafe
If we are between the stopped check and an established subscription (via the jedis.subscribe call), ::stop() will not induce the current subscribe to complete due to unsubscription.

RedisShardSubscription::stop() needs to be made resilient to this state, either avoiding it, or ensuring that pending stops are completed when subscription is established.

acninetyfive:
fixes #1838 

If a call to `stop()` is made before any connection is established, the resulting JedisException will be thrown. 

Previously this exception was caught and swallowed, but this caused an issue where subscribe would never complete and the user would not be notified. Because the `stopped` Atomic Boolean is set at the beginning of the `stop()` method, the `mainLoop()` method would stop regardless of wether or not `unsubscribe()` could complete gracefully.

This change will throw an error in that case, to alert the user that the RedisSubscription object is now in "stopped" state and cannot accept connections.

The attached image shows how the race condition happens - by the time the initial call to `run()` progresses to the `mainLoop()` method, `stopped` is already set to true, and so `iterate()` is never called and the subscription is never connected. 

stop-start redis race condition

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
@@ -72,6 +72,12 @@ class RedisShardSubscription implements Runnable {
   }
 
   class SubscriptionAutoCloseable implements AutoCloseable {
+    public void initiateSubscription() {
+      SettableFuture subscribeFuture = SettableFuture.create();
+      subscriber.setSubscribeFuture(subscribeFuture);
+      manageState(SubscriptionAction.START_SUBSCRIBE);
+    }
+
     @Override
     public void close() {
       subscriber.setSubscribeFuture(null);
@@ -113,9 +119,9 @@ class RedisShardSubscription implements Runnable {
     }
 
     try (SubscriptionAutoCloseable subscriptionAutoCloseable = new SubscriptionAutoCloseable()) {
-      SettableFuture subscribeFuture = SettableFuture.create();
-      subscriber.setSubscribeFuture(subscribeFuture);
-      manageState(SubscriptionAction.START_SUBSCRIBE);
+
+      subscriptionAutoCloseable.initiateSubscription();
+
       if (subscriptionState.get() == SubscriptionState.SUBSCRIBING) {
         jedis.subscribe(subscriber, subscriptions.get().toArray(new String[0]));
       } else {
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
@@ -71,6 +71,14 @@ class RedisShardSubscription implements Runnable {
     this.client = client;
   }
 
+  class SubscriptionAutoCloseable implements AutoCloseable {
+    @Override
+    public void close() {
+      subscriber.setSubscribeFuture(null);
+      manageState(SubscriptionAction.END_SUBSCRIPTION);
+    }
+  }
+
   private synchronized void manageState(SubscriptionAction update) {
     SubscriptionState currentState = subscriptionState.get();
     switch (update) {
@@ -104,7 +112,7 @@ class RedisShardSubscription implements Runnable {
       onReset.accept(jedis);
     }
 
-    try {
+    try (SubscriptionAutoCloseable subscriptionAutoCloseable = new SubscriptionAutoCloseable()) {
       SettableFuture subscribeFuture = SettableFuture.create();
       subscriber.setSubscribeFuture(subscribeFuture);
       manageState(SubscriptionAction.START_SUBSCRIBE);
@@ -116,8 +124,6 @@ class RedisShardSubscription implements Runnable {
             "Cannot subscribe, RedisShardSubscription is in 'stopped' state "
                 + subscriptionState.get().name());
       }
-    } finally {
-      subscriber.setSubscribeFuture(null);
     }
   }
 
@@ -135,8 +141,6 @@ class RedisShardSubscription implements Runnable {
         default:
           throw e;
       }
-    } finally {
-      manageState(SubscriptionAction.END_SUBSCRIPTION);
     }
   }
 
--- a/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
+++ b/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
@@ -71,20 +71,6 @@ class RedisShardSubscription implements Runnable {
     this.client = client;
   }
 
-  class SubscriptionAutoCloseable implements AutoCloseable {
-    public void initiateSubscription() {
-      SettableFuture subscribeFuture = SettableFuture.create();
-      subscriber.setSubscribeFuture(subscribeFuture);
-      manageState(SubscriptionAction.START_SUBSCRIBE);
-    }
-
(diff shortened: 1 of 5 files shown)


src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
```diff
@@ -87,14 +87,17 @@ private void mainLoop() throws IOException {
     }
   }
 
-  public void stop() {
+  public void stop() throws JedisException {
```
werkt, line 90: JedisException is unchecked and unnecessary as a declaration here.

src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
```diff
@@ -87,14 +87,17 @@ private void mainLoop() throws IOException {
...
     if (stopped.compareAndSet(false, true)) {
       try {
         subscriber.unsubscribe();
       } catch (JedisException e) {
-        // subscriber validates with private member to determine connection status
-        // detect this condition and ignore it
-        if (!e.getMessage().endsWith(" is not connected to a Connection.")) {
+        // If stop() is called before a connection is established, log and throw the exception
+        if (e.getMessage().endsWith(" is not connected to a Connection.")) {
```
werkt, line 96: You're swallowing any JedisException other than the "is not connected" exception, which deviates from the previous behavior.

src/test/java/build/buildfarm/instance/shard/RedisShardSubscriptionTest.java
```diff
@@ -107,7 +108,6 @@ public void onSubscribe(String channel, int subscribedChannels) {
     Thread thread = new Thread(subscription);
     thread.start();
 
-    // FIXME if we stop before we are subscribed, this hangs
     subscribed.get();
```
werkt, line 111: This is now (hopefully) unnecessary complexity. Remove subscribed as it should be inconsequential.

werkt (changes requested):
Based on previous drafts, I'm not understanding this change - wasn't there a restructure to 'first'? How does just swallowing on stop accomplish the goal here?

src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
acninetyfive: I'll move the `throw` outside of the `if` block so all are thrown but the "no connection" ones get the log explaining.

src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
acninetyfive: OK, I'll remove

src/test/java/build/buildfarm/instance/shard/RedisShardSubscriptionTest.java
acninetyfive: This is required still. The first attempt to subscribe is set up to fail so that onReset will be called, so we need to wait for this future or else stop will be called before the subscription completes and we'll get a jedis exception.

acninetyfive:
> Based on previous drafts, I'm not understanding this change - wasn't there a restructure to 'first'? How does just swallowing on stop accomplish the goal here?

Not swallowing the `JedisException` on stop accomplishes the goal because it alerts the user that their object is now in the "stopped" state and that it failed to connect. 
There's no need to restructure "first". In the "onReset..." test, when the initial subscribe happens, first is set to false. Then after that fails, the second attempt triggers the `onReset` method because `iterate` and `subscribe` get called with `isReset == !first == true`

werkt:
Please have a closer look at the issue description: we're not trying to solve for a "stop()" hang, we're solving for the fact that if we stop between the stopped check and the subscribe, that the _subscription thread_ (on [this line](https://github.com/acninetyfive/bazel-buildfarm/blob/96f5b00e7e030917050a0fc57daa57c3c8cba4f7/src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java#L59)) will hang (and is what was observed to hang).

The relation to the test is only tangential - the core of the issue is that at any point in the subscription process, a stop can be called, and the subscription should never hang or block afterwards. That behavior hasn't changed here.

This is why the subscribed future should be irrelevant or reassociated - since the first response of the server is to break the connection, all we determine in that test is whether the onReset is called. The onReset callback can be the future::set, particularly because we could actually race the subscribe -> onReset -> subscribe loop as well, and the only implementation detail the test should depend upon is the onReset call.

src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
```diff
@@ -88,15 +117,26 @@ private void mainLoop() throws IOException {
...
+    } catch (JedisException e) {
+      // If stop() is called before a connection is established, log and throw the exception
+      if (e.getMessage().endsWith(" is not connected to a Connection.")) {
+        log.log(
+            Level.SEVERE,
+            "RedisShardSubscription::stop called but no connection is established. "
+                + "Subscription is now in 'Stopped' state and cannot subscribe.");
+      }
+      throw e;
+    } finally {
```
acninetyfive, line 132: This retry solves the issue that there is no way to tell if the subscription has fully completed. Instead we can tell that the unsubscribe call was successful because `attemptingSubscribe` is now false. The benefit here is that if `subscribe` was attempted, we can guarantee that `stop` will not complete successfully but leave an orphaned subscription. The downside of this change is that `stop` will now block.

src/main/java/build/buildfarm/instance/shard/RedisShardSubscription.java
```diff
@@ -52,11 +71,66 @@ class RedisShardSubscription implements Runnable {
...
+
   private void subscribe(UnifiedJedis jedis, boolean isReset) {
     if (isReset) {
       onReset.accept(jedis);
     }
-    jedis.subscribe(subscriber, subscriptions.get().toArray(new String[0]));
+
+    try (SubscriptionAutoCloseable subscriptionAutoCloseable = new SubscriptionAutoCloseable()) {
+
+      subscriptionAutoCloseable.initiateSubscription();
```
werkt, line 123: This behavior should be a part of the creation of the auto closable, not a separate method.
Make it a part of the constructor, a static on the class, or a method on the Subscription class to produce it (and initiate the transaction).

acninetyfive: I did this for the admittedly hacky reason that if all the action takes place in the constructor then the linter will complain that `subscriptionAutoCloseable` is an unused variable.
In my last commit https://github.com/buildfarm/buildfarm/pull/1877/commits/e4f05f63009765583634343ce8a5b3be7e14b959 I only used the closeable for handling the `close()` and I got that complaint from the linter. Is there a way to add an ignore/suppress, or can you suggest a better way to add a "use" for the variable?

werkt: You can just instantiate the object with no variable name, it will still auto-close, and this will likely defeat the linter.

werkt (approved):

PR #1882, opened 2024-10-06
fix(docker): set correct JVM args for java-util-logging

jasonschroeder-sfdc:
Fix the JVM argument for the java-util-logging parameter

### DIFF
--- a/jvm_flags.bzl
+++ b/jvm_flags.bzl
@@ -46,7 +46,7 @@ RECOMMENDED_JVM_FLAGS = [
     "-XX:+HeapDumpOnOutOfMemoryError",
 ]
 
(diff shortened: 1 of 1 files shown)


80degreeswest (commented):
Is logging.config not working? Which java version does it apply to? We use the existing one with no issues.

jasonschroeder-sfdc:
`logging.config` didn't work for me. This PR also aligns with what's documented in the `README`

werkt:
A bunch of our binaries have logging configuration supposedly built into the binaries, but I don't think that's working. Our logging situation is nebulous at best, but I'm still not interested (after numerous painful experiences with log4j and spring's logging) in chasing some oddball solve-it-all solution like slf4j.

PR #1884, opened 2024-10-06
feat: support cgroups v2 and v1

jasonschroeder-sfdc:
This adds support for cgroups v2 and maintains support for cgroups v1 for one more release. (v2.16.0)

Efforts made to keep v1 and v2 separate so that v1 support can be removed without a ton more refactoring.

cgroups v1 removal is tracked here: https://github.com/buildfarm/buildfarm/issues/2205

## Recommended upgrade path for existing cgroups v1 users:
- Upgrade buildfarm to v2.15.0. Note the warning in the exec-worker startup logs:
```
cgroups v1 detected at /sys/fs/cgroup ! This is the last Buildfarm version to
support v1 - Please upgrade your host to cgroups v2! See also 
https://github.com/buildfarm/buildfarm/issues/2205
```
- Ensure things still work in v1 mode
- upgrade kernel boot arguments for the node to enable v2. Reboot
- on the host node, ensure cgroups v2 is active.
```sh
stat -fc %T /sys/fs/cgroup/
```
should say `cgroup2fs`
- Start buildfarm
- Check that the cgroupv1 warning is no longer present in the start-up logs for exec workers.

----
Fixes: https://github.com/buildfarm/buildfarm/issues/1834

### DIFF
--- a/src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
+++ b/src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
@@ -26,11 +26,19 @@ import lombok.Data;
 @Data
 public class ExecutionWrappers {
   /**
-   * @field cgroups
-   * @brief The program to use when running actions under cgroups.
+   * @field cgroups2
+   * @brief The program to use when running actions under cgroups v2.
    * @details This program is expected to be packaged with the worker image.
    */
-  private String cgroups = "/app/build_buildfarm/cgexec-wrapper";
+  private String cgroups2 = "/app/build_buildfarm/cgexec-wrapper";
+
+  /**
+   * @field cgroups1
+   * @brief The program to use when running actions under cgroups v1.
+   * @details This program is expected to be packaged with the worker image.
+   */
+  @Deprecated(forRemoval = true)
+  private String cgroups1 = "/usr/bin/cgexec";
 
   /**
    * @field unshare
--- a/src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
+++ b/src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
@@ -30,7 +30,7 @@ public class ExecutionWrappers {
    * @brief The program to use when running actions under cgroups.
    * @details This program is expected to be packaged with the worker image.
    */
-  private String cgroups = "/usr/bin/cgexec";
+  private String cgroups = "/app/build_buildfarm/cgexec-wrapper";
 
   /**
    * @field unshare
--- a/src/main/java/build/buildfarm/worker/cgroup/Controller.java
+++ b/src/main/java/build/buildfarm/worker/cgroup/Controller.java
@@ -33,15 +33,19 @@ abstract class Controller implements IOResource {
     this.group = group;
   }
 
-  public abstract String getName();
+  public abstract String getControllerName();
 
   protected final Path getPath() {
-    return group.getPath(getName());
+    if (Group.VERSION == CGroupVersion.CGROUPS_V2) {
+      return group.getPath();
+    } else {
+      return group.getPath(getControllerName());
+    }
   }
 
   protected final void open() throws IOException {
     if (!opened) {
-      group.create(getName());
+      group.create(getControllerName());
       opened = true;
     }
   }
@@ -58,7 +62,7 @@ abstract class Controller implements IOResource {
     Path path = getPath();
     boolean exists = true;
     while (exists) {
-      group.killUntilEmpty(getName());
+      group.killUntilEmpty(getControllerName());
       try {
         Files.delete(path);
         exists = false;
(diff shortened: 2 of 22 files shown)


src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -521,6 +521,7 @@ private Code executeCommand(
 
     long startNanoTime = System.nanoTime();
     Process process;
+    log.log(Level.FINE, "executing " + processBuilder);
```
werkt, line 524: extra logging?

src/main/java/build/buildfarm/worker/cgroup/Controller.java
```diff
@@ -74,14 +81,15 @@ public void close() throws IOException {
...
-      return !group.isEmpty(getName());
+      return !group.isEmpty(getControllerName());
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
   }
 
   protected void writeInt(String propertyName, int value) throws IOException {
     Path path = getPath().resolve(propertyName);
+    log.log(Level.FINE, "Writing " + propertyName + "=" + value + " to " + path);
```
werkt, line 92: more logging?

src/main/java/build/buildfarm/worker/cgroup/Cpu.java
```diff
@@ -22,32 +22,50 @@ public class Cpu extends Controller {
...
+   *
+   * @param cpuCores whole cores. 1 == 1 CPU core.
+   */
+  public void setMaxCpu(int cpuCores) throws IOException {
     open();
-    return readLong("cpu.cfs_period_us");
+    if (Group.VERSION == CGroupVersion.CGROUPS_V2) {
+      writeInt("cpu.max", cpuCores * 100000);
+    } else if (Group.VERSION == CGroupVersion.CGROUPS_V1) {
+      setCFSPeriodAndQuota(100000, cpuCores * 100000);
```
werkt, line 60: Put this in as a configurable granularity constant

src/main/java/build/buildfarm/worker/cgroup/Cpu.java
```diff
@@ -22,32 +22,50 @@ public class Cpu extends Controller {
...
 
+  @Deprecated(forRemoval = true)
   public int getShares() throws IOException {
     open();
     return readLong("cpu.shares");
   }
 
+  /**
+   * Represents how much CPU shares are allocated. Note - this method does nothing for CGroups v2.
+   * All processes get equal weight.
```
werkt, line 37: This will not work - if we cannot allot more shares for a single process, then we can't have one execution able to take up multiple cores upon request.

src/main/java/build/buildfarm/worker/cgroup/Group.java
```diff
@@ -44,22 +55,67 @@ public final class Group {
...
+          if (split[2].equals("cgroup2")) {
+            return CGroupVersion.CGROUPS_V2;
+          } else if (split[2].equals("cgroup")) {
+            return CGroupVersion.CGROUPS_V1;
+          }
+        }
+        // by this point, we haven't found any `cgroup` or `cgroup2` fs types.
+      }
+    } catch (IOException e) {
+      // TODO logging
```
werkt, line 83: TODO yes, please.

src/main/java/build/buildfarm/worker/cgroup/Group.java
```diff
@@ -44,22 +55,67 @@ public final class Group {
...
-  private Group(String name, Group parent) {
+  @SuppressWarnings(
+      "PMD.MutableStaticState") // Unit tests set this. When CGroups v1 support is gone, this will
+  // go away, too.
+  protected static CGroupVersion VERSION = discoverCgroupVersion();
+
+  private static CGroupVersion discoverCgroupVersion() {
+    /* Try to figure out which version of CGroups is available. */
+    try {
+      List<String> allMounts = Files.readAllLines(Paths.get("/proc/mounts"));
```
werkt, line 66: This [stackexchange post](https://unix.stackexchange.com/questions/471476/how-do-i-check-cgroup-v2-is-installed-on-my-machine) highlights what I think is a better way forward here.  The java equivalent of `stat -fc %T /sys/fs/cgroup/` and splitting on tmpfs/cgroup is the tl;dr.

There are a lot of things I'd prefer to see made constants here regardless

src/main/java/build/buildfarm/common/config/ExecutionWrappers.java
```diff
@@ -26,11 +26,18 @@
...
    */
-  private String cgroups = "/usr/bin/cgexec";
+  private String cgroups2 = "/app/build_buildfarm/cgexec-wrapper";
+
+  /**
+   * @field cgroups1
+   * @brief The program to use when running actions under cgroups v1.
+   * @details This program is expected to be packaged with the worker image.
+   */
+  private String cgroups1 = "/usr/bin/cgexec";
```
jasonschroeder-sfdc, line 40: perhaps we leave this as `cgroups` so as not to break existing config files?

werkt (approved):
Let's. Go.

PR #1896, opened 2024-10-16
fix: NPE from e.getCause in RedisShardBackplane

jasonschroeder-sfdc:
I don't remember how I got into this state, but it seems `Throwables.throwIfInstanceOf()` doesn't like `null`. Wrap it in null check and log it.


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -779,8 +779,11 @@ private synchronized Set<String> getExecuteWorkers() throws IOException {
       return recentExecuteWorkers.get();
     } catch (RuntimeException e) {
       // unwrap checked exception mask
+      log.log(Level.WARNING, "getExecuteWorkers failed", e);
       Throwable cause = e.getCause();
-      Throwables.throwIfInstanceOf(cause, IOException.class);
+      if (cause != null) {
+        Throwables.throwIfInstanceOf(cause, IOException.class);
+      }
       throw e;
```
werkt, line 787: Prefer a wrap in IOException in this case - I want to fit within the expectation of IOException, and to figure out how it's possible that this happened.

werkt (approved):
Since you are bothering to fix this at all, you must have a backtrace pointing you here, can you post it in the description so that I see where we've missed something?

jasonschroeder-sfdc:
>you must have a backtrace pointing you here
I'll hold on merging this until it surfaces again

PR #1904, opened 2024-10-21
Build arm64v8 containers and publish

werkt:
Notable exceptions: arm compiler takes issue with the skip-sleep source usage of SYS_time and ptrace

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -20,8 +20,8 @@ buildifier(
 # additional dependencies alongside the buildfarm agents.
 ARCH = [
     # keep sorted
-    # "aarch64", # TODO
     "amd64",
+    "arm64v8",
 ]
 
 DEFAULT_IMAGE_LABELS = {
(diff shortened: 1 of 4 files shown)


BUILD
```diff
@@ -132,6 +152,12 @@ pkg_tar(
     tags = ["container"],
 )
 
+pkg_tar(
+    name = "layer_tini_arm64v8",
+    srcs = [":tini.binary-arm64v8"],
+    tags = ["container"],
+)
+
```
jasonschroeder-sfdc, line 160: Could we have absorbed this into `pkg_tar(name="layer_tini")` and use a `select()` to pick the `srcs`?

example:
```starlark
pkg_tar(
    name = "layer_tini",
    srcs = select({
        "@platforms//cpu:arm": [":tini.binary-arm64v8"],
        "//conditions:default": [":tini.binary"],
    }),
    tags = ["container"],
)
```

jasonschroeder-sfdc:
This PR works "out of the box" on:
- darwin-arm64 (MacBook pro with M3 Max, macOS Sonoma)
- linux-arm64 (RasberryPi OS on Pi5 with `./bazelisk build //src/...`)

jasonschroeder-sfdc (approved):

jasonschroeder-sfdc:
Resolved via https://github.com/buildfarm/buildfarm/pull/1946

PR #1915, opened 2024-10-27
fix: digests of 0 bytes

jasonschroeder-sfdc:
If the ByteString blob was 0-length, DigestUtils always returned `Digest.getDefaultInstance()`. Added tests to ensure the empty digests work, with test vectors.

Fixes: #1910


jasonschroeder-sfdc:
Superseded by https://github.com/buildfarm/buildfarm/pull/1917

PR #1985, opened 2024-11-13
refactor: prefer Durations to convert Duration protobufs

jasonschroeder-sfdc:


src/main/java/build/buildfarm/common/Poller.java
```diff
@@ -53,10 +53,8 @@ private Duration getWaitTime() {
     private void waitForNextDeadline() {
       try {
         Duration waitTime = getWaitTime();
-        if (waitTime.getSeconds() != 0 || waitTime.getNanos() != 0) {
-          wait(
-              waitTime.getSeconds() * 1000 + waitTime.getNanos() / 1000000,
-              waitTime.getNanos() % 1000000);
+        if (Durations.isPositive(waitTime)) {
+          wait(Durations.toMillis(waitTime), waitTime.getNanos() % 1000000);
```
werkt, line 57: The mod here is still complicated :/

werkt (approved):

src/main/java/build/buildfarm/common/Poller.java
jasonschroeder-sfdc: I didn't change this bit, but I'm glad it caught your eye too. I don't think the mod is needed? `getNanos()` should return the nano portion of the Duration, which _should be_ `<= 1e9`?

werkt: I think giving the wait time sub-millisecond granularity is probably unnecessary here. Would definitely accept a rejection of such a specification, but will accept this as is.

werkt (approved):

PR #2049, opened 2024-12-02
Fix CAS unreferenced percentage calculation

jagadeshjai:
- Fix CAS unreferenced percentage calculation to handle zero unreferenced entries and avoid 'Infinity%'

Resolves https://github.com/buildfarm/buildfarm/issues/2048

### DIFF
--- a/src/main/java/build/buildfarm/tools/WorkerProfile.java
+++ b/src/main/java/build/buildfarm/tools/WorkerProfile.java
@@ -71,7 +71,7 @@ class WorkerProfile {
       System.out.printf(
           (strFloatFormat) + "%n",
           "Percentage of Unreferenced Entry",
-          1.0 * response.getCasEntryCount() / response.getCasUnreferencedEntryCount());
+          1.0 * (unreferencedEntryCount / entryCount));
     }
     System.out.printf(
         (strIntFormat) + "%n",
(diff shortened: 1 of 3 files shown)


jagadeshjai:
@werkt Please take a look.

src/main/java/build/buildfarm/tools/WorkerProfile.java
```diff
@@ -68,10 +68,11 @@ private static void analyzeMessage(String worker, WorkerProfileMessage response)
     System.out.printf(
         (strIntFormat) + "%n", "Current Unreferenced Entry Count", unreferencedEntryCount);
     if (entryCount != 0) {
-      System.out.printf(
-          (strFloatFormat) + "%n",
-          "Percentage of Unreferenced Entry",
-          1.0 * response.getCasEntryCount() / response.getCasUnreferencedEntryCount());
+      System.out.println(
+          format(
```
jasonschroeder-sfdc, line 72: I think this line needs to be removed

jagadeshjai: I've pushed the changes, however @werkt already included these changes along with additional  fixups in https://github.com/buildfarm/buildfarm/pull/2054

PR #2052, opened 2024-12-03
Correct logspam for pid process kill loop

werkt:


src/main/java/build/buildfarm/worker/cgroup/Group.java
```diff
@@ -153,7 +153,7 @@ public void killUntilEmpty(String controllerName) throws IOException {
         !pids.isEmpty();
         pids = getPids(controllerName)) {
       killAllProcesses(pids);
-      if (deadline.isExpired() || !pids.containsAll(prevPids) || prevPids.containsAll(pids)) {
+      if (deadline.isExpired() || !pids.containsAll(prevPids) || !prevPids.containsAll(pids)) {
```
jasonschroeder-sfdc, line 156: `pids` should keep shrinking until `pids.isEmpty()`.
Something smells off with this condition: (`lhs.containsAll(rhs)|| rhs.containsAll(lhs)`).

 Should it just be `!pids.equals(prevPids)` ? or `pids.size() < prevPids.size()`?

werkt: pids is not guaranteed to keep shrinking, if a process in death throes is forking, which is rather believable if we've even gotten to this point. That throws off the '<' comparison being valid. !equals has merit, I'm not sure why this wasn't originally considered (not actually something I added originally, but definitely modified to try to stop the spam before).

jasonschroeder-sfdc (approved):

PR #2054, opened 2024-12-03
Fix WorkerProfile metrics for ExecutedActionResult

werkt:
Compute WorkerProfile metrics with ExecuteResponse result execution_metadata if the response has a result.
Implementation is duplicated in 3 places, this should eventually be resolved.


src/main/java/build/buildfarm/worker/PutOperationStage.java
```diff
@@ -144,7 +149,7 @@ void addOperation(ExecutedActionMetadata metadata) {
   }
 
   private static Duration betweenIfAfter(Timestamp a, Timestamp b) {
-    if (Timestamps.compare(a, b) > 0) {
+    if (Timestamps.compare(b, a) > 0) {
```
jasonschroeder-sfdc, line 152: Does this needs some tests? 😁 
Consider `Timestamps.compare(a,b) <= 0`?

werkt: The entire class needs a rework, I'm just trying to get it back on its feet before I take blowtorch to it.
No need for the <=, a < works here, as an equal duration would be the default instance. I switched it to < 0 and added a comment for now.

jasonschroeder-sfdc (approved):

PR #2067, opened 2024-12-11
refactor: prefer ThreadLocalRandom.current().nextDouble() over Math.r…

jasonschroeder-sfdc:
…andom()

openrewrite: https://docs.openrewrite.org/recipes/java/migrate/util/replacemathrandomwiththreadlocalrandomrecipe

ThreadLocalRandom has better multi-threaded performance


werkt (approved):
this is fine, but fwiw, I think this is currently dead code - there are no calls to the only use I'm able to find in terms of suggestedWorkersToScaleDown

PR #2068, opened 2024-12-11
Persistent worker nits

jasonschroeder-sfdc:
- prefer Lombok logger for consistency
- prefer `Files.readString(...)`
- Guava deprecations


persistentworkers/src/main/java/persistent/common/CommonsPool.java
```diff
@@ -21,7 +21,7 @@ public V borrowObject(K key) throws IOException, InterruptedException {
     try {
       return super.borrowObject(key);
     } catch (Throwable t) {
-      Throwables.propagateIfPossible(t, IOException.class, InterruptedException.class);
+      Throwables.throwIfUnchecked(t);
```
werkt, line 24: This does not manage to throw an IOException or InterruptedException as a result...

jasonschroeder-sfdc: 🤔 I need to re-(re-)read https://github.com/google/guava/wiki/Why-we-deprecated-Throwables.propagate

werkt: This seems more relevant: https://guava.dev/releases/23.0/api/docs/com/google/common/base/Throwables.html#propagateIfInstanceOf-java.lang.Throwable-java.lang.Class-

werkt (approved):

PR #2075, opened 2024-12-12
refactor(persistent): remove limits.persistentWorkerCommand

jasonschroeder-sfdc:
It's unused. Instead, log the action mnemonic.


werkt (approved):
I think this is correct. There's really no usage other than this log message? We're not intending on validating key + command? (seems wrong to do so anyway).

jasonschroeder-sfdc:
There's really no usage :smile: (it was unhooked a while ago - I can't immediately find that PR)

PR #2084, opened 2024-12-18
Unique Execution Owners

werkt:
Permit specification of an array of users that can be used for tree ownership and available as pool resource names for executionWrappers. The number of execOwners must be greater than or equal to the sum of input fetch, execution, and report result stage widths.


as-nobody.c
```diff
@@ -28,7 +36,10 @@ main(int argc, char *argv[])
     perror("setreuid");
     return EXIT_FAILURE;
   }
-  execvp(argv[1], argv + 1);
+  for (int i = argi; argv[i] != NULL; i++) {
+    printf("%s\n", argv[i]);
+  }
```
jasonschroeder-sfdc, line 41: intentional debugging?

werkt: Good Catch, yes.

src/main/java/build/buildfarm/common/config/Worker.java
```diff
@@ -34,6 +34,7 @@ public class Worker {
   private boolean linkInputDirectories = true;
   private List<String> linkedInputDirectories = Arrays.asList("(?!external/)[^/]+");
   private String execOwner;
+  private List<String> execOwners = new ArrayList<>();
```
jasonschroeder-sfdc, line 37: question: could be a `Set<String>`? I assume this owner is platform-independent and i could also specify a uid (int) for unix-y OS's?

jasonschroeder-sfdc: thought: possible future enhancement: a range of uids.

src/main/java/build/buildfarm/worker/resources/LocalResourceSetUtils.java
```diff
@@ -37,50 +43,95 @@
...
       this.resourceSet = resourceSet;
     }
 
+    @Override
+    public synchronized void release(Stage stage) {
+      // only remove resources that match the specified stage
+      // O(n) search for staged resources
+      Iterator<SemaphoreLease> semaphoreIter = semaphores.iterator();
+      while (semaphoreIter.hasNext()) {
+        SemaphoreLease semaphore = semaphoreIter.next();
```
jasonschroeder-sfdc, line 68: suggestion: is there a more modern way to do this? :) Streams?

src/main/java/build/buildfarm/worker/resources/LocalResourceSetUtils.java
```diff
@@ -37,50 +43,95 @@
...
+        SemaphoreResource resource = resourceSet.resources.get(resourceName);
+        semaphoreRelease(resource.semaphore(), resourceName, semaphore.amount());
+      }
+      Iterator<PoolLease> poolIter = pools.iterator();
+      while (poolIter.hasNext()) {
+        PoolLease pool = poolIter.next();
+        if (pool.stage() != stage) {
+          continue;
+        }
+        poolIter.remove();
```
jasonschroeder-sfdc, line 83: issue: safe to mutate while iterating?

src/main/java/build/buildfarm/worker/shard/FuseExecFileSystem.java
```diff
@@ -74,8 +80,10 @@ public Path createExecDir(
       Map<build.bazel.remote.execution.v2.Digest, Directory> directoriesIndex,
       DigestFunction.Value digestFunction,
       Action action,
-      Command command)
+      Command command,
+      UserPrincipal owner)
       throws IOException, InterruptedException {
+    // FIXME insist that the FUSE perform permission checks against the owner
```
jasonschroeder-sfdc, line 86: question: do we need a GH Issue to track this?

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -281,12 +290,63 @@ public QueuedOperation getQueuedOperation(QueueEntry queueEntry)
     return ProtoUtils.parseQueuedOperation(queuedOperationBlob, queueEntry);
   }
 
+  // FIXME make OwnedClaim with owner
+  // how will this play out with persistent workers, should we have one per user?
```
jasonschroeder-sfdc, line 294: maybe we just don't make persistent workers work with multiple exec owner feature...

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -311,30 +320,119 @@ private ExecFileSystem createFuseExecFileSystem(
...
     } catch (IOException e) {
       ConfigurationException configException =
-          new ConfigurationException("Could not locate exec_owner");
+          new ConfigurationException("Could not locate " + name);
       configException.initCause(e);
       throw configException;
     }
   }
 
+  private static final class Unqueue<T> extends AbstractCollection<T> implements Queue<T> {
```
jasonschroeder-sfdc, line 335: todo: a pinch of javadoc? the class name doesn't immediately tell me the purpose..

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -311,30 +320,119 @@ private ExecFileSystem createFuseExecFileSystem(
...
   private ExecFileSystem createExecFileSystem(
       InputStreamFactory remoteInputStreamFactory,
       ExecutorService removeDirectoryService,
       ExecutorService accessRecorder,
       ExecutorService fetchService,
-      ContentAddressableStorage storage)
+      ContentAddressableStorage storage,
+      LocalResourceSet resourceSet,
+      String ownerName,
+      List<String> ownerNames)
```
jasonschroeder-sfdc, line 401: suggestion: consider a `Set<String>` or `Collection<String>`?

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -615,6 +713,17 @@ public double get() {
...
 
+    int inputFetchStageWidth = configs.getWorker().getInputFetchStageWidth();
+    int executeStageWidth = configs.getWorker().getExecuteStageWidth();
+    int reportResultStageWidth = configs.getWorker().getReportResultStageWidth();
+    LocalResourceSet resourceSet = LocalResourceSetUtils.create(configs.getWorker().getResources());
+    List<String> execOwners = configs.getWorker().getExecOwners();
+    if (!execOwners.isEmpty()
+        && execOwners.size() < inputFetchStageWidth + executeStageWidth + reportResultStageWidth) {
+      throw new ConfigurationException(
+          "execOwners is not large enough to fill requested stage widths");
```
jasonschroeder-sfdc, line 724:
```suggestion
          "execOwners is not large enough to fill requested stage widths. Need at least " + (inputFetchStageWidth + executeStageWidth + reportResultStageWidth) + " owners.");
```

jasonschroeder-sfdc (approved):

src/main/java/build/buildfarm/worker/resources/LocalResourceSetUtils.java
werkt: Only safe to mutate with iterator while iterating:
https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/util/Iterator.html
> Iterators allow the caller to remove elements from the underlying collection during the iteration with well-defined semantics.

src/main/java/build/buildfarm/worker/shard/FuseExecFileSystem.java
werkt: Much bigger issue overall: what to do with the FUSE impl. It would be great, but needs to be rewritten (leaks inodes in mounted filesystem). This is a note for the future, when I come through here and think that this already had permission checks.

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
werkt: Roughly I agree - change justification is that executions should be isolated and not able to access peer address spaces. I think persistent workers is probably going to fly in the face of this. The problem with this feature and persistent workers is the life of a daemon beyond the end of the report result stage. Needs more thought overall.

src/main/java/build/buildfarm/worker/shard/Worker.java
werkt: Can add. If you have a good idea to name a thing like a queue that only puts out the same object, I'm all ears. Hose? Fountain? CloneGun?

src/main/java/build/buildfarm/worker/resources/LocalResourceSetUtils.java
werkt: Not with the removal that I can see. I'd change the reference if I did a filter, not sure if there's a stream-y predicated remover.

src/main/java/build/buildfarm/common/config/Worker.java
werkt: The convention here is that we have to be able to look up the user on the filesystem with https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/nio/file/attribute/UserPrincipalLookupService.html#lookupPrincipalByName(java.lang.String), and this change doesn't alter this convention from the original execOwner implementation to provide lookup by uid. I honestly don't know what would happen if we looked up that name (not in the use case). It is as platform independent as the above lookup.
This being the list of users that is presented by the queue, there's nothing conventionally keeping multiple of the same user (for god knows what reason) being presented here. A set isn't the right messaging here - you want 99% of your executions under one user, and 1% under another, this is a way to do it.
yaml's syntax permits some niceties here to deal with the ranges of names, and if we happen to work for uids, that will be true also:
`execOwners: #@ ["buildfarm-executor-{}".format(i) for i in range(1, 100)]`

src/main/java/build/buildfarm/worker/shard/Worker.java
werkt: Addressed in previous comment about the owner spectra.

src/main/java/build/buildfarm/worker/shard/Worker.java
werkt: Settled on a Dispenser and added javadoc

PR #2104, opened 2025-01-14
Migrate to Cgroups v2

jasonschroeder-sfdc:
Remove support for cgroups v1 and move to cgroups v2.

Not tested: 
- [ ] Windows

Outstanding before merge:
- [x] What do to for "minCpu" in cgroups v2? Shares are gone.
- [ ] should the config be renamed to `alwaysUseCgroupsv2` to specify v2?

### DIFF
--- a/src/main/java/build/buildfarm/worker/cgroup/Cpu.java
+++ b/src/main/java/build/buildfarm/worker/cgroup/Cpu.java
@@ -30,16 +30,21 @@ public class Cpu extends Controller {
   }
 
   /**
-   * Represents how much CPU shares are allocated. Note - this method does nothing for CGroups v2.
-   * All processes get equal weight.
+   * Represents how much CPU shares are allocated.
    *
+   * @see <a
+   *     href="https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#weights">Cgroups
+   *     v2: Weights</a>
    * @param shares
    * @throws IOException
    */
-  @Deprecated
   public void setShares(int shares) throws IOException {
-    open();
-    // TODO what to do for cgroups v2?
+    if (shares > 0) {
+      open();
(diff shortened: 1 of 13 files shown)


src/main/java/build/buildfarm/worker/cgroup/Cpu.java
```diff
@@ -17,37 +17,38 @@
...
+  /**
+   * Represents how much CPU shares are allocated. Note - this method does nothing for CGroups v2.
+   * All processes get equal weight.
+   *
+   * @param shares
+   * @throws IOException
+   */
+  @Deprecated
   public void setShares(int shares) throws IOException {
     open();
```
jasonschroeder-sfdc, line 41: A little stuck on what to put into this method

werkt: I've spent a good amount of time looking at the cgroups v2 docs, and I believe this should be cpu.weight. The overall limit should be accurate in the parent to limit to cpu.max, and the existing math to give this 1024 * slots should work out - I don't think the coefficient is meaningful at all if you want to take it out. I think min-cpu will work in this regard, and that the overall limit from the parent (which should reduce the cpu.max from the # of cores on the machine to the # of slots) will enforce correct behavior.

werkt: Actually, based on the docs, we should definitely not have the coefficient. The value is expected to be in the range [1, 10000], and we don't have an effective way of offering partial cpus currently.

jasonschroeder-sfdc: implemented in [72bf300](https://github.com/buildfarm/buildfarm/pull/2104/commits/72bf3009ab0fdef6f2081ad6f411b4d501c5f744)

PR #2106, opened 2025-01-17
Initial import of AOSP client side configuration document

alsutton:
Buildfarm works really well with the Android AOSP build, but the client side needs some configuration settings which are poorly documented.

This PR adds a documented client-side configuration which works with the Android AOSP build system (soong/ninja), allowing folk to easily setup their environment, verify RBE works, and then they can tweak Buildfarm as needed.


NopeNopeGuy:
Hey, it seems like you have forgotten some variables here.

RBE_JAR isn't set to 1 but the exec_strategy is set. You can also set RBE_LINT if you want to. 
Also, RBE_METALAVA_POOL, RBE_JAVA_POOL and RBE_LINT_POOL should be manually set because by default it will be java16, so if the pool java16 doesn't exist it will fail to build.

You can refer to my post here for all the variables I use: https://nopenopeguy.github.io/rbe

alsutton:
These are the set that worked for me. If you're having a better experience with another set of settings, it might be worth opening an alternate PR and we can close this one out.

maksonlee:
Question: cache footprint from examples/bf-run start (≈700 GiB at ~20% AOSP build)

Using examples/bf-run start on a single machine to verify client-side config, I observed the worker cache grow to ~700 GiB when the AOSP build was only ~20% complete.

Is this expected for the minimal example?

PR #2110, opened 2025-01-21
Interpret WriteCompleteException in WSO with future

werkt:
Prevent propagation of the WriteCompleteException to the observer response. If it is thrown by getOutput, the precondition is that the future is complete, and will eventually be handled with status.

### DIFF
--- a/src/main/java/build/buildfarm/common/Write.java
+++ b/src/main/java/build/buildfarm/common/Write.java
@@ -32,6 +32,10 @@ public interface Write {
 
   boolean isComplete();
 
+  /**
+   * Implementations may throw WriteCompleteException, but as a precondition, the future must be
+   * complete.
+   */
   FeedbackOutputStream getOutput(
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/common/Write.java
```diff
@@ -32,6 +32,10 @@ public interface Write {
 
   boolean isComplete();
 
+  /**
+   * Implementations may throw WriteCompleteException, but as a precondition, the future must be
+   * complete.
```
jasonschroeder-sfdc, line 37: Should we codify this precondition?

jasonschroeder-sfdc (approved):

src/main/java/build/buildfarm/common/Write.java
werkt: What would you recommend for codifying this?

jasonschroeder-sfdc: I got nothin'. ✅

PR #2116, opened 2025-01-29
chore(helm): add topologySpreadConstraints param

Smana:
Missing `topologySpreadConstraints` parameter.
Bump Helm version (though I'm not sure this is the proper workflow)

I don't know if you should also change the image tag versions?


jasonschroeder-sfdc (approved):

jasonschroeder-sfdc:
Thank you! I'll do a better job maintaining the helm chart.

PR #2121, opened 2025-02-03
Release claim when execute or report result stage errors

werkt:
Resources are leaked when execute or report result stages fail polls and subsequently invoke the error acceptor.


80degreeswest (approved):
:shipit:

PR #2122, opened 2025-02-04
feat: enable to set {max, min} idle for jedis

ynqa:
### What

- Enable to set `{Max, Min}Idle` and `TimeBetweenEvictionRuns` for jedis via backplane configs

### Why

- Observing `redis_connections_received_total` metric from redis-exporter, I noticed that a significantly higher number of connection requests occurred without settings (means `MaxTotal=4000`, `MaxIdle=8` and `MinIdle=0`)
  - Therefore, I think these parameters should be configurable the same as `MaxTotal` to reduce the overhead of reconnections
- For the same reason, it would be desirable to make `TimeBetweenEvictionRuns` configurable as well.
  - Additionally, setting this value to -1 disables it, allowing you to observe how many connections (e.g. by `redis_connected_clients` metric from redis-exporter) occurred across the entire build for adjusting `{Max, Min}Idle` and `MaxTotal`


examples/config.yml
```diff
@@ -60,6 +60,9 @@ backplane:
   redisPassword:
   redisNodes:
   jedisPoolMaxTotal: 4000
+  jedisPoolMaxIdle: 4000
+  jedisPoolMinIdle: 500
+  jedisTimeBetweenEvictionRunsMillis: -1
```
jasonschroeder-sfdc, line 65: Could we make these the same as the defaults (from `Backplane.java`?)

src/main/java/build/buildfarm/instance/shard/JedisClusterFactory.java
```diff
@@ -208,6 +209,10 @@ private static Supplier<UnifiedJedis> createJedisClusterFactory(
   private static ConnectionPoolConfig createConnectionPoolConfig() {
     ConnectionPoolConfig connectionPoolConfig = new ConnectionPoolConfig();
     connectionPoolConfig.setMaxTotal(configs.getBackplane().getJedisPoolMaxTotal());
+    connectionPoolConfig.setMaxIdle(configs.getBackplane().getJedisPoolMaxIdle());
+    connectionPoolConfig.setMinIdle(configs.getBackplane().getJedisPoolMinIdle());
+    connectionPoolConfig.setTimeBetweenEvictionRuns(
```
jasonschroeder-sfdc, line 214: You mentioned to set this to `-1` to disable the behavior. Does this need a `-1` check here, or does Jedis/Apache Commons Pool recognize a -1 Millis Duration?

ynqa: Since Apache Commons Pool2 sets -1 as the default value, I believe there is no need for Buildfarm to perform positive/negative checks.

See https://github.com/apache/commons-pool/blob/rel/commons-pool-2.12.0/src/main/java/org/apache/commons/pool2/impl/BaseObjectPoolConfig.java

```java
public static final long DEFAULT_TIME_BETWEEN_EVICTION_RUNS_MILLIS = -1L;

public static final Duration DEFAULT_DURATION_BETWEEN_EVICTION_RUNS = Duration.ofMillis(DEFAULT_TIME_BETWEEN_EVICTION_RUNS_MILLIS);

public void setTimeBetweenEvictionRuns(final Duration timeBetweenEvictionRuns) {
    this.durationBetweenEvictionRuns = PoolImplUtils.nonNull(timeBetweenEvictionRuns, DEFAULT_DURATION_BETWEEN_EVICTION_RUNS);
}
```

examples/config.yml
ynqa: Fixed at https://github.com/buildfarm/buildfarm/pull/2122/commits/615a8fb0561eb95e7b299d78f8420bd758e47b3f.

ynqa: See them for default
- https://github.com/apache/commons-pool/blob/rel/commons-pool-2.12.0/src/main/java/org/apache/commons/pool2/impl/GenericObjectPoolConfig.java
- https://github.com/redis/jedis/blob/v5.2.0/src/main/java/redis/clients/jedis/ConnectionPoolConfig.java

werkt (approved):
Seems fine for me, just adds to the pile of backplane configuration

PR #2141, opened 2025-02-20
Wrap Action during transform in queued metadata

werkt:
All operations delivered to watchers for strip and client presentation must contain metadata of either ExecuteOperationMetadata or QueuedOperationMetadata. Contain the Action that is possible during transformation and ensure non-sequenced interpretation for watcher action result write-through.


src/test/java/build/buildfarm/instance/shard/ServerInstanceTest.java
```diff
@@ -1001,7 +1001,7 @@ public void watchExecutionFutureIsErrorForObserveException()
   }
 
   @Test
-  public void watchExecutionCallsBackplaneForIncompleteOperation() throws IOException {
+  public void wtchExecutionCallsBackplaneForIncompleteOperation() throws IOException {
```
jasonschroeder-sfdc, line 1004:
```suggestion
  public void watchExecutionCallsBackplaneForIncompleteOperation() throws IOException {
```

jasonschroeder-sfdc (approved):

PR #2142, opened 2025-02-20
Migrate AC config lookup to instantiation

werkt:
Eliminates config dependency with no disadvantage for control


jasonschroeder-sfdc (approved):
Nice de-coupling! Thanks!!

PR #2144, opened 2025-02-20
AC Metrics + Log Spam Cleanup

werkt:


80degreeswest (approved):

src/main/java/build/buildfarm/server/services/ActionCacheService.java
```diff
@@ -96,24 +109,30 @@ public void onSuccess(@Nullable ActionResult actionResult) {
...
+            if (status.getCode() != Status.Code.CANCELLED) {
+              log.log(
+                  Level.WARNING,
+                  format(
+                      "getActionResult(%s): %s",
+                      request.getInstanceName(), DigestUtil.toString(actionDigest)),
+                  t);
+            }
+            try {
+              responseObserver.onError(status.asException());
```
jasonschroeder-sfdc, line 127: this was moved out of the "if" above. OK if responseObserver.onError(...) for Cancelled calls?

PR #2154, opened 2025-02-20
Provide inputDirectories as CFC entries

werkt:
Size charged for each input directory is the weight of the blobs stored uniquely within it. Storage within a directory is isolated from the blobs which supply contents, and a single unspecialized reference is maintained for each directory, eliminating lock contention and observed heavy collateral costs for wide input fetch stages and post-report result exec dir reclamation on input deference.

Previous behavior moved to LegacyDirectoryCFC, accessible through config switches.


jasonschroeder-sfdc:
>Previous behavior moved to LegacyDirectoryCFC, accessible through config switches.
Can you elaborate? I couldn't find a config switch (from a YAML config?)

werkt:
> > Previous behavior moved to LegacyDirectoryCFC, accessible through config switches.
> 
> Can you elaborate? I couldn't find a config switch (from a YAML config?)

Good catch, it was missed. I'm going to add the config option, but I don't want to document/exemplify it.
I expect the only reason to try it would be to compare performance between the two for anyone that experiences weirdness. Flag will be `legacyDirectoryFileCache`, default to false.

jasonschroeder-sfdc (approved):

PR #2178, opened 2025-02-26
fix: protect runtime exceptions from remote executor start-up

jasonschroeder-sfdc:
This aids in debugging for https://github.com/buildfarm/buildfarm/issues/2175

### DIFF
--- a/src/main/java/build/buildfarm/worker/Executor.java
+++ b/src/main/java/build/buildfarm/worker/Executor.java
@@ -521,16 +521,27 @@ class Executor {
 
       WorkFilesContext filesContext =
           WorkFilesContext.fromContext(execDir, execTree, executionContext.command);
-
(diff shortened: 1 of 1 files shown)


src/main/java/build/buildfarm/worker/Executor.java
```diff
@@ -521,16 +521,27 @@ private Code executeCommand(
...
+            limits,
+            timeout,
+            PersistentExecutor.defaultWorkRootsDir,
+            resultBuilder);
+      } catch (RuntimeException e) {
+        log.log(
+            Level.SEVERE,
+            format(
+                "error starting persistent worker for %s. argslist=%s",
+                operationName, String.join(",", arguments)),
```
werkt, line 539: I'd prefer an array-[inspect](https://apidock.com/ruby/Object/inspect) like output here (might require some hunting). I expect these arguments can get complex.

PR #2179, opened 2025-02-27
fix: histograms in seconds instead of milliseconds (#2132)

ankitpatisfdc:
Salesforce-Work-Item: W-17824138


jasonschroeder-sfdc:
Thanks @ankitpatisfdc - can you pull out 3/5 commits as separate PRs and just keep the histogram commits here? you can squash the format fix. Thank you!

PR #2181, opened 2025-02-27
fix: bind worker to port 9091 unconditionally

ankitpatisfdc:
@jasonschroeder-sfdc Please review.

### DIFF
--- a/run_worker
+++ b/run_worker
@@ -14,5 +14,11 @@ else
   config=$1
 fi
 
+declare -a prometheus_port_arg
+if lsof -ti :9090 &>/dev/null
+then
+    prometheus_port_arg=(--prometheus_port=9091)
+fi
+
 # Run Worker
-bazelisk run //src/main/java/build/buildfarm:buildfarm-shard-worker -- --jvm_flag=-Djava.util.logging.config.file=$PWD/examples/logging.properties $config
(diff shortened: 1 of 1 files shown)


run_worker
```diff
@@ -14,5 +14,11 @@ else
...
 
+declare -a prometheus_port_arg
+if lsof -ti :9090 &>/dev/null
+then
+    prometheus_port_arg=(--prometheus_port=9091)
+fi
+
 # Run Worker
-bazelisk run //src/main/java/build/buildfarm:buildfarm-shard-worker -- --jvm_flag=-Djava.util.logging.config.file=$PWD/examples/logging.properties $config
+bazelisk run //src/main/java/build/buildfarm:buildfarm-shard-worker -- --jvm_flag=-Djava.util.logging.config.file="$PWD/examples/logging.properties" "${prometheus_port_arg[@]}" "$config"
```
jasonschroeder-sfdc, line 24: I think maybe we just always do port 9091 every time, so as not to surprise the developer

80degreeswest: I agree, let's not making it a guessing game.

jasonschroeder-sfdc (approved):

80degreeswest (approved):

PR #2182, opened 2025-02-27
fix: histograms in seconds instead of milliseconds (#2132)

ankitpatisfdc:
@jasonschroeder-sfdc Please review.


jasonschroeder-sfdc (approved):

werkt (approved):
Thanks!

PR #2184, opened 2025-03-10
chore: reduce log spam for duplicate operations

jasonschroeder-sfdc:
This can come up a lot. Reducing to FINE.

(Curiously.. this logs a lot from the Worker but the Server's prometheus Counter for `merged_executions` doesn't grow. 🤔 )


werkt:
These are not merged_executions, this is actually an indicator of one of several rather severe problems with an installation:
Clock Skew between Servers and Workers
Redis lag in handling execution lease polls
Network lag for the same
CPU exhaustion or other reason for polls to be delayed (uninterruptible sleep?)

This happens when an execution lease is revoked for a worker by a server, which reschedules it and it arrives on the same worker, or perhaps an accounting error occurs as to which executions a worker believes it owns.

jasonschroeder-sfdc:
Is there additional troubleshooting I could do to make this severe issue more actionable?

PR #2188, opened 2025-03-14
Implement Redis retry functionality

meithon:
I have modified the code to use Jedis retry functionality. This allows Redis commands to be retried even if execution fails.

## Compatibility
This change modifies internal details of the Redis client implementation but does not affect existing APIs. As a default value of 0 seconds is provided, behavior remains unchanged from previous versions.
However, error messages may change from `unexpected steam end` to `No more attempts left.` or `retry deadline expired`.

If this change is not acceptable, it would be necessary to revert to using DefaultCommandExecutor from the previous Jedis client implementation.


werkt:
This doesn't apply symmetric behavior between Pooled and Cluster connection setups.

I'm not sure what this means:
> If this change is not acceptable, it would be necessary to revert to using DefaultCommandExecutor from the previous Jedis client implementation.

What would make this necessary?

meithon:
I thought the your "symmetric behavior" was about how retry duration is specified between cluster and pooled connections, so I've addressed this by ensuring both can be configured appropriately. If `jedisRetryDurationMillis` is not specified, the retry behavior remains the same as before PR for both pool and cluster connections. When specified, both connection types will use the same retry duration value.

---

> I'm not sure what this means:
>
 >>   If this change is not acceptable, it would be necessary to revert to using DefaultCommandExecutor from the previous Jedis client implementation.
>
>What would make this necessary?

This statement was simply highlighting that error messages will change (from "unexpected stream end" to "No more attempts left" or "retry deadline expired"). It wasn't clear how this might impact Buildfarm users, but I expect it wouldn't be problematic in most cases. I included this note out of caution rather than any specific concern - there's no deeper meaning to it.

PR #2200, opened 2025-03-24
Allow for merging of base and custom config yamls

80degreeswest:
Currently, we have defaults specified in the examples/config.yml as well as java objects. This increases likelyhood of things becoming out of sync as it currently is the case. Additionally There is no way to currently overlay multiple yaml configs on top of each other. This PR addresses both of these issues by adding the ability to overlay user specified yaml file on top of a single source of truth config.yml.

One of the main benefits here will be easier CI orchestration by allowing for a simpler way to generate configuration for each type (server/workers) and different environments.

Multiple config files are allowed as a comma separated list.

### DIFF
--- a/_site/docs/configuration/configuration.md
+++ b/_site/docs/configuration/configuration.md
@@ -5,20 +5,57 @@ nav_order: 4
 has_children: true
 ---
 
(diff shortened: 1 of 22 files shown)


jasonschroeder-sfdc:
closing in favor of https://github.com/buildfarm/buildfarm/pull/2219

PR #2218, opened 2025-04-08
Provide digestFunction in fetch response

werkt:


jasonschroeder-sfdc (approved):
does this also need to set Status to `OK`?

werkt:
No, the default status value is OK. That's leveraged numerous other places

PR #2228, opened 2025-04-17
Make persistent workers respect tool input hash

jnowjack-lucidchart:
This fixes an issue where remote persistent workers didn't respect tool inputs changing. When making changes to tool inputs and running actions remotely they would run on old worker code or cause problems for actions on other builds. For example an error in some worker code being changed / tested was then causing errors across other branches without the bad code.


jasonschroeder-sfdc (commented):
Generally looks good, but I'd like to pull it and test it out (and also look at the code coverage gaps in persistentworkers in general, not specific to this PR)

PR #2229, opened 2025-04-22
Update jedis pool config to a more sensible value

80degreeswest:
Current value is much too high and can cause issues with maxing out connections, at which point the servers will reject new connections causing failures on the client side. Reduce the default value to a more reasonable setting.


jasonschroeder-sfdc (approved):
FWIW we're also using `200` in production cluster.

PR #2231, opened 2025-04-22
80degreeswest/update requeue config

80degreeswest:
Reduce requeue attempts to 3. Current value of 5 is too high - if action needs to requeue several times there is a a high probability of it having an issue which will prevent it from being queued.


jasonschroeder-sfdc (approved):
LGTM. We may (later) want to have some metrics around re-queued actions?

sulunemre:
Hi, I’m just wondering what are the possible reasons for requeuing and its failure? I occasionally see logs like this:
```
[WARNING] build.buildfarm.instance.shard.ServerInstance canOperationBeRequeued - Operation shard/executions/<EXEC_ID> not requeued.  Operation has been requeued too many times ( 8 > 5). 
```

PR #2239, opened 2025-04-28
Return persistent worker action output streams even when actions have missing outputs

JaredNeil:
The error path here would swallow the output streams and send back a [different error message](https://github.com/buildfarm/buildfarm/blob/main/src/main/java/build/buildfarm/worker/persistent/ProtoCoordinator.java#L196-L203).


jasonschroeder-sfdc (approved):

jasonschroeder-sfdc:
Hi @JaredNeil - if you could either merge/rebase, or allow maintainers to merge from main into this branch, I'd be happy to get this merged in the next 24hours. Thanks!

PR #2240, opened 2025-04-28
fix(config): prevent NPE when no parent config directory

tyler-french:
Currently, running:
```
bazel run //src/main/java/build/buildfarm/server -- server.yaml
```
will fail with the NPE:
```
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.toString()" because the return value of "java.nio.file.Path.getParent()" is null
	at build.buildfarm.common.config.BuildfarmConfigs.loadConfigs(BuildfarmConfigs.java:94)
	at build.buildfarm.common.config.BuildfarmConfigs.loadServerConfigs(BuildfarmConfigs.java:112)
	at build.buildfarm.server.BuildFarmServer.main(BuildFarmServer.java:267)
```
This was introduced in PR [#2219](https://github.com/buildfarm/buildfarm/pull/2219) which added support for the `!include` tag in YAML config files.

The issue occurs because:
1. The new `!include` feature requires resolving relative paths from the config file's parent directory
2. When running from the current directory with a relative path like `server.yaml`, the parent directory is null
3. The code was not handling this case, causing an NPE when calling toString() on the null parent

### DIFF
--- a/src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
+++ b/src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
@@ -88,6 +88,18 @@ public final class BuildfarmConfigs {
   }
 
   public static BuildfarmConfigs loadConfigs(Path configLocation) throws IOException {
+    Path parent = configLocation.getParent();
+    if (parent == null || !Files.isDirectory(parent)) {
+      log.info("Loading configs from single file: " + configLocation);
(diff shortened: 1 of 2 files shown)


80degreeswest (approved):
Ah good catch, thanks!

tyler-french:
> Ah good catch, thanks!

Hey! Just FYI this is still not working so I'm going to keep it as a draft. cc @werkt

tyler-french:
I should have something ready closer to EOD

PR #2248, opened 2025-04-29
feat(tools): Worker Profile Proxy

jasonschroeder-sfdc:
Proxy the `WorkerProfile` gRPC service from server->worker.

There's a new field in the gRPC service for the `workerName`. The worker does not validate it (for backwards compatibility with existing tools).

I'll open a PR against https://github.com/werkt/bf-client to utilize this as well.


werkt:
This change is a little bespoke, and I have a slightly different take on it that I've already implemented and wanted to push out. I'll ensure that we have a workerName in the response field, and the interface will be a batch one (send list of names for server to proxy responses).

PR #2249, opened 2025-05-01
Worker profiles

werkt:
Proxy through Servers
Populate ExecutedActionMetadata with WorkerExecutedMetadata


jasonschroeder-sfdc (approved):
LGTM. `src/main/java/build/buildfarm/tools/WorkerProfile.java` seems to go directly to the workers still, and doesn't make use yet of the proxy service.

PR #2269, opened 2025-06-05
build: remove rules_go

jasonschroeder-sfdc:
I can't remember why we need it, and everything passes without it.


werkt (approved):
It was necessary due to rules_docker iirc. Go for it

PR #2291, opened 2025-06-26
Made execOwner and execOwners work with persistent workers

jadenPete:
My attempt at fixing #2289. This PR modifies `ProtoCoordinator` to acquire a claim when creating a `PersistentWorker`, which is released when the worker is destroyed.

### DIFF
--- a/persistentworkers/src/main/java/persistent/bazel/BUILD
+++ b/persistentworkers/src/main/java/persistent/bazel/BUILD
@@ -8,7 +8,9 @@ java_library(
     deps = [
         "//persistentworkers/src/main/java/persistent/common:persistent-common",
         "//persistentworkers/src/main/protobuf:worker_protocol_java_proto",
+        "//src/main/java/build/buildfarm/common",
         "@maven//:com_github_pcj_google_options",
+        "@maven//:com_google_code_findbugs_jsr305",
         "@maven//:com_google_guava_guava",
         "@maven//:com_google_protobuf_protobuf_java",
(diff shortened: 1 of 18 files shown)


jadenPete:
One issue I'm still encountering is that the persistent worker is getting started under a different user (via the `as-nobody` wrapper) than the user who owns its inputs. Here's what the configuration for my worker looks like:
```yaml
worker:
    ...
    execOwners:
        ...
    executionPolicies:
        - executionWrapper:
              path: /app/build_buildfarm/as-nobody
              arguments:
                  - "-u"
                  - "<exec-owner>"
```

I suspect this is because I need to use `executionContext.claim` instead of acquiring a new claim, where `executionContext` is this:
https://github.com/buildfarm/buildfarm/blob/d0b05aeb977873a6cfa088c9d886858eaad8fd69/src/main/java/build/buildfarm/worker/Executor.java#L82

But I believe `executionContext.claim` is released after the _work request_ finishes, not the persistent worker. We could release it once the persistent worker finishes, but I'm not sure if that's the right move because we probably want most resources to be tied to the work request, not the persistent worker.

@werkt, do you have any thoughts? It looks like you've touched a lot of this code.

jadenPete:
After some discussion with @werkt, we think this calls for a different strategy: obtaining a persistent worker by factoring the UID into the key in addition to the action key. That way, we can select a persistent worker that's running under the user to which the claim in `ExecutionContext` belongs.

I'll close this PR and work on implementing that strategy.

PR #2293, opened 2025-07-01
fix(cfc): memory leak in LockMap

jasonschroeder-sfdc:
This fixes a leak in the LockMap static instance. It was only creating objects and olding on to them forever.

This changes the implementation to use a Caffeine Cache with a weakValues wrapper. When the `SharedLock()` is out-of-scope and GC'd, it can be removed from the cache.

Fixes: #2274


werkt (approved):
If this weren't legacy, I'd probably want a) a break out into a separate file for this static final class, and b) some tests to presume it does what we want, but I really want to get rid of this soon.

PR #2294, opened 2025-07-01
chore(worker): clean up confusing javadoc and Nullable annotations

jasonschroeder-sfdc:
- this method was marked @NotNull and @Nullable (AND NullableProblems)
- remove javadoc arg for parameter that doesn't exist


werkt (approved):
Yeesh, yes, good cleanup.

PR #2296, opened 2025-07-01
feat: support experimental_remote_downloader_propagate_credentials

jasonschroeder-sfdc:
Bazel 8 includes a new flag [--experimental_remote_downloader_propagate_credentials](https://bazel.build/versions/8.0.0/reference/command-line-reference?hl=en#flag--experimental_remote_downloader_propagate_credentials) which sends headers with a new prefix:

>The server implementation needs to support the new `http_header_url:<url-index>:<header-key>` qualifier where the `<url-index>` is a 0-based position of the URL inside the FetchBlobRequest's `uris` field. The URL-specific headers should take precedence over the global headers.


src/main/java/build/buildfarm/instance/server/NodeInstance.java
```diff
@@ -738,19 +741,57 @@ public ListenableFuture<Digest> fetchBlob(
...
+   * <p>If a header is prefixed with "<int>:", it is intended for the URL at index <int>. If a
+   * header is not prefixed, it is intended for all URLs. If a header is indexed, it is higher
+   * priority and will override existing headers with the same key.
+   *
+   * @param allHeaders
+   * @param urlIndex
+   * @return
+   */
+  @VisibleForTesting
+  static @Nonnull Map<String, String> getHeadersForUrlIndex(
```
werkt, line 757: Not sure I'm on board with the Nonnull specification - we've been Nullable-ing things that _could_ be null, but everything else is presumed (at least initially) to not be null

werkt (approved):

PR #2307, opened 2025-07-15
Update key expiration on get

80degreeswest:
Currently, when blobs are accessed, their expirations are not updated. This results in Redis expiring content when original expiration is reached, even if the key is still being accessed recently. This behavior, while as designed by Redis, introduces a major inefficiency. This change introduces an atomic getex operation, which updates expiration on key access.


jacobmou (approved):
The code change makes sense to me. The hot/frequent accessed key should be kept around by reset TTL every time they are accessed. 
A few things might want to do as follow up: 
1. the [configuration doc](https://github.com/buildfarm/buildfarm/blob/main/_site/docs/configuration/configuration.md) explicitly mentions that ``` The TTL maintained for ActionCache entries, not refreshed on getActionResult hit``` and ```The TTL maintained for CAS entries, which is not refreshed on any read access of the blob``` which might need to be updated. Also wondering if there are specific reasons we chose not to refresh TTL on `getActionResult` (@werkt do you happen to remember the rationale?)  and, if that's the case, we might want the TTL refreshing behavior configurable.
2. if we enable this TTL refreshing on every access, if we can log the remaining TTL whenever CAS have to expire cache entries to make room for new ones, we might be able to easily figure out if current CAS capacities is large enough. For example, if a CAS entry was accessed and expired within the same day, then the CAS size is definitely not large enough, but if a content was expired and only accessed a few day ago, then we probably are confident to reduce CAS capacity without losing much hot keys.

PR #2316, opened 2025-07-29
Make execOwner and execOwners work with persistent workers

jadenPete:
This is another attempt at #2289. It takes a different approach that was recommended by @werkt: adding the exec owner to `WorkerKey` and choosing the exec owner on the basis of the persistent workers that are currently alive (i.e. trying to choose an exec owner for which we already have a persistent worker).


persistentworkers/src/main/java/persistent/bazel/client/BasicWorkerKey.java
```diff
@@ -0,0 +1,107 @@
...
+    this.env = Preconditions.checkNotNull(env);
+    this.mnemonic = Preconditions.checkNotNull(mnemonic);
+    this.sandboxed = sandboxed;
+    this.cancellable = cancellable;
+
+    this.hash = calculateHashCode();
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
jasonschroeder-sfdc, line 66: Can we use a Lombok annotation instead?

persistentworkers/src/main/java/persistent/bazel/client/BasicWorkerKey.java
```diff
@@ -0,0 +1,107 @@
...
+ * WorkerKey} class as well. The difference is in addition to containing {@link BasicWorkerKey},
+ * {@link WorkerKey} also contains the exec owner (the user the worker will be run as) and the
+ * fields that are derived from it (namely, {@link build.buildfarm.common.config.ExecutionWrapper
+ * execution wrapper} arguments and the exec root). This sort of separation allows us to, given a
+ * {@link BasicWorkerKey}, preemptively select an exec owner for which we already have an alive
+ * worker with that key. That way, we avoid creating too many workers, which would happen by virtue
+ * of the fact that {@link WorkerKey} is the key of the {@link CommonsWorkerPool worker pool}.
+ */
+@ToString(onlyExplicitlyIncluded = true)
+public final class BasicWorkerKey {
```
jasonschroeder-sfdc, line 26: I see where you're going, I am wondering if we can find a better name than `BasicWorkerKey`

src/main/java/build/buildfarm/worker/filesystem/CFCExecFileSystem.java
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package build.buildfarm.worker;
+package build.buildfarm.worker.filesystem;
```
jasonschroeder-sfdc, line 15: Consider pulling this change into a separate PR (to be merged first?)

PR #2320, opened 2025-07-31
Apply fixes from copilot suggestions

80degreeswest:


werkt (changes requested):
This is a big old no. Copilot is not recognizing the intended thread-safety and concurrency implications of ConcurrentMap or LoadingCache, and synchronizing on things like storage and writes will crush performance.

PR #2346, opened 2025-09-07
Cleanup implicated google-java-format

werkt:


80degreeswest (approved):
Funny because that line was added as part of me running the formatter.

jasonschroeder-sfdc (approved):

PR #2368, opened 2025-09-12
Add worker tags configuration

80degreeswest:
A new worker tags will allow for tagging workers with arbitrary values. This will allow for assigning action execution to different groups or worker types, which will allow to easily gain visibility into which pool actions are being executed in without having to decipher ip addresses or hostnames. This will also be used by buildfarm admin to categorize workers, but could also be used for other metadata-like functionality. Multiple tags can be assigned to each worker.

Example:
```
"auxiliaryMetadata":[{"@type":"type.googleapis.com/build.buildfarm.v1test.WorkerExecutedMetadata","linkedInputDirectories":["build","external"],"tags":["cpu"]}]
```

### DIFF
--- a/src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
+++ b/src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
@@ -488,6 +488,7 @@ public class RedisShardBackplaneTest {
             .setFirstRegisteredAt(1703065913000L)
             .addTags("test-queue")
             .build();
+    assertThat(shardWorker.getTagsList()).containsExactly("test-queue");
     UnifiedJedis jedis = mock(UnifiedJedis.class);
     when(mockJedisClusterFactory.get()).thenReturn(jedis);
     when(jedis.hset(anyString(), anyString(), anyString())).thenReturn(1L);


src/test/java/build/buildfarm/instance/shard/RedisShardBackplaneTest.java
```diff
@@ -483,7 +483,11 @@ public void getDigestInsertTime() throws IOException {
   @Test
   public void testAddWorker() throws IOException {
     ShardWorker shardWorker =
-        ShardWorker.newBuilder().setWorkerType(3).setFirstRegisteredAt(1703065913000L).build();
+        ShardWorker.newBuilder()
+            .setWorkerType(3)
+            .setFirstRegisteredAt(1703065913000L)
+            .addTags("test-queue")
```
jasonschroeder-sfdc: can you add an assert for the tag?

werkt: The assert added verifies that Protobuf works, not that the worker stored in the backplane contains "test-queue", and in reality, this test does nothing at all with the tags - the only possible behavior verification would be on metrics population.

PR #2390, opened 2025-09-15
Recover from an ActionResult json print error

werkt:
Clients will upload results that contain execution_metadata that may contain unknown protobuf message types in the auxiliary_metadata repeated Any field. Purge unparseable objects and print warnings indicating what happened.


src/main/java/build/buildfarm/instance/shard/RedisShardBackplane.java
```diff
@@ -907,6 +909,7 @@ public static ActionResult parseActionResult(String json) {
       actionResultParser.merge(json, builder);
       return builder.build();
     } catch (InvalidProtocolBufferException e) {
+      e.printStackTrace();
```
jasonschroeder-sfdc, line 912: `logger.warning('Unexpected protobuf', e)` ?

PR #2397, opened 2025-10-02
Add dynamic option for reportResultStageWidth

JaredNeil:
This mirrors how `inputFetchStageWidth` works.


JaredNeil:
I didn't change the default to `0`, but happy to make that change if you think that's a better default.

jasonschroeder-sfdc (approved):

PR #2406, opened 2025-10-13
fix(ExecutionService): scoping of Context.current()

jasonschroeder-sfdc:
When running Buildfarm with opentelemetry-javaagent attached, occasionally remote exec gRPCs will get returned to remoteapi client (Bazel 8.4.1) with "Context cancelled without error" which halts the build.

This was only observable when opentelemetry-javaagent is attached. Without it, the issue is not reproducible.

It seems that there's some scoping/Context.current() issue when the javaagent has attached itself to grpc engine. This fix pins the context.

Co-Authored-by: Gemini 2.5 Pro


werkt (commented):
This should not be necessary. isCancelled is only called from members here. onSuccess and onFailure are only called in via the executor executor. The executor's current context should be fixed to the instance of the context you're acquiring in the change before the addCallback call already. There's no functional change to this request, and I don't see how it could affect something with otel...

werkt (approved):

PR #2410, opened 2025-10-16
chore(protobuf): remove/deprecate unused fields

jasonschroeder-sfdc:
- added `[deprecated = true]` for `BackplaneStatus.active_workers`
- removed unused fields that were never set.

### DIFF
--- a/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
+++ b/src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
@@ -452,8 +452,8 @@ message BackplaneStatus {
 
   DispatchedOperationsStatus dispatched_operations = 9;
 
-  // Maintained for backward compatibility.
-  repeated string active_workers = 4;
+  // Maintained for backward compatibility. Union of active_storage_workers and active_execute_workers
+  repeated string active_workers = 4 [deprecated = true];
 
   repeated string active_storage_workers = 12;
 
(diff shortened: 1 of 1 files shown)


src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -450,24 +450,21 @@ message BackplaneStatus {
...
-  repeated string active_execute_workers = 13;
-
-  int64 cas_lookup_size = 5;
-
-  int64 action_cache_size = 6;
+  reserved 5;
+  reserved 6;
+  reserved 7;
+  reserved 8;
+  reserved 10;
```
werkt: You can reserve a range of field sequence numbers: https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto#L608, and let's get a comment in there

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -450,24 +450,21 @@ message BackplaneStatus {
...
-  int64 blocked_invocations_size = 8;
+  // Count of dispatched operations (non-negative)
+  int64 dispatched_size = 11;
 
-  int64 fetch_time_ms = 10;
+  repeated string active_storage_workers = 12;
 
-  int64 dispatched_size = 11;
+  repeated string active_execute_workers = 13;
 }
```
werkt: the ordering is a preference to group items together, and certainly muddies up this change. Putting things in seq is something I'm not terribly fond of, but I recognize it makes it tricky to get the next one.

jasonschroeder-sfdc: OK. I put it all the deprecations at the end.

PR #2437, opened 2025-11-09
bf-diff - compare two invocations

werkt:
Align operations between builds via output paths across actions, input roots, etc.
Note asymmetries in output

### DIFF
--- a/src/main/java/build/buildfarm/tools/Diff.java
+++ b/src/main/java/build/buildfarm/tools/Diff.java
@@ -1,0 +1,597 @@
+// Copyright 2025 The Buildfarm Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


src/main/java/build/buildfarm/tools/Diff.java
```diff
@@ -0,0 +1,564 @@
...
+    String instanceName = args[1];
+    String invocationIdA = args[2];
+    String invocationIdB = args[3];
+
+    ManagedChannel channel = createChannel(host);
+    Instance instance = new StubInstance(instanceName, "bf-diff", channel, Durations.fromHours(10));
+    try {
+      diff(instance, invocationIdA, invocationIdB);
+    } finally {
+      instance.stop();
```
jasonschroeder-sfdc:
```suggestion
      instance.stop();
      channel.shutdown();
      channel.awaitTermination(1, SECONDS);
    }
```

src/main/java/build/buildfarm/tools/Diff.java
```diff
@@ -0,0 +1,564 @@
...
+    List<Operation> opsA = loadOrListOps(instance, invocationIdA);
+    List<Operation> opsB = loadOrListOps(instance, invocationIdB);
+
+    if (opsA.isEmpty()) {
+      return;
+    }
+
+    DigestFunction.Value digestFunction =
+        executeOperationMetadata(opsA.get(0).getMetadata()).getDigestFunction();
+    DigestUtil digestUtil = new DigestUtil(HashFunction.get(digestFunction));
```
jasonschroeder-sfdc: Warn/abort if the digest function of A & B doesn't match?

jasonschroeder-sfdc (approved):

PR #2498, opened 2026-02-27
fix: bind jedis-mock to explicit address in RedisShardSubscriptionTest

jasonschroeder-sfdc:
The no-arg RedisServer.newRedisServer() binds to a default address whose getHost() returns "localhost". On macOS, the Jedis client may resolve "localhost" to ::1 (IPv6) while the server socket is IPv4-only, causing the client to connect to a different protocol stack. This results in an infinite reconnect loop and a 60-second test timeout.

Bind explicitly via newRedisServer(0, InetAddress.getByName("localhost")) to match the pattern used by the passing JedisCasWorkerMapTest, and add @After tearDown to stop servers between tests.

### DIFF
--- a/src/test/java/build/buildfarm/instance/shard/RedisShardSubscriptionTest.java
+++ b/src/test/java/build/buildfarm/instance/shard/RedisShardSubscriptionTest.java
@@ -14,6 +14,7 @@
 
 package build.buildfarm.instance.shard;
 
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.truth.Truth.assertThat;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.doAnswer;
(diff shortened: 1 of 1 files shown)


src/test/java/build/buildfarm/instance/shard/RedisShardSubscriptionTest.java
```diff
@@ -55,14 +60,45 @@ RedisShardSubscriber getDefaultRedisSubscriber() {
...
   }
 
+  /**
+   * Creates a jedis-mock server bound explicitly to the IPv4 loopback address. Using the no-arg
+   * {@code newRedisServer()} causes {@code getHost()} to return "localhost", which on macOS may
+   * resolve to {@code ::1} (IPv6) while the server socket is IPv4-only, resulting in the Jedis
+   * client connecting to a different protocol stack than the server is listening on.
+   */
+  private RedisServer startServer() throws IOException {
+    server = RedisServer.newRedisServer(0, InetAddress.getByName("localhost")).start();
```
werkt: nit: please add `/* port=*/` convention here, as the parameter is magically defined

PR #2499, opened 2026-03-03
feat: Leader Election for dispatched monitor

### LINKED ISSUE #2434: Excessive DispatchedMonitor resources for a busy cluster
Every server instance, by default, starts a DispatchedMonitor with a 1sec interval. For a busy cluster, with lots of dispatched operations and a lot (>20) server pods, this can be a lot of resources.

Thoughts:
- can one (or two or three) Servers run the DispatchedMonitor via a leader election algorithm?
 
 
Workaround:
- reduce the frequency (linearly) for each Server. i.e. for 20 servers, reduce frequency to every DispatchedMonitor running every 20 seconds (per Server). See `dispatchedMonitorIntervalSeconds` in the config.

jasonschroeder-sfdc:
When "many" buildfarm-servers are running, they are all doing DispatchedMonitor quite frequently. With a large cluster, this is taxing when "many" operations are dispatched.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/LeaderElection.java
+++ b/src/main/java/build/buildfarm/common/redis/LeaderElection.java
@@ -165,8 +165,7 @@ public class LeaderElection implements Closeable {
     this.leaseDuration = leaseDuration;
     this.renewalInterval = renewalInterval;
     this.renewalExecutor =
-        Executors.newScheduledThreadPool(
-            1,
+        Executors.newSingleThreadScheduledExecutor(
             r -> {
               Thread thread = new Thread(r, "leader-election-renewal");
               thread.setDaemon(true);
--- a/src/main/java/build/buildfarm/common/redis/LeaderElection.java
+++ b/src/main/java/build/buildfarm/common/redis/LeaderElection.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
+import java.util.logging.Level;
 import lombok.extern.java.Log;
 import redis.clients.jedis.UnifiedJedis;
 import redis.clients.jedis.params.SetParams;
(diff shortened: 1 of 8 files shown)


werkt (changes requested):

src/main/java/build/buildfarm/common/redis/LeaderElection.java
```diff
@@ -0,0 +1,584 @@
...
+            String.format(
+                "Node '%s' failed to become leader for '%s' (already held by another node)",
+                nodeId, electionKey));
+      }
+
+      return acquired;
+    } catch (Exception e) {
+      log.severe(
+          String.format(
+              "Error attempting to become leader for '%s': %s", electionKey, e.getMessage()));
```
werkt: drop the getMessage() here and put `log.log(Level.SEVERE, "...", e)`, assuming this is a genuinely exceptional case.

Also consider whether Exception is the right guard for this, that is very broad, and there may be something that we want the caller to handle, or at least be aware of, that indicates significant failure. At least translate non-IO exceptions into IO-exceptions here per the rest of the redis interface

werkt (commented):
otel is concerning for individual introduction - I'd like it more broadly and in dedicated PRs.
Some minor nits and other questions, but this looks good generally. I'll do another pass in on this with some changes.

src/test/java/build/buildfarm/common/redis/BUILD
```diff
@@ -113,3 +114,15 @@ java_test(
     test_class = "build.buildfarm.AllTests",
     deps = COMMON_DEPS,
 )
+
+java_test(
+    name = "leaderelection-redis",
+    size = "small",
+    srcs = ["LeaderElectionTest.java"],
+    tags = [
+        "exclusive",
```
werkt, line 123: why exclusive?

src/main/java/build/buildfarm/common/redis/LeaderElection.java
```diff
@@ -0,0 +1,581 @@
...
+ * election.close();
+ *
+ * }</pre>
+ */
+@Log
+public class LeaderElection implements Closeable {
+  private static final String REDIS_KEY_PREFIX = "buildfarm:leader:";
+
+  // Lua script for atomic renewal - checks ownership before extending TTL
+  private static final String RENEWAL_SCRIPT =
```
werkt, line 77: Put this into a static function that applies the precondition wrapper and takes the action params to generate the lua string.

PR #2514, opened 2026-03-16
Match Worker property with specified name

werkt:
In addition to endpoint and identifier


src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -737,9 +738,13 @@ public double get() {
       writer = new LocalCasWriter(execFileSystem);
     }
 
+    String endpointName = configs.getWorker().getPublicName();
```
jasonschroeder-sfdc: Consider also "endpointLabel"

PR #2538, opened 2026-03-31
fix: use UnifiedJedis instead of raw Jedis connections in queue layer

jasonschroeder-sfdc:
When Redis Cluster topology changes (slot migrations, failovers), BalancedRedisQueue crashes with JedisMovedDataException because getJedisFromKey() extracts raw Connection objects from JedisCluster and wraps them in plain Jedis instances. Plain Jedis does not handle MOVED redirections — only JedisCluster (via UnifiedJedis) does this transparently with automatic retries.

This commit removes the raw connection extraction entirely and passes the UnifiedJedis client straight through to RedisQueue and RedisPriorityQueue. JedisCluster handles MOVED redirections, slot cache refresh, and retries internally.

Changes:
- QueueDecorator: decorate(Jedis, String) → decorate(UnifiedJedis, String)
- RedisQueue: field, constructors, and decorate() accept UnifiedJedis
- RedisPriorityQueue: same — field, all constructors, decorate()
- TranslatedQueueDecorator: decorate() parameter updated
- BalancedRedisQueue: deleted getJedisFromKey() method; removed all try-with-resource Jedis blocks from offer, removeFromDequeue, take, takeAny, pollAny, size, visit, visitDequeue, and scan methods; removed Connection/JedisCluster/JedisPooled/JedisClusterCRC16 imports
- Tests: mock tests use @Mock UnifiedJedis (Jedis does not extend UnifiedJedis in Jedis 7.x); integration tests pass JedisPooled (which does extend UnifiedJedis) to queue constructors while keeping a raw Jedis for test setup/teardown operations

Trade-off — blocking take interruption:
The old code used jedis::disconnect to immediately cancel blocking blmove calls during thread interruption (e.g. worker shutdown). With UnifiedJedis, the internal connection is not directly accessible for disconnect. Blocking blmove calls will now wait for their natural timeout (1–8 seconds via exponential backoff) instead of being cancelled instantly. This is acceptable because the timeout is bounded and worker shutdown already has grace periods — and is far better than crashing with unhandled JedisMovedDataException.

### DIFF
--- a/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
+++ b/src/main/java/build/buildfarm/common/redis/BalancedRedisQueue.java
@@ -37,13 +37,8 @@ import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import lombok.Getter;
 import redis.clients.jedis.AbstractPipeline;
-import redis.clients.jedis.Connection;
-import redis.clients.jedis.Jedis;
-import redis.clients.jedis.JedisCluster;
-import redis.clients.jedis.JedisPooled;
 import redis.clients.jedis.UnifiedJedis;
 import redis.clients.jedis.resps.ScanResult;
-import redis.clients.jedis.util.JedisClusterCRC16;
 
 /**
  * @class BalancedRedisQueue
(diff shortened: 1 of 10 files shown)


jasonschroeder-sfdc:
We can wait for the resiliency, I just don't know the timeline for that.

PR #2552, opened 2026-04-08
refactor: replace FileInputStream/FileOutputStream with Files.newInpu…

jasonschroeder-sfdc:
…tStream/newOutputStream

Enables the PMD AvoidFileStream rule and fixes all 6 violations. The legacy File stream classes have a finalizer that adds GC overhead; the NIO alternatives avoid this.

### DIFF
--- a/src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
+++ b/src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
@@ -54,6 +54,8 @@ public final class BuildfarmConfigs {
         return yaml.load(input);
       } catch (NoSuchFileException ex) {
         throw new RuntimeException("Could not find config file: " + value);
+      } catch (IOException ex) {
+        throw new RuntimeException("Could not read config file: " + value, ex);
       }
     }
   }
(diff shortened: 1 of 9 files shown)


src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
```diff
@@ -51,11 +49,13 @@ public Object construct(Node node) {
       final ScalarNode scalarNode = (ScalarNode) node;
       final String value = scalarNode.getValue();
       try {
```
werkt: This needs a try-with-resources. The input is never closed...

src/main/java/build/buildfarm/common/io/Utils.java
werkt: this should be converted to take a Path instead of a File. The only caller of this already does a `new File(path.toString())`...

werkt (changes requested):

src/main/java/build/buildfarm/common/config/BuildfarmConfigs.java
jasonschroeder-sfdc: Fixed — wrapped the include InputStream in try-with-resources.

src/main/java/build/buildfarm/common/io/Utils.java
jasonschroeder-sfdc: Fixed — unTar now takes a Path, and the caller passes its existing Path directly instead of re-wrapping in a File.

werkt (approved):

PR #2556, opened 2026-04-13
fix: action deduplication bypass in mergeOrSchedule loop

alex-torok:
Fix a bug in ServerInstance.mergeOrSchedule() where the `ignoreMerge` flag was incorrectly set to `true` on the very first scheduling attempt, effectively disabling action deduplication for concurrent Execute requests. We regularly saw identical action digest hashes be scheduled and executed.

Full disclosure: This issue was found and fixed by claude code. I have not yet tested it in our buildfarm deployment, but the logic seems pretty straight forward to follow, so I figured I'd open this PR now. I'll update this PR as we test and evaluate in our deployment.

## Root cause

Line 2431 had:

    ignoreMerge = ignoreMerge || lookupAttempts != 0;

The while loop condition `lookupAttempts-- != 0` decrements *before* the loop body runs, so on the first iteration `lookupAttempts` is already 4 (decremented from 5). Since `4 != 0` is true, `ignoreMerge` was set to `true` immediately.

When `ignoreMerge=true` is passed to `schedule()` -> `prequeue()`, the clause in RedisShardBackplane.prequeue():

    if (state.executions.create(...) || ignoreMerge) { ... }

allows the operation onto the prequeue even when `create()` returns false (i.e., another server already claimed this action key via SET NX). This defeats the atomic Redis SET NX dedup protection.

## How duplicates occur with multiple server replicas

With N server replicas behind a gRPC load balancer, two Execute RPCs for the same action digest can hit different servers nearly simultaneously:

    Server A (replica 1)                    Server B (replica 7)
    ─────────────────                       ─────────────────
    Execute(action=abc123)                  Execute(action=abc123)
      │                                       │
      ├─ mergeExecution() → null              ├─ mergeExecution() → null
      │  (nothing in Redis yet)               │  (nothing in Redis yet)
      │                                       │
      ├─ ignoreMerge = true  (BUG)            ├─ ignoreMerge = true  (BUG)
      │                                       │
      ├─ schedule(ignoreMerge=true)           ├─ schedule(ignoreMerge=true)
      │  ├─ SET NX Action:abc123 → OK         │  ├─ SET NX Action:abc123 → FAIL
      │  ├─ prequeue.offer() → queued         │  ├─ || ignoreMerge → TRUE
      │                                       │  ├─ prequeue.offer() → DUPLICATE
      │                                       │
      ▼                                       ▼
      Worker X executes action                Worker Y executes SAME action

The correct behavior (after fix) is: Server B's prequeue() returns false because create() fails and ignoreMerge is false, schedule() returns null, and the retry loop calls mergeExecution() again — which now finds Server A's operation and merges onto it.

## Fix

Change `!= 0` to `== 0` so that `ignoreMerge` is only forced to `true` on the *last* retry attempt (when lookupAttempts has been exhausted), as a fallback to avoid returning RESOURCE_EXHAUSTED. The first 4 iterations now properly rely on the SET NX atomic check for deduplication.

## Test

Add `executeRetriesMergeWhenPrequeueLosesRace` to ServerInstanceTest which simulates the race: mergeExecution returns null, prequeue returns false (NX race lost), then verifies the retry merges onto the winner's operation instead of forcing through with ignoreMerge=true.

### DIFF
--- a/src/main/java/build/buildfarm/instance/shard/ServerInstance.java
+++ b/src/main/java/build/buildfarm/instance/shard/ServerInstance.java
@@ -2427,7 +2427,7 @@ public class ServerInstance extends NodeInstance {
           return immediateFailedFuture(
               Status.RESOURCE_EXHAUSTED.withDescription("Too many jobs pending").asException());
         }
-        ignoreMerge = ignoreMerge || lookupAttempts != 0;
+        ignoreMerge = ignoreMerge || lookupAttempts == 0;
         execution =
             schedule(
                 actionDigest,
--- a/src/test/java/build/buildfarm/instance/shard/ServerInstanceTest.java
+++ b/src/test/java/build/buildfarm/instance/shard/ServerInstanceTest.java
@@ -862,7 +862,8 @@ public class ServerInstanceTest {
         .prequeue(any(ExecuteEntry.class), any(Operation.class), eq(true));
     // Verify the second iteration merged onto the winner's execution.
     verify(mockBackplane, times(2)).mergeExecution(actionKey);
-    verify(mockBackplane, times(1)).watchExecution(eq(winnerExecution.getName()), any(Watcher.class));
+    verify(mockBackplane, times(1))
+        .watchExecution(eq(winnerExecution.getName()), any(Watcher.class));
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


werkt:
Fix looks good, run the formatter and we can merge. Tell your buddy Claude nice find.

jjudd:
We cherry picked this onto our Buildfarm fork and saw a big improvement to action merging. We would previously only see action merging be a fraction of a %. Now it spikes up into the single digit % pretty regularly. Sometimes it even make it into the double digits.

Thanks for the fix!

jasonschroeder-sfdc (approved):

PR #2561, opened 2026-05-01
Make action cache size configurable and set default equal to other server caches

jjudd:
After the action merging fix from #2556 was applied to our Buildfarm cluster we started running into OOMs on our Buildfarm servers. We were seeing a heap usage pattern akin to a memory leak. A heap dump showed that the `ShardActionCache` was using roughly 18 GB of heap with ~921k entries. The cache never hit its max entries of 1,000,000, so memory usage kept climbing.

Best I can tell when an action merge occurs a cache invalidation for the action cache doesn't occur because `schedule()` isn't called. Looking at heap usage from before the action merging fix it looks like we were seeing a similar problem. However, it was happening much more slowly because action merging was happening far less often.

This change makes the `ShardActionCache` size configurable and sets its default to the same size as the other server caches.

This change also adds a handful of `ShardActionCache` tests.

We've deployed this change to our Buildfarm cluster and it resolved our issue. Action merging has continued working with this change applied.

Here's JVM heap usage on our servers before this change with the problematic heap usage:
image

Here's JVM heap usage with this change deployed to one server. The yellow line has the change applied and its heap looks healthy:
image

Here's the JVM heap usage on our servers with this change deployed to all servers. The heap usage no longer is problematic:
image

### DIFF
--- a/src/main/java/build/buildfarm/common/config/ServerCacheConfigs.java
+++ b/src/main/java/build/buildfarm/common/config/ServerCacheConfigs.java
@@ -52,4 +52,11 @@ public class ServerCacheConfigs {
    * @details This will not dictate the max memory used.
    */
   private long recentServedExecutionsCacheMaxEntries = 64 * 1024;
+
+  /**
+   * @field actionCacheMaxEntries
+   * @brief The max number of entries that the action cache will hold.
+   * @details This will not dictate the max memory used.
+   */
+  private long actionCacheMaxEntries = 1_000_000;
 }
--- a/src/test/java/build/buildfarm/actioncache/ShardActionCacheTest.java
+++ b/src/test/java/build/buildfarm/actioncache/ShardActionCacheTest.java
@@ -1,0 +1,117 @@
+// Copyright 2026 The Buildfarm Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.actioncache;
+
+import static com.google.common.truth.Truth.assertThat;
+import static com.google.common.util.concurrent.MoreExecutors.newDirectExecutorService;
+import static org.mockito.ArgumentMatchers.any;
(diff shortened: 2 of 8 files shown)


src/main/java/build/buildfarm/common/config/ServerCacheConfigs.java
```diff
@@ -52,4 +52,11 @@ public class ServerCacheConfigs {
    * @details This will not dictate the max memory used.
    */
   private long recentServedExecutionsCacheMaxEntries = 64 * 1024;
+
+  /**
+   * @field actionCacheMaxEntries
+   * @brief The max number of entries that the action cache will hold.
+   * @details This will not dictate the max memory used.
+   */
+  private long actionCacheMaxEntries = 64 * 1024;
```
werkt: Please make this (and consistent with the docs) 1,000,000 so as to not change the default value in the interest of not changing behavior silently.

src/main/java/build/buildfarm/tools/Cat.java
```diff
@@ -126,7 +126,6 @@
       Cat.CatWriteStatus.class,
     })
 class Cat implements Callable<Integer> {
-
```
werkt, line 129: unrelated (format.sh disagreement with enforcement). Please discard and it will be brought in separately.

werkt (changes requested):

src/test/java/build/buildfarm/actioncache/ShardActionCacheTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Mock private Backplane backplane;
+
+  @Before
+  public void setUp() {
+    MockitoAnnotations.initMocks(this);
+  }
+
+  @Test
+  public void get_cacheMiss_loadsFromBackplaneAndCachesResult() throws Exception {
```
werkt: no underscores in function names please
prefer test naming [interaction][sS]hould[Not]PerformActivityUnderConditions.

src/main/java/build/buildfarm/common/config/ServerCacheConfigs.java
jjudd: Done. Updated this to 1,000,000.

src/main/java/build/buildfarm/tools/Cat.java
jjudd: Done. Dropped this from the PR.

src/test/java/build/buildfarm/actioncache/ShardActionCacheTest.java
jjudd: Done. Updated all test names to follow this test naming convention.

jjudd:
Thanks for the review. Much appreciated.

I made all of the requested changes and believe this is ready for another pass.

werkt (approved):

PR #2564, opened 2026-05-15
Ignore output_paths specified as linkable input dirs

### LINKED ISSUE #2562: bazel + REAPI ≥ 2.1 + linkInputDirectories: true leads to output directories that point to a shared empty-directory CAS node (AI-assisted)
(The issue was mostly diagnosed by Claude, but it started from a real failure with my build and I spent quite a bit of time trying to understand it myself too).

It looks like the combination of bazel, REAPI ≥ 2.1 + linkInputDirectories: true leads to a situation where empty _output_ directories are created as symlinks to input CAS empty directory nodes. When actions write to those directories the outputs all end up in the same directory. I ran into this when trying to enabled linkInputDirectories for an android build that unpacks aars into various subdirectories.

Excluding blaze-out from linkedInputDirectories or turning off linkInputDirectories fixes the issue.

Repro and more explanation in https://github.com/mikaoura/buildfarm-linkinputdirectories-repro

The flow is (this part written by Claude but I tried to verify it myself):
1. **Bazel side.** For each tree-artifact output, Bazel adds an
   `EmptyInputDirectory` entry to the action's input root MerkleTree so the
   leaf exists on disk before the action runs:
   [`MerkleTreeComputer.java:292-298`](https://github.com/bazelbuild/bazel/blob/main/src/main/java/com/google/devtools/build/lib/remote/merkletree/MerkleTreeComputer.java).
   All such placeholders are at paths under `bazel-out/...`.
2. **Bazel side.** When the server advertises REAPI ≥ 2.1 (BuildFarm does),
   Bazel sends outputs in the flat `Command.output_paths` field rather than
   the older `output_files` + `output_directories` pair. The flat field
   doesn't distinguish file vs. directory outputs:
   [`RemoteExecutionService.java:487`](https://github.com/bazelbuild/bazel/blob/main/src/main/java/com/google/devtools/build/lib/remote/RemoteExecutionService.java).
3. **Worker side.** `CFCLinkExecFileSystem.createOutputDirectory` treats
   every `output_paths` entry as a file path (`files = output_paths; dirs =
   []`) — source comment: *"output paths require the action to create their
   own directory"*:
   [`CFCLinkExecFileSystem.java:230-236`](https://github.com/buildfarm/buildfarm/blob/main/src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java).
   The `OutputDirectory` tracker therefore stops at the *parent* of each
   output path (e.g. `_aar/unzipped/resources` — gets stamped as a real dir);
   the actual tree-artifact leaf (e.g. `lib_foo`) is **not** tracked.
4. **Worker side.** `LinkExecFileVisitor.preVisitDirectory` walks the input
   root. At the un-tracked leaf, `outputDirectory == null && linkedDirectories.contains(dir)`
   is true (because Bazel placed an empty Directory placeholder there and
   the default `linkedInputDirectories` regex `^(?!external$).*$` matches
   anything outside `external/`), so it calls `linkDirectory(dir, digest,
   …)` which symlinks `lib_foo` to `<cache>/<digest>_dir`. The placeholder
   is empty ⇒ digest is the canonical empty-Directory digest ⇒ **every
   empty tree-artifact leaf across every concurrent action symlinks to the
   same physical inode** at `/tmp/worker/cache/e3b0c4..._dir`.
5. **Consequence.** The action writes into its declared output_directories;
   all writes land in the shared inode. The worker then walks each declared
   output path to compute its result Tree, sees the merged content, and
   reports the same Tree digest for every output_directory. Downstream
   consumers fetching any of those digests see the union — and the cached
   `e3b0c4...` entry is now poisoned for every subsequent action.

werkt:
Filtering on linkedDirectory iteration for output_paths
Ignores output_paths and all subdirectories
Simplify createExecDir signature - no action

Fixes #2562

### DIFF
--- a/src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
+++ b/src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
@@ -22,7 +22,6 @@ import static com.google.common.util.concurrent.Futures.transformAsync;
 import static com.google.common.util.concurrent.MoreExecutors.listeningDecorator;
 import static java.util.Collections.synchronizedList;
 
-import build.bazel.remote.execution.v2.Action;
 import build.bazel.remote.execution.v2.Command;
 import build.bazel.remote.execution.v2.DigestFunction;
 import build.bazel.remote.execution.v2.Directory;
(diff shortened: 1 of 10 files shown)


src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
```diff
@@ -373,11 +404,15 @@ public Path createExecDir(
     }
     Files.createDirectories(execDir);
 
+    // output_paths should never be linked themselves, and their output path chain
```
jasonschroeder-sfdc, line 399: >output_paths should never be linked themselves
should we log/throw an error in this situation?

jasonschroeder-sfdc (approved):

src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
werkt: No, the statement is not [a reflection of] a validation requirement (those would be handled by servers during action validation).
This is a 'we should never link output_paths' - if the output path itself exists as an input directory it should not be symlinked, 'and their output path chain will terminate at their parent' means that they can appear, with guidance from linkedInputDirectories.
Essentially all of this is valid, and this constructs the mechanism by which we avoid the behavior. I don't think we need an error to guard it afterwards, just the test should be sufficient that we get the behavior we expect.

PR #2566, opened 2026-05-28
fix(cas): make `hexBucketLevels` work with BLAKE3 digests

ankitpatisfdc:
`HexBucketEntryPathStrategy.getPath` rejected every CAS key whose hex hash didn't sit at offset 0, throwing `IllegalStateException` from the `checkState` on line 31. The pattern matched the start of the *whole* key against `[0-9a-f]{levels*2}`, which works for the digest functions in `OMITTED_DIGEST_FUNCTIONS` (MD5, SHA1, SHA-2 family) — those keys are bare hex hashes — but not for BLAKE3, whose keys are prefixed with `blake3_` per `digestFilename`.

Result: `hexBucketLevels > 0` was unusable with BLAKE3 clients; every read or write failed before reaching disk.

Allow the pattern to match an optional `<digestfn>_` prefix, capture the hex portion in group 1, and bucket on `Matcher.start(1)`. Group capture is correct where character-level heuristics aren't: testing `isHexChar(key.charAt(0))` is fooled by `blake3_…`, since `b` is in `[a-f]`, and silently routes BLAKE3 writes to nonexistent `<root>/bl/ak/…` directories.

The full key remains the leaf filename, so blobs of identical hex hash but different digest functions don't collide.

Adds tests for the BLAKE3 prefix (chosen so its first character is also hex, to guard against the heuristic regression) and the trailing `_exec` suffix on SHA-family keys.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -406,7 +406,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     Entry entry = getEntry(key);
     if (entry != null && entry.referenceCount < 0) {
       try {
-        entry = new Entry(key, Files.size(getPath(key)), null);
+        entry = new Entry(key, Files.size(getPath(digest, key)), null);
       } catch (IOException e) {
         return false;
       }
@@ -475,7 +475,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
       return true;
     }
 
-    if (Files.exists(getPath(e.key))) {
+    Digest digest = entryDigest(e.key, e.size);
+    if (digest != null && Files.exists(getPath(digest, e.key))) {
       e.existsDeadline = Deadline.after(10, SECONDS);
       return true;
     }
@@ -572,7 +573,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
       if (e != null) {
         InputStream input = null;
         try {
-          input = compressorInputStream(compressor, Files.newInputStream(getPath(key)));
+          input = compressorInputStream(compressor, Files.newInputStream(getPath(digest, key)));
           input.skip(offset);
         } catch (IOException ioEx) {
           if (!(ioEx instanceof NoSuchFileException)) {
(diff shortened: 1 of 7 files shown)


jasonschroeder-sfdc (commented):
LGTM. Not sure about the performance of a Regex here. Can we be more clever?

ankitpatisfdc:
@jasonschroeder-sfdc I got Claude to replace it with a more performant non-regex alternative. More details in the latest commit message.

ankitpatisfdc:
The failing check is due to a Buildkite error:
```
Error in download_and_extract: java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_kotlin/releases/download/v1.9.6/rules_kotlin-v1.9.6.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/rules_kotlin+/temp14878704322085006727/rules_kotlin-v1.9.6.tar.gz: GET returned 502 Bad Gateway or Proxy Error
(09:39:52) ERROR: Analysis of target '//:buildifier' failed; build aborted: java.io.IOException: Error downloading [https://github.com/bazelbuild/rules_kotlin/releases/download/v1.9.6/rules_kotlin-v1.9.6.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/rules_kotlin+/temp14878704322085006727/rules_kotlin-v1.9.6.tar.gz: GET returned 502 Bad Gateway or Proxy Error
(09:39:52) INFO: Elapsed time: 12.623s, Critical Path: 0.05s
(09:39:52) INFO: 3 processes: 3 internal.
(09:39:52) ERROR: Build did NOT complete successfully
(09:39:53) INFO: Streaming build results to: https://btx.cloud.google.com/invocations/04180502-5811-4785-a2eb-1276bc53e375
```

werkt:
I don't think we should be treating these as strings - they're computed from the inside out. Drop the complicated run treatment and just switch it to buildfarm's digest (with digestFunction)

werkt (approved):

PR #2575, opened 2026-06-16
Null-out Android env for non-Android macos builds

ted-xie:
The Intel `macos` deployment in BazelCI does not have the latest Android build tools installed. By default, BazelCI adds `ANDROID*` environment variables to builds, which registers an Android toolchain for any transitively-loaded Android deps, which causes a repo rule-level error.

To prevent this, this PR nulls out the `ANDROID*` environment variables.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -74,6 +74,9 @@ tasks:
     - '...'
   macos:
     name: Unit Tests
+    environment:
+      ANDROID_HOME: ""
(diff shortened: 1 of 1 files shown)


ted-xie:
NOTE: It is also possible to avoid this error by swapping `macos` to `macos_arm64` in presubmit.yml. The reason  why `macos` on BazelCI supports newer SDKs than `macos_arm64` is due to an implementation detail of an internal system.

werkt (approved):
Very well. Thanks.

PR #2576, opened 2026-06-17
Specify core count with bfssh

werkt:


src/main/java/build/buildfarm/tools/Bfssh.java
```diff
@@ -255,6 +262,9 @@ private static Iterable<ListenableFuture<ExecuteResponse>> runOnWorkers(
       if (!worker.isEmpty()) {
         platform.addPropertiesBuilder().setName("Worker").setValue(worker);
       }
+      if (cores > 1) {
+        platform.addPropertiesBuilder().setName("cores").setValue(Integer.toString(cores));
+      }
```
jasonschroeder-sfdc: consider setting it even if it == 1. It might not default to 1 (in the RBE cluster) for everybody :) Or use `0` as the guard value.

jasonschroeder-sfdc (approved):

PR #2578, opened 2026-06-17
Initial CPU Farmer's Market

werkt:
Instead of slots of single core CPU usage, provide a marketplace for executions to buy and sell the CPU they [think they] need or observably, don't.
Initialize actions with a single core of cpu request. Executions will not sell below 1/10th of a single core - this is mostly for stability reasons, but is also a limitation of cgroups.

Extend deadline for execution polls

Attempting to extend deadlines for market tolerance leaves the poll deadline based on original timeout ready to kill processes. This deadline seems unnecessary, and should live closer to the input fetcher.

Wallet is represented by a claimed resource for an execution, as safe as that system for preventing leaks. Resources are reworked to emphasize this capacity for mutation.

### DIFF
--- a/src/main/java/build/buildfarm/common/ExecutionProperties.java
+++ b/src/main/java/build/buildfarm/common/ExecutionProperties.java
@@ -297,4 +297,44 @@ public class ExecutionProperties {
    * @details See https://github.com/bazelbuild/bazel/issues/10091
    */
   public static final String PERSISTENT_WORKER_COMMAND = "persistentWorkerCommand";
+
+  /**
+   * @field CPU_SHARE_FLOOR
+   * @brief minimum number of cpu shares an execution must own
+   * @details Prevent available cpu from being reduced beyond this amount Must not be <= 0 for
+   *     market interactions
+   */
+  public static final String CPU_SHARE_FLOOR = "cpuShareFloor";
+
+  /**
+   * @field PCT_MIN_UNUSED
+   * @brief Percentage of owned shares that must be unused before sale
+   * @details Headroom for per-sample volatility of use before its cpu may be sold Must not be <= 0.
+   *     Values should be between <code>0 < x < 1</code> for market interactions
+   */
+  public static final String PCT_MIN_UNUSED = "pctMinUnused";
+
+  /**
+   * @field PCT_MIN_THROTTLE
+   * @brief Percentage of throttle that must be observed before buying cpu shares
+   * @details Headroom for per-sample volatility of throttle before cpu may be bought Must not be <=
+   *     0 for market interactions
+   */
+  public static final String PCT_MIN_THROTTLED = "pctMinThrottled";
+
+  /**
+   * @field MIN_SHARES_SOLD
+   * @brief CPU Shares minimum threshold sold
+   * @details Headroom for per-sample volatility of use before its cpu may be sold Must not be <= 0
+   *     for market interactions
+   */
+  public static final String MIN_SHARES_SOLD = "minSharesSold";
+
+  /**
+   * @field MAX_SHARES_SOLD
+   * @brief CPU Shares maximum sold clamp
+   * @details Headroom for per-sample volatility of unuse that may be sold Must not be <= 0 for
+   *     market interactions
+   */
+  public static final String MAX_SHARES_SOLD = "maxSharesSold";
 }
--- a/src/main/java/build/buildfarm/worker/CPULease.java
+++ b/src/main/java/build/buildfarm/worker/CPULease.java
@@ -1,0 +1,66 @@
+// Copyright 2026 The Buildfarm Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build.buildfarm.worker;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import build.buildfarm.common.Claim.Lease;
+import build.buildfarm.common.Claim.Stage;
+
+public class CPULease implements Lease {
+  public static final String RESOURCE_NAME = "cpu.shares";
+  private int balance = 0;
+
+  CPULease(int amount, Market market) throws InterruptedException {
+    Order order = market.buy(amount, this::accumulate);
+    try {
+      synchronized (this) {
+        while (balance != amount) {
+          wait();
+        }
+      }
+    } catch (InterruptedException e) {
+      order.cancel();
+      // objectively finally, but we could still be interrupted while waiting for the last item
+      market.sell(balance);
+      throw e;
+    }
+  }
+
+  public synchronized void accumulate(int n) {
+    checkArgument(n >= 0);
+    balance += n;
+    notifyAll();
+  }
+
+  public synchronized void deplete(int n) {
+    checkArgument(n >= 0);
+    balance -= n;
+  }
+
+  @Override
+  public String name() {
+    return RESOURCE_NAME;
+  }
+
+  @Override
+  public synchronized int amount() {
+    return balance;
+  }
+
+  public Stage stage() {
+    return Stage.EXECUTE_ACTION_STAGE;
+  }
+}
--- a/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
+++ b/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
@@ -15,6 +15,7 @@
 package build.buildfarm.worker;
 
 import build.buildfarm.worker.resources.ResourceLimits;
+import io.prometheus.client.Histogram;
 import java.io.IOException;
 import java.util.Map;
 import java.util.concurrent.ArrayBlockingQueue;
@@ -29,6 +30,13 @@ import lombok.extern.java.Log;
 
 @Log
 public class ExecuteActionStage extends PipelineStage {
+  private static final Histogram executionTime =
+      Histogram.build().name("execution_time_ms").help("Execution time in ms.").register();
+  private static final Histogram executionStallTime =
+      Histogram.build()
+          .name("execution_stall_time_ms")
+          .help("Execution stall time in ms.")
+          .register();
   private static final int IDLE_STAGE_CLOSED_MS = 10;
   private static final TimeUnit IDLE_STAGE_CLOSED_UNIT = TimeUnit.MILLISECONDS;
 
@@ -237,6 +245,8 @@ public class ExecuteActionStage extends PipelineStage {
   }
 
   void releaseExecutor(String executionName, long usecs, long stallUSecs, int exitCode) {
+    executionTime.observe(usecs / 1000.0);
+    executionStallTime.observe(stallUSecs / 1000.0);
     complete(executionName, usecs, stallUSecs, String.format("exit code: %d", exitCode));
   }
 }
--- a/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
+++ b/src/main/java/build/buildfarm/worker/ExecuteActionStage.java
@@ -15,45 +15,30 @@
 package build.buildfarm.worker;
 
 import build.buildfarm.worker.resources.ResourceLimits;
-import io.prometheus.client.Gauge;
-import io.prometheus.client.Histogram;
 import java.io.IOException;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import lombok.extern.java.Log;
 
 @Log
-public class ExecuteActionStage extends SuperscalarPipelineStage {
-  private static final Gauge executionSlotUsage =
-      Gauge.build().name("execution_slot_usage").help("Execution slot Usage.").register();
-  private static final Histogram executionTime =
-      Histogram.build().name("execution_time_ms").help("Execution time in ms.").register();
-  private static final Histogram executionStallTime =
-      Histogram.build()
-          .name("execution_stall_time_ms")
-          .help("Execution stall time in ms.")
-          .register();
+public class ExecuteActionStage extends PipelineStage {
+  private static final int IDLE_STAGE_CLOSED_MS = 10;
+  private static final TimeUnit IDLE_STAGE_CLOSED_UNIT = TimeUnit.MILLISECONDS;
 
-  private final AtomicInteger executorClaims = new AtomicInteger(0);
+  public static final int SHARES_PER_SLOT = 1000;
 
-  public ExecuteActionStage(
-      WorkerContext workerContext, PipelineStage output, PipelineStage error) {
-    this(workerContext, output, error, workerContext.getExecuteStageWidth());
-  }
+  private final BlockingQueue<ExecutionContext> queue = new ArrayBlockingQueue<>(1);
+  private final ExecutorService executor = Executors.newCachedThreadPool();
+  private final Map<String, Executor> executions = new ConcurrentHashMap<>();
 
-  public ExecuteActionStage(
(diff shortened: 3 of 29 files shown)


src/main/java/build/buildfarm/worker/cgroup/Mem.java
```diff
@@ -45,4 +52,8 @@ public void setMemorySwapLimit(long limitBytes) throws IOException {
     open();
     writeLong("memory.memsw.limit_in_bytes", limitBytes);
   }
+
+  public void setCpu(int cpu_us) {
```
jasonschroeder-sfdc, line 57:
```suggestion
  @Override
  public void setCpu(int cpu_us) {
```
is this defined at a parent? I noticed `Cpu.java` has `@Override`

src/main/java/build/buildfarm/worker/shard/ShardWorkerContext.java
```diff
@@ -908,6 +947,12 @@ public void destroyExecutionLimits() {
       try {
         operationsGroup.getCpu().close();
         executionsGroup.getCpu().close();
+        try {
+          market.stop();
+        } catch (InterruptedException e) {
+          log.severe("market await shutdown interrupted");
```
jasonschroeder-sfdc: worth logging the `e`?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -595,6 +595,20 @@ message StageInformation {
   int32 slots_used = 3;
 
   repeated string operation_names = 4;
+
+  repeated WorkerExecution executions = 5;
+}
+
+message WorkerExecution {
+  string name = 1;
```
jasonschroeder-sfdc, line 604: doc the 'name' , and give example. is this the queue name?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -595,6 +595,20 @@ message StageInformation {
...
+
+  repeated Resource resources = 2;
+}
+
+message Resource {
+  string name = 1;
```
jasonschroeder-sfdc, line 611: doc the name and give example

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -595,6 +595,20 @@ message StageInformation {
...
+
+  int32 amount = 2;
```
jasonschroeder-sfdc, line 613: i'm assuming this cannot be negative?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
```diff
@@ -735,4 +749,6 @@ message WorkerExecutedMetadata {
   int64 fetched_bytes = 1;
 
   repeated string linked_input_directories = 2;
+
+  map<string, int64> usage = 3;
```
jasonschroeder-sfdc, line 758: doc !

src/main/java/build/buildfarm/worker/Order.java
```diff
@@ -0,0 +1,50 @@
...
+  synchronized void cancel() {
+    balance = 0;
+  }
+
+  // returns the number of shares bought from the available pool
+  synchronized int buy(int available) {
+    if (balance == 0 || available == 0) {
+      return 0;
+    }
+    checkState(available > 0);
```
jasonschroeder-sfdc:
```suggestion
    checkArgument(available > 0);
```

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -111,20 +133,51 @@ protected void iterate() throws InterruptedException {
...
     // might want to move this to actual teardown of the stage
+    executor.shutdown();
+    boolean interrupted = false;
+    try {
+      if (!executor.awaitTermination(1, TimeUnit.MINUTES)) {
+        getLogger().severe("executor did not terminate");
+      }
+    } catch (InterruptedException e) {
+      getLogger().severe("executor await shutdown interrupted");
+      interrupted = Thread.interrupted() || true;
```
jasonschroeder-sfdc, line 177: isn't this always `true`? was this a typo?

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -111,20 +133,51 @@ protected void iterate() throws InterruptedException {
...
-      start(executionContext.operation.getName(), getUsage(slotUsage));
+      start(executionContext.operation.getName(), actionExecutor);
       executor.execute(() -> actionExecutor.run(limits));
     }
   }
 
+  @Override
+  void put(ExecutionContext executionContext) throws InterruptedException {
+    while (!isClosed() && !output.isClosed()) {
+      if (queue.offer(executionContext, 10, TimeUnit.MILLISECONDS)) {
```
jasonschroeder-sfdc: is there any scenario when this timeout needs to be longer?

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -70,38 +62,68 @@ public void put(ExecutionContext executionContext) throws InterruptedException {
...
-    int slotUsage = executorClaims.addAndGet(-claims);
-    executionSlotUsage.set(slotUsage);
-    return slotUsage;
+  @Override
+  ExecutionContext take() throws InterruptedException {
+    boolean interrupted = false;
+    InterruptedException exception;
+    try {
+      while (!isClosed() && !output.isClosed()) {
+        ExecutionContext context = queue.poll(10, TimeUnit.MILLISECONDS);
```
jasonschroeder-sfdc: this "10 millis" is used (at least) twice. should we define a const? do they need to be the same?

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
```diff
@@ -35,25 +38,14 @@ public class ExecuteActionStage extends SuperscalarPipelineStage {
...
-    super(
-        "ExecuteActionStage",
-        "executor",
-        workerContext,
-        output,
-        createDestroyExecDirStage(workerContext, error),
-        width);
-  }
+  private final BlockingQueue<ExecutionContext> queue = new ArrayBlockingQueue<>(1);
+  private volatile boolean catastrophic = false;
```
jasonschroeder-sfdc: where is this set to `true`? I couldn't find it in the diff view

src/main/java/build/buildfarm/worker/CPULease.java
```diff
@@ -0,0 +1,61 @@
...
+      }
+    } catch (InterruptedException e) {
+      order.cancel();
+      // objectively finally, but we could still be interrupted while waiting for the last item
+      market.sell(balance);
+      throw e;
+    }
+  }
+
+  public synchronized void accumulate(int n) {
```
jasonschroeder-sfdc, line 42: checkArgument(n > 0)?

src/main/java/build/buildfarm/worker/CPULease.java
```diff
@@ -0,0 +1,61 @@
...
+    balance += n;
+    notifyAll();
+  }
+
+  public synchronized void deplete(int n) {
+    balance -= n;
```
jasonschroeder-sfdc, line 50: checkState(balance >= 0)?

src/main/protobuf/build/buildfarm/v1test/buildfarm.proto
werkt: this follows the convention set by remote-apis, the rest of our message decls, and prescribed by the proto type hints - I'm not going to preclude this value being negative here, but it is discouraged due to efficiency for fluidly-signed ints in favor of sint32.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: No, this clears the current interrupted state, but also guarantees that we assert that we were interrupted, regardless of whether the thread interrupted existed prior.
https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/lang/Thread.html#interrupted()
Without this clear, subsequent activity in the current thread will potentially throw its own interrupted exception (observed during IO calls).
What _is_ a bug here for sure is that the destroyExecutionLimits() is being called with the thread interrupted flag set, which will leak the stage resources in the event we were interrupted. Moving it above, as I'm pretty sure I intended before.
And the `|| true` just gives us assurance that we set the thread state if we were interrupted, as I'd like everything _after_ this method to know that we were.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: This isn't a time-limited insertion (note the while loop above) - it is a idle-loop ensuring that if the stage is closed, we get a reaction from the waiting process within 10ms

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: yeah, throwing that up top

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: All those LLMs and you can't find an identifier in the file ;)
https://github.com/buildfarm/buildfarm/blob/main/src/main/java/build/buildfarm/worker/SuperscalarPipelineStage.java#L184

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
jasonschroeder-sfdc: Consider:
```java
Thread.interrupted();
interrupted = true;
```
to be clearer about side-effects. i'd hate for a clever Javac to optimize-away the left-hand-side of that `||` ;-)

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
jasonschroeder-sfdc: That's a different file and also `private` so these are different variables?

src/test/java/build/buildfarm/actioncache/ShardActionCacheTest.java
```diff
@@ -84,23 +84,6 @@ public void readThroughShouldWriteLocalCacheOnlyAndNeverCallBackplane() throws E
...
-    for (int i = 0; i < 100; i++) {
-      cache.readThrough(key("filler-" + i), result("filler-" + i));
-    }
-
-    cache.get(KEY_A).get();
-
-    // Make sure the key is fetched from the backplane after being evicted due to cache size.
-    verify(backplane, times(1)).getActionResult(KEY_A);
-  }
-
```
jasonschroeder-sfdc, line 103: Intentionally deleted?

jasonschroeder-sfdc:
Two issues from review:

**2. Javadoc copy-paste in `ExecutionProperties.java`**

The new constants have wrong/duplicated Javadoc:
- `MAX_SHARES_SOLD` has `@field MIN_SHARES_SOLD` and the brief reads "CPU Shares will not be sold" — should describe the maximum cap.
- `MIN_SHARES_SOLD`, `PCT_MIN_THROTTLED`, and `CPU_SHARE_FLOOR` all reuse the same "Headroom for per-sample volatility of use before its cpu may be sold" detail, which only fits `PCT_MIN_UNUSED`.

**9. Possible double-close of `IOResource`**

In `Executor.executePolled` the resource is held in a try-with-resources:

```java
try (IOResource resource = workerContext.limitExecution(...)) {
    ...
    statusCode = executeCommand(..., resource, ...);
}
```

`executeCommand` → `executeProcess` → `executeNativeProcessAndDebug` → `executeNativeProcess`, which calls `resource.close()` explicitly before returning:

```java
// ensure resource release on process completion or timeout
resource.close();
```

Then the try-with-resources will call `close()` again. The `MultiResource` in `ShardWorkerContext` iterates child resources, and `Cpu.close()` / cgroup teardown may not be idempotent. Worth either removing the explicit `close()` or making the close path idempotent and documenting it.

src/main/java/build/buildfarm/worker/ExecuteActionStage.java
werkt: ok, apologies. Yes, this needs a proper solution. This was a poor relocation from superscalar when we switched modes.

werkt:
> **9. Possible double-close of `IOResource`**
> 
> In `Executor.executePolled` the resource is held in a try-with-resources:
> 
> ```java
> try (IOResource resource = workerContext.limitExecution(...)) {
>     ...
>     statusCode = executeCommand(..., resource, ...);
> }
> ```
> 
> `executeCommand` → `executeProcess` → `executeNativeProcessAndDebug` → `executeNativeProcess`, which calls `resource.close()` explicitly before returning:
> 
> ```java
> // ensure resource release on process completion or timeout
> resource.close();
> ```
> 
> Then the try-with-resources will call `close()` again. The `MultiResource` in `ShardWorkerContext` iterates child resources, and `Cpu.close()` / cgroup teardown may not be idempotent. Worth either removing the explicit `close()` or making the close path idempotent and documenting it.

The close() on the resource is sequentially idempotent, and it is important that we release its resources as early as possible - in particular before the join()s that could get killed by the poller. Not sure what the AI needs to see for that, but I'm adding a comment in the close() method in `interface IOResource`. Conventionally, I've always seen/ensured close() implemented as sequentially idempotent, and the pattern of t-w-r offers that in different docs about being fine to early-release.

src/main/java/build/buildfarm/common/ExecutionProperties.java
```diff
@@ -297,4 +297,44 @@ public class ExecutionProperties {
...
+   * @details Prevent available cpu from being reduced beyond this amount Must not be <= 0 for
+   *     market interactions
+   */
+  public static final String CPU_SHARE_FLOOR = "cpuShareFloor";
+
+  /**
+   * @field PCT_MIN_UNUSED
+   * @brief Percentage of owned shares that must be unused before sale
+   * @details Headroom for per-sample volatility of use before its cpu may be sold Must not be <= 0
+   *     for market interactions
```
jasonschroeder-sfdc:
```suggestion
   * @details Headroom for per-sample volatility of use before its cpu may be sold Must not be <= 0. Values should be between <code>0 < x < 1</code>
   *     for market interactions
```

src/main/java/build/buildfarm/worker/shard/Worker.java
```diff
@@ -141,6 +141,8 @@ public final class Worker extends LoggingMain {
           .register();
   private static final Counter workerPausedMetric =
       Counter.build().name("worker_paused").help("Worker paused.").register();
+  private static final Gauge executionSlotUsage = // now Market stock - balance
```
jasonschroeder-sfdc, line 144: Worthy of a release note?

src/main/java/build/buildfarm/worker/Market.java
```diff
@@ -0,0 +1,170 @@
...
+            if (balance != 0 || !clearCancelledOrders()) {
+              waitCycle();
+            }
+          }
+        }
+      }
+    } catch (InterruptedException e) {
+      // ignore
+    } catch (RuntimeException e) {
+      log.log(SEVERE, "Market run threw exception", e);
```
jasonschroeder-sfdc, line 124: what's the worse that can happen if the Market stops running?

src/main/java/build/buildfarm/worker/Market.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public void stop() throws InterruptedException {
+    broker.interrupt();
+    broker.join();
+  }
+
+  public void sell(int amount) {
+    if (amount != 0) {
+      // checkState(amount > 0);
```
jasonschroeder-sfdc: uncomment or delete this line

src/main/java/build/buildfarm/worker/WorkerContext.java
```diff
@@ -142,4 +151,12 @@ IOResource limitExecution(
   int commandExecutionClaims(Command command);
 
   ResourceLimits commandExecutionSettings(Command command);
+
+  default Market market() {
```
jasonschroeder-sfdc, line 156: Consider `@Nullable` or we can do it after NullAway

jasonschroeder-sfdc (approved):

src/main/java/build/buildfarm/worker/shard/Worker.java
werkt: Yes, though the math stays the same from the observer's perspective.

src/main/java/build/buildfarm/worker/WorkerContext.java
werkt: definitely not doing this after nullaway, there's too much downstream that I need to reconcile with this.

src/main/java/build/buildfarm/worker/Market.java
werkt: buys will not be processed and no executions will happen. objectively the worker should fail/shut down

PR #2579, opened 2026-06-19
fix(CFCLinkExecFileSystem): advance directory iterator correctly

stsun1026:
## Summary

This fixes `CFCLinkExecFileSystem.DirectoryIterator` so linked input directory discovery advances correctly after every returned directory path.

The iterator added for `output_paths` filtering cached the next `DirectoryNode`, but the cached value was not always advanced after `next()`. In particular, when the returned directory had the empty-directory digest, the iterator could keep returning the same path while `hasNext()` remained true because the current sibling iterator still had entries. That can make `linkedDirectories()` spin during exec directory creation.

The fix replaces the cached-node state with explicit DFS frames containing the current path prefix and child iterator. This preserves ignored-output-path filtering while ensuring each directory path is returned once and exhausted iteration follows the `Iterator` contract.

Related to:
* https://github.com/buildfarm/buildfarm/issues/2562

## Testing

Ran:
```shell
bazel test //src/test/java/build/buildfarm/worker:tests
bazel build //container:buildfarm-worker
bazel run //tools/lint/hawkeye -- check --fail-if-unknown
./.bazelci/format.sh
```

Added a regression test for an empty directory followed by a non-empty sibling.

Also validated the change in a staging Buildfarm cluster with `linkInputDirectories: true`. Before this patch, some remote actions would get stuck while preparing the exec directory, with worker thread dumps showing CPU time in `CFCLinkExecFileSystem.linkedDirectories()` / `DirectoryIterator.next()`. Deploying a worker image with this fix allowed the same build workload to get past that original input-fetch hang.

### DIFF
--- a/src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
+++ b/src/main/java/build/buildfarm/worker/CFCLinkExecFileSystem.java
@@ -50,6 +50,7 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.Stack;
 import java.util.concurrent.CancellationException;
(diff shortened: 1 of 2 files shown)


werkt (approved):
I poked at the test case (thank you!) without the fix for long enough to know that I didn't want to make the current impl fit into the empty/nonempty model, so I'll merge. I hope that the DFS doesn't throw off the size/perf from the current impl, but this is all pretty inefficient regardless, so thanks for the patch!

PR #2601, opened 2026-07-06
fix(worker): make processor count source configurable

### LINKED ISSUE #2600: Worker executeStageWidth auto-sizing uses host CPU count (OSHI), ignoring the container's cgroup CPU limit
When `worker.executeStageWidth` is 0 (auto), it's sized from `SystemProcessors.get()`, which returns:

[Math.max(get(PROCESSOR_DERIVE.JAVA_RUNTIME), get(PROCESSOR_DERIVE.OSHI))](https://github.com/buildfarm/buildfarm/blob/af34f28b586cd4c06e71da4786bf19028eb6bb49/src/main/java/build/buildfarm/common/SystemProcessors.java#L52-L54)

JAVA_RUNTIME resolves to `Runtime.getRuntime().availableProcessors()` (container-aware, honors the cgroup CPU limit). OSHI resolves to CentralProcessor.getLogicalProcessorCount(), which reads the physical host and isn't namespaced to the container.
Issue: when the node has more cores than the pod's CPU limit, max() discards the container-aware value and sizes the execute stage to the host.

Example: container CPU limit = N cores, host = M cores (M > N):
  - JAVA_RUNTIME → N (honors the cgroup limit)
  - OSHI → M (host, ignores the limit)
  - max(N, M) = M → executeStageWidth = M - executeStageWidthOffset

So the width is derived from the host core count, not the container's allocation, over-provisioning the worker.

Impact: the worker admits more concurrent actions than its CPU allocation supports, causing CPU contention/throttling. Because the count is also unrelated to the container's memory limit, memory-heavy action mixes can push the container past its memory limit and trigger OOMKills. (The execute stage is sized purely from CPU count with no memory bound; the host-vs-container CPU discrepancy amplifies this.)

adityarar:
Closes https://github.com/buildfarm/buildfarm/issues/2600

The default is MAX_OF_SOURCES, which is exactly the prior max(availableProcessors, OSHI) behavior, so this is fully backward compatible; no existing deployment changes.

#### Why not just change max() to min()?
The obvious one-line alternative, but unsafe as an unconditional change. OSHI and the max() were introduced in #1290 because Runtime.availableProcessors() under-reports in some container runtimes, returning 1 (per the class Javadoc: "k8s with containerd... give back 1"). Under min:
  - Runtime.availableProcessors() → 1
  - OSHI → real count
  - min(1, real) = 1 → worker sized to a single execution slot

So min reintroduces the exact bug #1290 fixed. max was a deliberate "trust whichever source sees more," chosen for a known under-count failure mode. This PR preserves that default and lets the operator opt into the container-aware source when their runtime reports CPU limits correctly, rather than flipping the tradeoff for everyone.

In practice, if your containers have CPU limits set, `JAVA_RUNTIME` is the one you want. It reads the cgroup limit the pod actually runs under, whereas `OSHI` reports the whole host's cores.

#### Design notes
  - SystemProcessors.get(strategy) is stateless (strategy passed as a parameter), so no global mutable state, no set-before-first-call ordering hazard.
  - All four call sites (`BuildfarmConfigs.adjustExecuteStageWidth`, `ShardWorkerContext.createOperationExecutionLimits`, both `BuildfarmExecutors` cache pools) source the strategy from worker.processorCountSource, so behavior is consistent.
  - Each source count is **memoized** to avoid repeated OSHI native (JNA) lookups.

#### Docs
Added processorCountSource to the worker config reference table and examples/config.yml.


werkt:
I think this encoding into a single field is unnecessarily complex. A single bool flag per field in terms of selection valve would be simpler and then the combination choice is still max, which seems incorrect on the surface.

adityarar:
Thanks for taking a look.

On the first point: I had kept the default as max deliberately, so existing setups don't change behavior on upgrade. The idea was to have a opt-in for containerized deployments rather than changing it for everyone.

I agree with the second point. `max` is the wrong combiner no matter how the config is shaped. `Runtime.availableProcessors()` answers "CPUs this container may use" (it honors the cgroup limit) and OSHI answers "CPUs on the host" (and ignores the limit). So `max(container, host)` effectively always returns the host count, which is exactly the over-provisioning I am trying to fix here.

#1290 added `OSHI` to work around availableProcessors() returning 1 on some containerd setups. JDK 21 with
UseContainerSupport reads the cgroup quota directly and only returns 1 when the container actually has ~1 CPU, it doesn't under-report. I see that the project already standardizes on JDK 21, so that failure case shouldn't arise on a supported runtime anymore.

Given that, I think the simplest and most correct fix is to drop OSHI and the max and just trust availableProcessors(), no config field at all. That removes the complexity you flagged and fixes the correctness at the same time. I'll rework the PR in that direction unless you see a supported setup where availableProcessors() would still under-report.

werkt:
Centering on removing OSHI here fixes both problems given JDK 21, please proceed with reworking the PR to default and remove any unnecessary OSHI implementation.

adityarar:
Made the changes. This cleaned up a lot of code by removing the existing OSHI implementation.

There is another unrelated OSHI reference in [Backplane.java](https://github.com/buildfarm/buildfarm/blob/af34f28b586cd4c06e71da4786bf19028eb6bb49/src/main/java/build/buildfarm/common/config/Backplane.java#L156) which, if we want to clean it up, should be handled in a separate PR since it would require a behavioral change. Happy to take that up in a separate PR.

https://github.com/buildfarm/buildfarm/blob/af34f28b586cd4c06e71da4786bf19028eb6bb49/src/main/java/build/buildfarm/common/config/Backplane.java#L156

adityarar:
@werkt Is this good to merge?

werkt (approved):

PR #2604, opened 2026-08-13
fix(worker): accept bazel's "resources:" prefix for local resources

fmeum:
Bazel spells extra resource requests as `resources:<name>`, both in `tags` and in `exec_properties` (see [`ExecutionRequirements.RESOURCES`](https://github.com/bazelbuild/bazel/blob/d62b2b8bc05f8ee4d82262d3eddff179490c857a/src/main/java/com/google/devtools/build/lib/actions/ExecutionRequirements.java#L165)).

Buildfarm only matches the singular `resource:<name>`, even though the comment on `getResourceName` says the prefix "originates from bazel's syntax". An action written using Bazel's documented syntax therefore never actually claimed a local resource.

This change makes it so that both spellings are accepted, leaving existing configurations and actions are unaffected.


werkt (approved):
Matches the currently parsed execution format recommended by bazel.
Continues to support resource: for now. Looks good.

PR #2606, opened 2026-08-21
Make AtomicFileWriter more

blainefreestone:
The "atomic" swap on success was Files.delete(target) followed by Files.createLink(target, temp) — two syscalls with a window where a JVM crash, host crash, or fault between them leaves the target gone and only the temp .uuid file behind. Subsequent startup paths that expected the target to exist (CASFileCache's lru.txt) would see a missing file and fall back to a full filesystem scan; on a 54TB CAS worker that's an extra long minute of cold cache misses on every unclean restart.

Switch replace() to Files.move(temp, target, ATOMIC_MOVE, REPLACE_EXISTING) — a single renameat2 on POSIX filesystems with the atomicity the class name has claimed all along. Filesystems that don't support atomic rename (rare; some Windows cross-volume cases) fall back to the legacy delete + createLink with a WARNING log so operators see the reduced durability rather than discovering it after a crash. The outer finally switches Files.delete(temp) to deleteIfExists since after a successful move the temp file no longer exists; on the failure paths the temp still does and gets cleaned up as before.

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -70,6 +70,7 @@ import build.buildfarm.common.io.CountingOutputStream;
 import build.buildfarm.common.io.Directories;
 import build.buildfarm.common.io.FeedbackOutputStream;
 import build.buildfarm.common.io.FileStatus;
+import build.buildfarm.common.io.FileSystemMover;
 import build.buildfarm.v1test.BlobWriteKey;
 import build.buildfarm.v1test.Digest;
 import com.google.common.annotations.VisibleForTesting;
@@ -174,7 +175,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
   private final Consumer<Iterable<Digest>> onExpire;
   private final Executor accessRecorder;
   private final ExecutorService expireService;
-  private final LRUDB db = new TextLRUDB();
+  private LRUDB db;
+  private FileSystemMover fileSystemMover;
   private volatile Deadline saveLRUAfter = Deadline.after(10, MINUTES);
   private final Path lru;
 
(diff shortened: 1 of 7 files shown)


src/main/java/build/buildfarm/common/io/AtomicFileWriter.java
```diff
@@ -109,19 +114,30 @@ public void close() throws IOException {
...
+      Files.move(temp, target, ATOMIC_MOVE, REPLACE_EXISTING);
+    } catch (AtomicMoveNotSupportedException e) {
+      // Filesystem doesn't support atomic rename (rare on POSIX; possible on some networked or
+      // cross-volume Windows setups). Fall back to the non-atomic delete + createLink sequence.
+      // A JVM crash between the delete and createLink leaves the target gone — log loudly so
+      // operators see the reduced durability rather than discovering it after a crash.
+      log.log(
+          Level.WARNING,
+          "atomic move not supported for " + target + "; falling back to non-atomic replace",
+          e);
```
werkt: We can't fall back to this on every write - the capability for this on the filesystem of the writer (that really should not change while the worker is running) should be determined once (on startup), and then declare the fixed path for this.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -174,7 +175,8 @@ public abstract class CASFileCache implements ContentAddressableStorage {
   private final Consumer<Iterable<Digest>> onExpire;
   private final Executor accessRecorder;
   private final ExecutorService expireService;
-  private final LRUDB db = new TextLRUDB();
+  private LRUDB db;
+  private FileSystemMover fileSystemMover;
```
werkt: Please change this to fileStoreMover to match the selection criteria

blainefreestone: Done

PR #2608, opened 2026-08-25
Add maxSizePercent config for CAS filesystem storage

blainefreestone:
Currently when maxSizeBytes is 0, the CAS size limit is auto configured to 90% of the filesystem. We found that value to be too aggressive for our exec workers and were running out of disk space.

This change adds a configurable maxSizePercent option that is mutually exclusive with maxSizeBytes. When maxSizePercent is set, it is used to limit the percent of the filesystem used for the CAS.

If neither maxSizeBytes nor maxSizePercent is set, then the maxSizePercent is used and set to 90%, which preserves the prior default behavior.


jasonschroeder-sfdc (approved):
Approved. Though a `mazSizePercent==0` is probably not a good idea 🤓

PR #2610, opened 2026-08-27
Otel

werkt:


jasonschroeder-sfdc (approved):
LGTM, there are some better ways to configure otel via env vars, and make it work better in conjunction with javaagent. 

I'd rather use env vars to configure the endpoint and transport. If you're OK with a fast-follow pr, we can merge this, but i would like to revise before we cut a release.

PR #2611, opened 2026-08-28
Cas size accounting

blainefreestone:
* Adds a rough accounting for the disk space consumed by directory inodes. The accounting is accurate at startup and an estimate at runtime.
* Changes the CAS size tracking to use an estimate of actual size on disk. Currently the CAS tracks its size using logical file sizes. This likely under estimates the physical size of the CAS on-disk. It estimates physical size by rounding file size up to the nearest block boundary.

### DIFF
--- a/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
+++ b/src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
@@ -1288,6 +1288,11 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     }
   }
 
+  @VisibleForTesting
+  void setBlockSizeForTesting(long blockSize) {
+    this.blockSize = blockSize;
+  }
+
   @SuppressWarnings({"PMD.CompareObjectsWithEquals"})
   private synchronized List<SizeEntry> lruSizeEntryList() {
     /**
@@ -1541,7 +1546,10 @@ public abstract class CASFileCache implements ContentAddressableStorage {
     } else {
       // if cas is full or entry is oversized or empty, mark file for later deletion.
       long size = entry.size();
-      if (sizeInBytes + size > maxSizeInBytes || size > maxEntrySizeInBytes || size == 0) {
+      if (sizeInBytes + estimateSizeOnDisk(size, blockSize, /* isHardlink= */ false)
+              > maxSizeInBytes
+          || size > maxEntrySizeInBytes
+          || size == 0) {
         synchronized (deleteFiles) {
           deleteFiles.add(path);
         }
@@ -1564,7 +1572,7 @@ public abstract class CASFileCache implements ContentAddressableStorage {
             if (e.decrementReference(header)) {
               unreferencedEntryCount++;
             }
-            sizeInBytes += size;
+            sizeInBytes += estimateSizeOnDisk(size, blockSize, /* isHardlink= */ false);
           }
         }
       }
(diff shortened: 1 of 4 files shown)


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -219,7 +219,11 @@ public SettableFuture<Long> load(Digest digest) {
                 }
               });
 
+  private static final long DEFAULT_BLOCK_SIZE = 4096;
+  private static final long ESTIMATED_BYTES_PER_DIRECTORY_ENTRY = 32;
+
   protected FileStore fileStore; // bound to root
+  protected long blockSize = DEFAULT_BLOCK_SIZE;
```
werkt, line 226: I prefer having fileStore as a container for this value - it has to be set for every use case that I can think of, and it has a getBlockSize() field. Mock/instrument the filestore for testing per below

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2012,14 +2030,66 @@ private void putDirectoryFiles(
...
+   * already-accounted inode is being made.
+   */
+  @VisibleForTesting
+  static long estimateSizeOnDisk(long logicalSize, long blockSize, boolean isHardlink) {
+    checkArgument(blockSize > 0, "blockSize (%s) must be positive", blockSize);
+    checkArgument(logicalSize >= 0, "logicalSize (%s) must be non-negative", logicalSize);
+    if (isHardlink) {
+      // Hardlinks reuse an existing inode's blocks and add only a directory entry (~64 bytes on
+      // typical filesystems, which rounds to 0 under block alignment).
+      return 0;
```
werkt, line 2056: Nothing practical calls this with isHardLink = true that I can see.

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2012,14 +2030,66 @@ private void putDirectoryFiles(
...
+    }
+    if (logicalSize == 0) {
+      return 0;
+    }
+    // Use long division to get the ceiling in order to round up to the next largest blocksize.
+    // This should return correct answers for sizes that are block aligned as well as those that
+    // are not.
+    return ((logicalSize + blockSize - 1) / blockSize) * blockSize;
```
werkt, line 2064: Seems like this should be `ceilDiv` as below

src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -2012,14 +2030,66 @@ private void putDirectoryFiles(
...
+   * the compression ratio.
+   *
+   * <p>When {@code isHardlink} is {@code true}, the caller is creating (or has verified) a hardlink
+   * to an already-resident inode rather than writing a fresh inode; the physical cost is then just
+   * a directory entry (~64 bytes on typical filesystems, rounding to 0 under block alignment), so
+   * this method returns 0. Callers must only pass {@code true} when a hardlink to an
+   * already-accounted inode is being made.
+   */
+  @VisibleForTesting
+  static long estimateSizeOnDisk(long logicalSize, long blockSize, boolean isHardlink) {
```
werkt, line 2050: Ask, and it's a big one, please make this and the derivatives `estimateFileStoreSize` in keeping with the nomenclature and the fact that we don't usually run on platters or floppies.

werkt (commented):
I think I'm not completely sold on the principle of this change. "The filesystem adds some overhead" needs to come to some sort of practical limit that isn't being accounted for, and therefore represents an unconstrained or non-cardinal use: what factor is being omitted already that can definitely overwhelm usage beyond a healthy buffer from the filestore capacity?

werkt:
@blainefreestone mind cleaning up the commits here? I'd like to land this with a rebase with clear authors (ubuntu is listed) and meaningful changes (elaborate on 'stuff')

PR #2612, opened 2026-08-28
Fix native memory leak in StubWriteOutputStream

blainefreestone:
This commit fixes two native memory leaks in StubWriteOutputStream:

1. Currently StubWriteOutputStream.close() can throw an exception and prevent the onError() or onCompleted() functions from being called on the finishedWriteObserver. That error being thrown can also prevent writeObserver from being nulled.

2. Currently When the remote gRPC stream closes before the local, writeObserver is set to null, which prevents onError() or onCompleted() from being called when close() is called.

In both of these scenarios the local gRPC stream doesn't get properly closed and Netty never releases a native memory byte buffers that were allocated using Unsafe. This causes a native memory leak.

This change fixes both leaks and adds tests for them that fail without the fix and pass with the fix.

It fixes scenario 1 by moving observer cleanup into a finally block. That way the cleanup functions are always called. We also make sure that writeObserver is always nulled out.

It fixes scenario 2 by using a boolean to separately track whether the remote gRPC stream is closed. That way the local stream still gets closed in the close() function regardless of whether the remote is closed.

This change by itself doesn't fully fix the native memory leaks. There's a few other ways native memory can leak in this class.

This change also introduces a test that fails without the fix and passes with the fix.

### DIFF
--- a/src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
+++ b/src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
@@ -16,6 +16,7 @@ package build.buildfarm.common.grpc;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.util.concurrent.Futures.immediateFailedFuture;
 import static com.google.common.util.concurrent.Futures.immediateFuture;
 import static java.lang.String.format;
 import static java.util.logging.Level.WARNING;
@@ -28,6 +29,7 @@ import com.google.bytestream.ByteStreamProto.QueryWriteStatusRequest;
 import com.google.bytestream.ByteStreamProto.QueryWriteStatusResponse;
 import com.google.bytestream.ByteStreamProto.WriteRequest;
 import com.google.bytestream.ByteStreamProto.WriteResponse;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.base.Throwables;
@@ -111,6 +113,22 @@ public class StubWriteOutputStream extends FeedbackOutputStream implements Write
   @GuardedBy("this")
   private StreamObserver<WriteRequest> writeObserver = null;
 
+  @GuardedBy("this")
+  private boolean isGrpcRemoteCallClosed = false;
+
+  @GuardedBy("this")
+  private boolean isFullyClosed = false;
+
+  @VisibleForTesting
+  synchronized boolean isGrpcRemoteCallClosed() {
+    return isGrpcRemoteCallClosed;
+  }
+
+  @VisibleForTesting
+  synchronized boolean isFullyClosed() {
+    return isFullyClosed;
+  }
+
   private long deadlineAfter = 0;
   private TimeUnit deadlineAfterUnits = null;
   private Runnable onReadyHandler = null;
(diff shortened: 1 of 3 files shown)


src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -141,26 +159,52 @@ public StubWriteOutputStream(
 
   @Override
   public void close() throws IOException {
+    synchronized (this) {
+      if (isFullyClosed) {
+        return;
+      }
+    }
```
werkt: `if (isFullyClosed()) {` functional equivalence

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -141,26 +159,52 @@ public StubWriteOutputStream(
...
-      } else {
-        finishedWriteObserver.onCompleted();
+    } catch (Exception e) {
+      cancelled = true;
+      // If we're currently racing a concurrent tear down, then don't throw unnecessarily
+      boolean racingWithTeardown;
+      synchronized (this) {
+        racingWithTeardown = isFullyClosed;
+      }
+      if (!racingWithTeardown) {
```
werkt: `if (!isFullyClosed()) {` functional equivalence

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -216,7 +260,24 @@ private boolean checkComplete() throws IOException {
...
+  private synchronized void initiateWrite() throws IOException {
+    if (isFullyClosed) {
+      // If we're already fully closed, then don't initiate another write on this same instance
+      Throwable cause = null;
+      try {
+        writeFuture.get();
+      } catch (ExecutionException e) {
+        cause = e.getCause();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
```
werkt, line 271: cause is not set here, should probably be the InterruptedException, no?

werkt (commented):
Some notes on minor issues.
These streams are so incredibly difficult to get right that I don't understand how every implementation of bytestream is required to lift this, even if just for java. I feel that there's state logic that I'm going to have to look at this for, and would love to have enumerated in a diagram in comments in this class's implementation.

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
blainefreestone: Replaced

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
blainefreestone: Replaced

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
blainefreestone: Correct, added it

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -216,7 +254,25 @@ private boolean checkComplete() throws IOException {
...
+  synchronized boolean hasActiveWriteObserver() {
+    return writeObserver != null;
+  }
+
+  private synchronized void initiateWrite() throws IOException {
+    if (isFullyClosed) {
+      // If we're already fully closed, then don't initiate another write on this same instance
+      Throwable cause = null;
+      try {
+        writeFuture.get();
```
werkt, line 267: we should probably have a checkState expecting `isDone()` here - otherwise we will be sitting with the synchronized lock on this stream in a wait

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
```diff
@@ -216,7 +254,25 @@ private boolean checkComplete() throws IOException {
     return false;
   }
 
-  private synchronized void initiateWrite() {
+  @VisibleForTesting
+  synchronized boolean hasActiveWriteObserver() {
+    return writeObserver != null;
+  }
```
werkt, line 260: This is an awkward internal-state leak - the determination here should be whether the observer returned by the stub, mocked in all cases, receives appropriate method signals.

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
blainefreestone: Added the `checkState(writeFuture.isDone())` precondition before `get()`

src/main/java/build/buildfarm/common/grpc/StubWriteOutputStream.java
blainefreestone: Removed the accessor and state assertions. Updated the test accordingly.

werkt (approved):

PR #2613, opened 2026-08-31
Close read/write streams on zstd-decompression instantiation failures

denbeigh2000:
Currently, when we fail to construct our zstd-decompressing {in,out}put streams, we hard-fail without closing the streams, meaning readers/writers will be polling forever.

This ensures that we catch exceptions at these times, and reliably close said output streams.

---

I'll be following up with a couple of other small PRs, that:
1. add a configurable timeout to borrows from the fixed-size zstd buffer pool
2. adds some additional logging, gauges, etc around this area


src/main/java/build/buildfarm/cas/cfc/CASFileCache.java
```diff
@@ -1096,14 +1096,28 @@ public synchronized FeedbackOutputStream getOutput(
...
+                // closing uniqueOut, that future never completes and the next getOutput() for
+                // this write waits on it forever, before it can reach reset(). close() resolves
+                // the future and leaves the delegate open, so a partial write stays resumable.
+                // The zstd constructor reaches here whenever the buffer pool refuses a borrow.
+                try {
+                  uniqueOut.close();
+                } catch (IOException closeError) {
+                  e.addSuppressed(closeError);
+                }
+                throw e;
```
denbeigh2000: this ends up tripping cognitive complexity limits, moving to private method

werkt (approved):
Looks good. Some of the lack of encapsulation of the buffer pools is bothering me, but not enough to hold it up.
I think this should eventually end up in a zstd stream specialization class that can prevent leaks before the exception reaches the consumer.

